PR #70, opened 2015-03-16
Modification of the spool function in Notify.pm

mohadesz:
Modified the spool function to use NOW() instead of UNIX_TIMESTAMP() for TimeSpooled column. The reason for this is change is that the data-type for the TimeSpooled column in the notification_spool table is now changed from int to datetime.


samirdas:
Is this a different issue from previous.

mohadesz:
The code is already merged and is therefore closed now

PR #74, opened 2015-03-20
Revert "Bug fixes and improvements"

samirdas:
Reverts aces/Loris-MRI#72


samirdas:
Add to Mac install. And add caveat label.

PR #83, opened 2015-04-22
fix syntax error

liangfu:
fix syntax error in running installation package;

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,10 @@ See aces/Loris README.md for further information and Loris installation informat
 
 3. Run installer to install MINC & DICOM toolkits, Perl libraries, configure environment, and setup directories:
 
+   ```bash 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,8 +28,10 @@ See aces/Loris README.md for further information and Loris installation informat
 
 3. Run installer to install MINC & DICOM toolkits, Perl libraries, configure environment, and setup directories:
 
+   ```bash 
    cd /data/$projectname/bin/mri/
-   ```bash imaging_install.sh```
```
mohadesz: Can you please make sure that it is explicitly running bash? We have had problems in the past if bash is not used.

liangfu: I think so, because the first line of the file is

```bash
#!/bin/bash
```

Jkat: Awesome thanks!

PR #87, opened 2015-06-01
Fixed the issue with the queries

mohadesz:
Modified data-extraction method, for example 
$row['TarchiveID']  is modified to $row[1] as it was causing issues.


rathisekaran:
tested and validated

PR #90, opened 2015-06-14
Inclusion of PET in the Imaging pipeline

cmadjar:


uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -398,39 +398,46 @@ Returns: Textual name of scan type
...
+    my ($tr, $te, $ti);
+    if ($fileref->{parameters}{modality} eq "MR") {
+        $tr = ${fileref}->getParameter('repetition_time');
+        $te = ${fileref}->getParameter('echo_time');
+        $ti = ${fileref}->getParameter('inversion_time');
+        if (defined($tr)) {  $tr = &Math::Round::nearest(0.01, $tr*1000);  }
+        if (defined($te)) {  $te = &Math::Round::nearest(0.01, $te*1000);  }
+        if (defined($ti)) {  $ti = &Math::Round::nearest(0.01, $ti*1000);  }
+        my $time = ${fileref}->getParameter('time'); 
+    } elsif ($fileref->{parameters}{modality} eq "MR") {
```
mohadesz: This  should be PT , not MR

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -398,39 +398,46 @@ Returns: Textual name of scan type
...
+        # Place to add stuff specific to PET images
+    }
```
mohadesz: This condition is missing an else

cmadjar: I corrected the typo in the elsif statement. Good catch!

For the else statement, if there is nothing to do on an else statement there is no need to add useless lines of code. if, elsif is enough in that case.

PR #91, opened 2015-07-28
Implemented NIfTI conversion in the imaging pipeline.

cmadjar:
The NIfTI files will be linked to their MINC files using the parameter_file table with ParameterTypeID corresponding the "check_nii_filename". 

ATTENTION!!!
Need to update profileTemplate to add either:
- $no_nii = 0   # if want to use this functionality, or 
- $no_nii = 1   # if do not want to create NIfTI files.


driusan:
Can you add a default in the code so that existing projects don't break if the template isn't modified?

PR #104, opened 2016-01-04
Enabled Notification Spool in MRIProcessingUtility.pm and Informative messages print only if $verbose option is set

MounaSafiHarab:
- notification_spool_table from MRIProcessingUtility.pm gets updated now with messages (the messages are similar to the messages in the log files), to be consistent with the ImagingUpload.pm
- printing messages to LOG files is still enabled for now
- printing messages on-screen only if $verbose option is set (when 'launching imaging_upload_file.pl' OR 'imaging_upload_file_cronjob.pl' with -verbose)
- Messages printed to LOG files and to notification_spool table  irrespective of the verbose option. Error messages print all the time, irrespective of verbose option as well. The notification_spool table now has a column with a verbose flag. This flag will be used by the Imaging Upoloader in the front-end to decide which messages are reported under the "Summary" or "Detailed" options
- In ImagingUpload.pm, dcmdump() and isDicom() (more specifically the 'file' command) do NOT print a message for each file even if $verbose is set to 1
- the messages in the notification_spool table have the upload_id parameter set, so this assumes that the MRI images are uploaded using the imaging_uploader which creates this upload_id in the mri_upload table (running tarchiveLoader will get the upload_id from mri_upload table using the tarchive_id directly, or indirectly using tarchive SourceLocation)
- if $debug is now reserved for printing queries
- a few cleanup on the printed messages to be more informative
- To be used in conjunction with:
https://github.com/aces/dicom-archive-tools/pull/31
and
https://github.com/aces/Loris/pull/1567 (for the SQL patches)

### DIFF
--- a/uploadNeuroDB/NeuroDB/ImagingUpload.pm
+++ b/uploadNeuroDB/NeuroDB/ImagingUpload.pm
@@ -130,7 +130,7 @@ sub IsValid {
         $message =
             "\n The uploadID "
           . $this->{'upload_id'}
-          . "Does Not Exist ";
+          . " Does Not Exist \n";
         $this->spool($message, 'Y');
         return 0;
     }
@@ -146,7 +146,8 @@ sub IsValid {
             "\n The Scan for the uploadID "
           . $this->{'upload_id'}
           . " has already been ran with tarchiveID: "
-          . $row[1];
+          . $row[1]
+	  . "\n";
         $this->spool($message, 'Y');
         return 0;
     }
@@ -177,7 +178,7 @@ sub IsValid {
 
     if ( $files_not_dicom > 0 ) {
         $message = "\n ERROR: there are $files_not_dicom files which are "
-          . "Are not of type DICOM";
+          . "Are not of type DICOM \n";
         $this->spool($message, 'Y');
         return 0;
     }
@@ -185,7 +186,7 @@ sub IsValid {
     if ( $files_with_unmatched_patient_name > 0 ) {
         $message =
             "\n ERROR: there are $files_with_unmatched_patient_name files"
-          . " where the patient-name doesn't match ";
+          . " where the patient-name doesn't match \n";
         $this->spool($message, 'Y');
         return 0;
     }
@@ -354,7 +355,7 @@ sub PatientNameMatch {
     if ($pname ne  $this->{'pname'}) {
         my $message = "The patient-name $pname does not Match " .
             $this->{'pname'};
-    $this->spool($message, 'Y');
+    	$this->spool($message, 'Y');
         return 0; ##return false
     }
     return 1;     ##return true
@@ -554,7 +555,7 @@ sub updateMRIUploadTable  {
     my $this = shift;
 
     my ( $field, $value ) = @_;
-        ########################################################
+	########################################################
         #################Update MRI_upload table accordingly####
         ########################################################
         my $where = "WHERE UploadID=?";
--- a/uploadNeuroDB/NeuroDB/ImagingUpload.pm
+++ b/uploadNeuroDB/NeuroDB/ImagingUpload.pm
@@ -12,6 +12,15 @@ use NeuroDB::Notify;
 use File::Temp qw/ tempdir /;
 
 ################################################################
+################# Define Constants #############################
+################################################################
+
+my $notify_detailed  = 'Y';  # notification_spool message flag for messages to be displayed 
+                             # with DETAILED OPTION in the front-end/imaging_uploader 
+my $notify_notsummary = 'N'; # notification_spool message flag for messages to be displayed 
+                             # with SUMMARY Option in the front-end/imaging_uploader 
+
+################################################################
 #####################Constructor ###############################
 ################################################################
 =pod
@@ -47,8 +56,8 @@ sub new {
     ############### Create a Notify Object #####################
     ############################################################
 
-    my $Notify = NeuroDB::Notify->new( $dbhr );
-    $self->{'Notify'} = $Notify;
+    my $Notify 			    = NeuroDB::Notify->new( $dbhr );
+    $self->{'Notify'} 		    = $Notify;
     $self->{'uploaded_temp_folder'} = $uploaded_temp_folder;
     $self->{'dbhr'}                 = $dbhr;
     $self->{'pname'}                = $pname;
@@ -58,12 +67,12 @@ sub new {
 }
 
 ################################################################
-#####################IsValid####################################
+#####################IsCandidateInfoValid#######################
 ################################################################
 =pod
-IsValid()
+IsCandidateInfoValid()
 Description:
- Validates the File to be upload:
+ Validates the File to be uploaded:
  If the validation passes the following will happen:
   1) Copy the file from tmp folder to the /data/incoming
   2) Set the IsCandidateInfoValidated to true in the 
@@ -74,7 +83,7 @@ Arguments:
 
  Returns: 0 if the validation fails and 1 if passes
 =cut
-sub IsValid {
+sub IsCandidateInfoValid {
     my $this = shift;
     my ($message,$query,$where) = '';
     ############################################################
@@ -96,13 +105,10 @@ sub IsValid {
     ############################################################
     my $files_not_dicom                   = 0;
     my $files_with_unmatched_patient_name = 0;
-    my $is_valid                          = 0;
+    my $is_candinfovalid                  = 0;
     my @row                               = ();
-    my $notification_verbose		  	  = 'N';
     ############################################################
     ####Get a list of files from the folder#####################
-    ############################################################
-    ############################################################
     #############Loop through the files#########################
     ############################################################
     my @file_list;
@@ -115,7 +121,6 @@ sub IsValid {
     );
     ############################################################
     ############### Check to see if the uploadID exists ########
-    ############################################################
     ############################################################
     $query =
         "SELECT PatientName,TarchiveID,number_of_mincCreated,"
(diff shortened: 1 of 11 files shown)


uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -142,10 +142,11 @@ sub IsValid {
     if ( ( $row[1] ) || ( $row[2] ) ) {
 
         $message =
-            "\n The Scan for the uploadID "
+            "\nThe Scan for the uploadID "
```
mohadesz: why not extra space?

uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -376,9 +383,10 @@ Arguments:
 sub isDicom {
     my $this         = shift;
     my ($dicom_file) = @_;
-    my $file_type    = $this->runCommand("file $dicom_file");
+    my $cmd    = "file $dicom_file";
+    my $file_type    = `$cmd`;
     if ( !( $file_type =~ /DICOM/ ) ) {
-        print "not of type DICOM" if $this->{'verbose'};
+        print "\n Not of type DICOM \n";
```
mohadesz: isn't verbose needed for this?

uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -478,7 +486,7 @@ Arguments:
 sub runCommand {
     my $this = shift;
     my ($command) = @_;
-    print "\n\n $command \n\n " if $this->{'verbose'};
+    print "\n\n $command \n\n " if $this->{verbose};
```
mohadesz: shouldn't it be $this->{'verbose'} ?

uploadNeuroDB/NeuroDB/ImagingUpload.pm
MounaSafiHarab: @mohadesz,

To maintain consistency in the indentation across all the messages from the complete steps (whether displayed on screen as the preprocessing is launched from the back end, or in the notification spool table). This is purely aesthetic.

uploadNeuroDB/NeuroDB/ImagingUpload.pm
MounaSafiHarab: @mohadesz,

We print *error* messages irrespective of what $verbose is set to.

uploadNeuroDB/NeuroDB/ImagingUpload.pm
MounaSafiHarab: @mohadesz,

It only worked as expected as I removed the single quotation. If you check the MRIProcessing utility (for example lines 315 and 319 in computeMd5hash(), it seems to be used without single quotations as well). I am really not sure why.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -643,14 +730,20 @@ sub registerScanIntoDB {
...
+        if ($this->{verbose}){
+            $message = "\n FileID: $fileID\n";
+            $this->spool($message, 'N', $upload_id);
+        }  
 
         ########################################################
         ### update mri_acquisition_dates table #################
         ########################################################
-        &$this->update_mri_acquisition_dates(
+        $this->update_mri_acquisition_dates(
```
MounaSafiHarab: this will still need to be maintained at &

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -361,9 +419,19 @@ sub getAcquisitionProtocol {
...
                   );
-        $this->{LOG}->print(
-            "Worst error: $checks[0]\n"
-        ) if $this->{debug};
+    	if ($this->{verbose}){
+		$message = "\n Worst error: $checks[0]\n";
+		$this->{LOG}->print($message);
+		# Here I am assuming that 'warn' and 'pass' are
+		# not errors, and only 'exclude' is an error in  
+		# the notification_spool_table
```
MounaSafiHarab: Please verify this is a valid assumption

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -852,11 +964,16 @@ sub CreateMRICandidates {
     my $query = '';
     my ($subjectIDsref,$gender,$tarchiveInfo,$User,$centerID) = @_;
     my ($message);
+    my $tarchive_id = '';
+    my $upload_id   = '';
```
MounaSafiHarab: should we initialize these (tarchive_id and upload_id)  to undef instead?

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -212,6 +238,10 @@ sub createTarchiveArray {
                       "This seems not to be a valid archive for this study!".
                       "\n\n";
         $this->writeErrorLog($message, 3);
+	# no $tarchive nor $tarchive_id can be fetched since the archive 
+	# could not be found, making $upload_id hard to trace, so send some 
+	# random number (99999) instead for $upload_id in the notification_spool
+        $this->spool($message, 'Y', 99999);
```
MounaSafiHarab: this is the only place where we would not be able to fetch the upload_id from the tarchive_id as the tarchive can not be found, so I chose to spool this number... please verify this is acceptable

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -309,18 +350,25 @@ sub get_acquisitions {
 ################################################################
 sub computeMd5Hash {
     my $this = shift;
-    my ($file) = @_;
-    $this->{LOG}->print(
-        "==> computing md5 hash for MINC body.\n"
-    ) if $this->{verbose};
+    my ($file, $tarchive_id) = @_;
```
MounaSafiHarab: in 2 or 3 instances like here in computeMd5Hash(), the function has been changed to pass on the $tarchive_id (so we can spool the message with the needed information). Please verify this is Okay.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -94,29 +96,41 @@ sub lookupNextVisitLabel {
...
-Most important function now. Gets the tarchive and 
+Most important function now. Gets the tarchive and
 extracts it so data can actually be uploaded
 =cut
 sub extract_tarchive {
     my $this = shift;
-    my ($tarchive) = @_;
-    print "Extracting tarchive $tarchive in $this->{TmpDir} \n" 
-    if $this->{verbose};
+    my ($tarchive, $tarchive_id, $verbose) = @_;
```
MounaSafiHarab: theoretically, I should not need to pass on $verbose here (given it can be obtained from this->{verbose}), but it only worked if I did what I did...
in 2 more places down as well...

uploadNeuroDB/NeuroDB/ImagingUpload.pm
nicolasbrossard: You are accessing the verbose attribute of object $this so the quotes must not be there. This is typical Perl OO field access syntax.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -695,9 +790,13 @@ sub dicom_to_minc {
         close MAIL;
         croak("dicom_to_minc failure, exit code $exit_code");
```
MounaSafiHarab: @nicolasbrossard ,
should we spool there? it is an error at this point but it does NOT get logged... maybe we should spool anyway?

uploadNeuroDB/NeuroDB/ImagingUpload.pm
MounaSafiHarab: Okay, so line 465 above ( line 457 in the old file) should be changed then to $this->{verbose} as well.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
nicolasbrossard: Yep, this should be spooled IMHO. This way when the notification tables are used to forward the pipeline output back to the user's web browser, this message will appear on the webpage.

uploadNeuroDB/NeuroDB/ImagingUpload.pm
nicolasbrossard: On second thoughts, your object is also a hash array so I think the verbose property can be accessed both ways (with or without quotes). You sure it does not work *without* quotes (I think it should) ?

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
nicolasbrossard: Why would you want to put the & back ? I think $this->update_mri_acquisition_dates(...) is the appropriate way to call method update_mri_acquisition_dates on object $this.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
MounaSafiHarab: @nicolasbrossard, Ok done.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
nicolasbrossard: I could only find one call to computeMd5Hash: in uploadNeuroDB/minc_insertion.pl. You updated the call made in that file so I would say everything is OK with respect to this API change.

christinerogers:
Tagging as Caveat for Existing projects given accompantying updates in dicom-archive-tools and MySQL (Loris repo)

PR #105, opened 2016-01-07
Making MINC Toolkit check better + Clarifications on the toolkits

Jkat:
No longer apt-get installing wrong MINC Toolkit.
Directing users to look at the BIC MINC site.
Misc toolkit clarifications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 # System Requirements
  * Perl
- * DICOM toolkit (Will be installed by Install Script if not already installed)
+ * DICOM toolkit
  * MINC tools (http://www.bic.mni.mcgill.ca/ServicesSoftware/MINC)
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,7 +1,7 @@
 # System Requirements
  * Perl
- * DICOM toolkit
- * MINC tools
+ * DICOM toolkit (Will be installed by Install Script if not already installed)
+ * MINC tools (http://www.bic.mni.mcgill.ca/ServicesSoftware/MINC)
 
 Note: For Ubuntu installations, DICOM toolkit and MINC toolkit are installed by the imaging install script (see step 3 below). This script will apt-get install these packages: minc-tools and dcmtk.
```
MounaSafiHarab: @Jkat, 

I think this note should be updated (remove that the MINC toolkit will be installed).

Jkat: Thanks! I've clarified it, hopefully it's better now.

christinerogers:
Adding caveat tags so that existing projects know we are recommending starting now a different installation of minc tools - in case the change in algorithms impacts file conversions and datasets.

PR #107, opened 2016-01-25
'Processed' renamed and set to '1' in tarchiveLoader

MounaSafiHarab:
- Renamed Flags "Processed" to "InsertionComplete" and "CurrentlyProcessed" to "Inserting" (since we are essentially performing insertion of images into the database)
- Changed 'Processed' to 'Inserting' when changed to '1' in ImagingUpload.pm and imaging_upload_file.pl
- Updated 'InsertionComplete' to 1 within the tarchiveLoader
- Set 'Inserting' back to 0 within the imaging_upload_file.pl
- To be used with pull request: https://github.com/aces/Loris/pull/1503


uploadNeuroDB/imaging_upload_file.pl
```diff
@@ -214,6 +218,7 @@
 ############### Run runTarchiveLoader###########################
 ################################################################
 $output = $imaging_upload->runTarchiveLoader();
+$imaging_upload->updateMRIUploadTable('Processing', 0);
```
mohadesz: Why set to false here?

MounaSafiHarab: Because at this point, processED is set to 1 in tarchiveLoader, so processING should be set back to 0

uploadNeuroDB/imaging_upload_file.pl
```diff
@@ -214,6 +218,7 @@
 ############### Run runTarchiveLoader###########################
 ################################################################
 $output = $imaging_upload->runTarchiveLoader();
+$imaging_upload->updateMRIUploadTable('Inserting', 0);
```
cmadjar: Why is Inserting set to 0 when running tarchiveLoader?

uploadNeuroDB/tarchiveLoader
```diff
@@ -574,6 +574,17 @@ if (!$valid_study) {
     print "No Mincs inserted \n \n";
     exit 9;
 }
+
+###############################################################
+### Set Processed to 1 in mri_upload table#####################
+###############################################################
+my $where = "WHERE TarchiveID=?";
+my $query = "UPDATE mri_upload SET InsertionComplete=1 ";
```
cmadjar: Should the Inserting flag be set back to 0? Not inserting anymore.

uploadNeuroDB/imaging_upload_file.pl
MounaSafiHarab: This is actually the imaging_upload_file.pl (not tarchiveLoader). Line 220 of this script takes you eventually to the tarchiveLoader, you rurn it, you return to this script, then on line 221 we set Inserting back to '0' (as tarchiveLoader is the last step of the insertion scripts).

uploadNeuroDB/tarchiveLoader
MounaSafiHarab: It can be done here at the same time as setting InsertionComplete = 1. I do it right at the return from this call (namely in the imaging_upload_file.pl; please see comment above), and that, irrespective of whether runTarchiveLoader() passed or failed.

uploadNeuroDB/imaging_upload_file_cronjob.pl
```diff
@@ -72,7 +73,9 @@
 ################################################################
 my $dbh = &NeuroDB::DBI::connect_to_db(@Settings::db);
 my @row=();
-my $query = "SELECT UploadID, UploadLocation FROM mri_upload WHERE Processed=0 AND (TarchiveID IS NULL AND number_of_mincInserted IS NULL)";
+my $query = "SELECT UploadID, UploadLocation FROM mri_upload 
+		WHERE Inserting <> 1 AND InsertionComplete <> 1 
+		AND (TarchiveID IS NULL AND number_of_mincInserted IS NULL)";
 print "\n" . $query . "\n";
```
cmadjar: Should use following syntax if want to code queries and be able to copy it easily into the mysql database when debugging:
(my $query = <<QUERY) =~ s/\n/ /gm;
SELECT UploadID, UploadLocation FROM mri_upload 
    WHERE Inserting <> 1 AND InsertionComplete <> 1 
        AND (TarchiveID IS NULL AND number_of_mincInserted IS NULL)";
QUERY

mohadesz: That's a good point but I believe we should be consistent none of the codes written for imaging-uploader uses '<QUERY'. And I personally prefer concatenation of the strings (i.e. "QUERY" . "QUERY")

PR #109, opened 2016-03-10
Fixes for the MRI pipeline when run in auto-launch mode.

nicolasbrossard:
- Fixes path problems in the environment file when the pipeline is run in auto-launch mode.
- Updates the install script to reflect the changes done in environment.

### DIFF
--- a/environment
+++ b/environment
@@ -1,4 +1,7 @@
-export PATH=/data/%PROJECT%/bin/mri:/data/%PROJECT%/bin/mri/uploadNeuroDB:/data/%PROJECT%/bin/mri/dicom-archive:$PATH
+export PATH=%MINC_TOOLKIT_DIR%/bin:/data/%PROJECT%/bin/mri:/data/%PROJECT%/bin/mri/uploadNeuroDB:/data/%PROJECT%/bin/mri/dicom-archive:$PATH
 export PERL5LIB=/data/%PROJECT%/bin/mri/uploadNeuroDB:$PERL5LIB
 export TMPDIR=/tmp
 export LORIS_CONFIG=/data/%PROJECT%/bin/mri/dicom-archive
+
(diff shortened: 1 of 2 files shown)


MounaSafiHarab:
@nicolasbrossard 
I get two errors:

1) if the upload path is NOT left blank (I ste it to /data/incoming/), auto launch gives the error that it cna not find the uploaded file (the file is there and the permissions look Ok for www-data)

2) as I leave the path blank, it upload to /tmp/, auto launch starts, but I get two errors (the first one you probably know about):
a) DBD::mysql::st execute failed: Column 'CreatingUser' cannot be null at /data/loris-MRI/bin/mri/dicom-archive/DICOM/DCMSUM.pm line 200.
b) DBD::mysql::st execute failed: Cannot add or update a child row: a foreign key constraint fails (`abcd_xyzw`.`tarchive_series`, CONSTRAINT `tarchive_series_ibfk_1` FOREIGN KEY (`TarchiveID`) REFERENCES `tarc
hive` (`TarchiveID`) ON DELETE CASCADE) at /data/loris-MRI/bin/mri/dicom-archive/DICOM/DCMSUM.pm line 285.

MounaSafiHarab:
@nicolasbrossard,

1) both errors are resolved. I tested with /tmp as incoming directory and /data/incoming/ and both work.
2) This error (in my earlier comment) was mainly due to the fact that there is a dicom_archive pull request (https://github.com/aces/dicom-archive-tools/pull/32) that has to be used together with this PR, and when this is done, the error disappears.

christinerogers:
@MounaSafiHarab @nicolasbrossard  -- What should Existing Projects do to take advantage of this change?    (If anything, please tag Caveat for Existing projects, and specify what the steps are)

PR #111, opened 2016-03-17
Removing the message "The file cannot be moved. Make sure the getIncomingDir is set" (exit 9 code)

MounaSafiHarab:
- This gets rid of the exit 9 code we all get "The file cannot be moved. Make sure the getIncomingDir config option is set"
- Removed the moveUploadedFile() and CleanUpTMPDir() from imaging_upload_file.pl, and replaced with a removal of the uploaded file cleanUpDataIncomingDir().
- Before removal, one check is performed to ensure the file that was used to run tarchiveLoader is in the tarchive directory
- Once PR104 is merged, the spool messages need to be updated with the 'Verbose' flag

### DIFF
--- a/DTIPrep/DTI/DTI.pm
+++ b/DTIPrep/DTI/DTI.pm
@@ -56,7 +56,7 @@ sub createOutputFolders{
                         $visit  . "/mri/processed/" .
                         substr(basename($protocol),0,-4);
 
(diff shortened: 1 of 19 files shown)


MounaSafiHarab:
will need to update the message spooling once the notification_spool pull request is merged

PR #112, opened 2016-03-18
Standardization of permissions for all MRI data files/dirs.

nicolasbrossard:
Since the MRI pipeline can now be run either from the command line or from the front-end (using the auto-launch), all data files/dirs have to be readable/writeable by user lorisadmin (command line) and www-data (front-end). Here's what I did to satisfy this requirement:

1. I modified the install script so that all files and directories under /data/$PROJ/data will be readable/writable by anyone who is a member of the Unix group lorisadmin. 
2. Any directory created by the MRI pipeline will have permission 770. 
3. User www-data should be part of the Unix group lorisadmin.

Caveat for existing projects:

Current projects should run these commands:

% sudo chgrp -R lorisadmin /data/$PROJ/data
% sudo chmod -R 2770 /data/$PROJ/data
% sudo chgrp -R lorisadmin /data/incoming
% sudo chmod -R 2770 /data/incoming

and also make sure user www-data is part of the Unix group lorisadmin

### DIFF
--- a/DTIPrep/DTI/DTI.pm
+++ b/DTIPrep/DTI/DTI.pm
@@ -56,7 +56,7 @@ sub createOutputFolders{
                         $visit  . "/mri/processed/" .
                         substr(basename($protocol),0,-4);
 
-    system("mkdir -p -m 755 $QC_out")   unless (-e $QC_out || !$runDTIPrep);
+    system("mkdir -p -m 770 $QC_out")   unless (-e $QC_out || !$runDTIPrep);
 
 
     return  ($QC_out) if (-e $QC_out);
--- a/MNI/FileUtilities.pm
+++ b/MNI/FileUtilities.pm
@@ -264,7 +264,7 @@ sub check_output_dirs
       }
       else                      # no file, no dangling link
       {
-	 if (! mkdir ($dir, 0755))
+	 if (! mkdir ($dir, 0770))
 	 {
 	    warn "couldn't create \"$dir\": $!\n";
 	    $num_err++;
@@ -412,7 +412,7 @@ sub check_output_path
 
       if (! (-e $partial || -l $partial))
       {
-	 unless (mkdir ($partial, 0755))
+	 unless (mkdir ($partial, 0770))
 	 {
 	    warn "\"$path\" not a writeable path: couldn't create \"$partial\": $!\n";
 	    return 0;
@@ -442,7 +442,7 @@ sub check_output_path
 
    if (! (-e $partial || -l $partial))
    {
-      unless (mkdir ($partial, 0755))
+      unless (mkdir ($partial, 0770))
       {
 	 warn "\"$path\" not a writeable path: couldn't create \"$partial\": $!\n";
 	 return 0;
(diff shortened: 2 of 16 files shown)


imaging_install_MacOSX.sh
```diff
@@ -37,23 +37,26 @@ mridir=`pwd`
...
-  sudo -S su $USER -c "mkdir -p /data/$PROJ/data/"
-  sudo -S su $USER -c "mkdir -p /data/$PROJ/data/trashbin"          #holds mincs that didn't match protocol
-  sudo -S su $USER -c "mkdir -p /data/$PROJ/data/tarchive"          #holds tared dicom-folder
-  sudo -S su $USER -c "mkdir -p /data/$PROJ/data/pic"               #holds jpegs generated for the MRI-browser
-  sudo -S su $USER -c "mkdir -p /data/$PROJ/data/logs"              #holds logs from pipeline script
-  sudo -S su $USER -c "mkdir -p /data/$PROJ/data/jiv"               #holds JIVs used for JIV viewer
-  sudo -S su $USER -c "mkdir -p /data/$PROJ/data/assembly"          #holds the MINC files
-  sudo -S su $USER -c "mkdir -p /data/$PROJ/data/batch_output"      #contains the result of the SGE (queue
-  sudo -S su $USER -c "mkdir -p $mridir/dicom-archive/.loris_mri"
+  sudo -S su $USER -c "mkdir -m 2770 -p /data/$PROJ/data/"
```
mohadesz: why does the /data/$PROJ/data need a sticky bit?

nicolasbrossard: Say you have this scenario:

Using auto-launch, you upload a scan than was done in 2015 and you haven't upload any 2015 scans. MRProcessingUtilities will have to create /data/<proj>/data/tarchive/2015 (and when it does, gid will be set to www-data since weas re using the auto-launch). Once the scan is uploaded, you insert another one (done in 2015) but using the pipeline on the cmd line this time, run by lorisadmin. An error will occur since lorisadmin cannot create files in directory 2015.

Adding the sticky bit to /data/$PROJ/data will solve this. Makes sense ?

DTIPrep/DTI/DTI.pm
```diff
@@ -56,7 +56,7 @@ sub createOutputFolders{
                         $visit  . "/mri/processed/" .
                         substr(basename($protocol),0,-4);
 
-    system("mkdir -p -m 755 $QC_out")   unless (-e $QC_out || !$runDTIPrep);
+    system("mkdir -p -m 770 $QC_out")   unless (-e $QC_out || !$runDTIPrep);
```
prioux: WTF ? Why invoke the external executable "mkdir" when there is already a mkdir function in PERL ?!?

MNI/FileUtilities.pm
```diff
@@ -264,7 +264,7 @@ sub check_output_dirs
       }
       else                      # no file, no dangling link
       {
-	 if (! mkdir ($dir, 0755))
+	 if (! mkdir ($dir, 0770))
```
prioux: Here's an example!!!

PR #116, opened 2016-04-14
fixed the directory name when doing chgrp during installation

MounaSafiHarab:


christinerogers:
Great fix - thanks @MounaSafiHarab 

This bug was included in the 15.10 release.

PR #117, opened 2016-04-21
Bypassing redundant checking on center name; check using ScannerID

MounaSafiHarab:
This is for the very unique case when the mri_protocol table has a center name (as opposed to ZZZZ or AAAA) and a scanner ID, AND the candidate gets scanned at an MRI site other than the recruiting site, the protocol will return unknown.
The check of center name AND scannerID is redundant, irrespective of the scenario above, but the proposed change should make it work for the scenario above.

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -398,7 +398,7 @@ Returns: Textual name of scan type
 
 sub identify_scan_db {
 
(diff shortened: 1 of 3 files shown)


MounaSafiHarab:
@PierreEMorin 
can you please test this thoroughly?

MounaSafiHarab:
This pull request was created specifically for @PierreEMorin 
in case where Candidates get scanned at another site that is not the recruiting site whereas the mri_protocol table has entries based on the recruiting site;
the changes to the code are minimal and in this pull request (if ever needed in the future for some projects/users).

PR #121, opened 2016-05-06
Modified the creation of tmp directories to be configurable.

cmadjar:
It will look into the environment file for the $TMPDIR variable to determine where the pipeline should create temporary directory for imaging extraction, conversion etc...


MounaSafiHarab:
@cmadjar,

This is a good feature! but I was just wondering why you need to make any code changes. Currently, without any code changes, I set the TEMPDIR to for example /data/test (assuming this directory exists), and the Loris-MRI code will be executing its extraction in this path (/data/test). I am just not sure why the code change?

cmadjar:
No need for it in the end.

PR #122, opened 2016-05-18
More informative success message upon insertion scripts completion

MounaSafiHarab:
- Added a message with the correct number of minc created and minc inserted into the database.
- A similar messaging clarification is done in the imaging_uploader in this Loris pull request:
https://github.com/aces/Loris/pull/1821
as per @nicolasbrossard feedback

### DIFF
--- a/uploadNeuroDB/imaging_upload_file.pl
+++ b/uploadNeuroDB/imaging_upload_file.pl
@@ -229,8 +229,6 @@ if ( !$output ) {
     print $message;
     exit 8;
 }
-$message = "\nThe insertion scripts have successfully completed\n";
-spool($message,'N', $notify_notsummary);
 
 ################################################################
 ### If we got this far, dicomTar and tarchiveLoader completed###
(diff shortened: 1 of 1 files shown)


MounaSafiHarab:
@nicolasbrossard, 
should any of the imaging uploader "success" message be changed? 
the added message will appear right before your final "success" message in the imaging uploader "Summary" and "Detailed" views.

nicolasbrossard:
Good idea. Instead of writing "Success" in the Progress column of the MRI
upload table, we could write "Success (20 out of 25)" or something like
this. What do you think ?

MounaSafiHarab:
@nicolasbrossard,
I agree with you re. the Progress column. Just did that in a Loris PR:
https://github.com/aces/Loris/pull/1821

uploadNeuroDB/imaging_upload_file.pl
```diff
@@ -280,6 +289,48 @@ sub getPnameUsingUploadID {
...
+  - Get the count of minc files created and inserted using the upload_id
+
+Arguments:
+  $upload_id: The Upload ID
+
+  Returns: $minc_created and $minc_inserted: count of minc created and inserted
+=cut
+
+
+sub getNumberMincFiles {
```
mohadesz: Maybe getNumberOfMincFiles would be more informative?

PR #124, opened 2016-05-26
From Tom Beaudry: mri_protocol table checking when the row has a regex entry

MounaSafiHarab:
If the mri_protocol table has more than one entry with a regex in the series_description column, all the files will be classified according to the scan type of the first entry in the mri_protocol table which has a regex entry. This pull request should fix this.


MounaSafiHarab:
I cleared the 16.1 milestone. This merged one was sent to master, so the decision is to re-send this same pull request to 16.0 and tag it accordingly.

PR #125, opened 2016-05-26
Update Loris config.xml to include the MIncToolsPath

MounaSafiHarab:
Related to this Loris pull request:
https://github.com/aces/Loris/pull/1843

### DIFF
--- a/imaging_install.sh
+++ b/imaging_install.sh
@@ -63,7 +63,7 @@ fi
  
 read -p "Enter the list of Site names (space separated) " site
 mridir=`pwd`
-#read -p "Enter Full Loris-code directory path "   lorisdir
+read -p "Enter Full Loris-code directory path "   lorisdir
 
 ################################################################################################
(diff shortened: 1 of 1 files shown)


christinerogers:
Hi @MounaSafiHarab - 
Can this be added/updated as a Config Setting in the database? This would solve several concerns: 
* This change assumes that the Loris-MRI and Loris (main repo) code scripts are installed on the same machine.  The code is housed in 2 separate repos because this co-location is not assumed to be the case (hence $lorisdir was commented out in the script a few years ago).
* If co-located, this script change would clobber instead of update the working project/config.xml file, overwriting its settings - I'm looking at the intake of the docs/ blank template and output to (overwriting) the existing (working) config file. 
> < $lorisdir/docs/config/config.xml > $lorisdir/project/config.xml

A database Config table setting would solve both aspects.

MounaSafiHarab:
close this pull request as the Readme by @christinerogers will be the better way to document this change

PR #130, opened 2016-06-06
jiv_path in the parameter_file table is more specific now; includes f…

MounaSafiHarab:
…ilename not just CandID


MounaSafiHarab:
it was decided that this addition is not necessary, so closing the pull request!

PR #131, opened 2016-06-07
jiv_path in parameter file is more specific; resend to 16p0

MounaSafiHarab:
This is a re-send of pull request 130:
https://github.com/aces/Loris-MRI/pull/130


MounaSafiHarab:
close it as I issued it from master

PR #133, opened 2016-06-09
Batch upload of scans for projects not using the front-end imaging uploader

MounaSafiHarab:
Run it as follows form where the Loris-MRI code base is:

$ ./batch_uploads_imageuploader -profile prod < imageuploader_list >log_batch_imageuploader.txt 2>&1

where imageuploader_list is a text file that has one scan details per line (example for 2 entries shown below); and each line (for each scan) consists of space delimited 1) the full path to the DICOM zipped scan, 2) Y or N depending on whether the scan is for a phantom or not, and 3) patient name following the PSCID_CandID_VisitLabel Loris convention for real candidates and left BLANK for phantoms 

An example of uploading 2 entries/scans to be uploaded:
/data/incoming/PSC0001_123457_V1.tar.gz  N  PSC0000_123456_V1
/data/incoming/Lego_Phantom_MNI_20140101.zip  Y

### DIFF
--- a/batch_uploads_imageuploader
+++ b/batch_uploads_imageuploader
@@ -1,0 +1,123 @@
+#!/usr/bin/perl -w
+use strict;
+use warnings;
+use File::Basename;
+
+#####Get config setting#######################################################
+# checking for profile settings
+if(-f "$ENV{LORIS_CONFIG}/.loris_mri/prod") {
+    { package Settings; do "$ENV{LORIS_CONFIG}/.loris_mri/prod" }
+} ##Possibly the script can exit if the prod doesn't exist
+#######################################################################################
+
+# define project space
+my ($debug, $verbose) = (0,1);
+my ($stdoutbase, $stderrbase) = ("$Settings::data_dir/batch_output/imuploadstdout.log", "$Settings::data_dir/batch_output/imuploadstderr.log");
+my $stdout = '';
+my $stderr = '';
+
+my @opt_table           = (
+    [ "Basic options", "section" ],
+    ["-verbose", "boolean", 1,   \$verbose, "Be verbose."]
+);
+
+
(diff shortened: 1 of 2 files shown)


christinerogers:
@MounaSafiHarab  This looks like a major change for existing projects, and information that should be easily available to New projects as well.  
Let's create/add tag "add to documentation" as it should definitely go in the wiki as well.

MounaSafiHarab:
@christinerogers 

This is a new feature that should not impact at all existing projects. imaging_upload_file.pl should run for existing projects without any change (the extra options added to the script are not required for existing projects, and are needed for those using batch_uploads_imageuploader which is a new script).

Bye yes, it should be documented and tested more thoroughly.

gluneau:
@MounaSafiHarab, please add -profile prod option to the example on how to run this tool.

MounaSafiHarab:
@gluneau,
 
thanks for the catch!

batch_uploads_imageuploader
```diff
@@ -0,0 +1,129 @@
...
+    if (($type ne '.tgz') && ($type ne '.tar.gz') && ($type ne '.zip')) {
+	print "The file on line $counter is not of type .tgz, tar.gz, or .zip and will not be processed\n";
+	exit(1);
+    }
+    if ($base ne $patientname) {
+	print "Make sure the patient name $patientname " .
+	      "matches the file name $base in $path\n";
+	exit(2);
+    }
+    if (($phantom eq '') || ($phantom ne 'N') || ($phantom eq 'Y')) {
```
gluneau: @MounaSafiHarab If you modify your logic for this line to:

`if (($phantom eq '') || (($phantom ne 'N') && ($phantom ne 'Y'))) `

the phantoms will get inserted.

PR #134, opened 2016-06-16
Creation of mkdir with correct permissions when user=apache

MounaSafiHarab:
Creation of some directories are now explicitely invoked so that permissions are set properly (to allow the front end user or lorisadmin to run the pipeline seamlessly)

### DIFF
--- a/imaging_install.sh
+++ b/imaging_install.sh
@@ -148,6 +148,9 @@ else
     read -p "Cannot find the apache group name for your installation. Please provide? " group
 fi
 
+# Making lorisadmin part of the apache group
+sudo usermod -a -G $group $USER
+
 #Setting group permissions to apache and group ID for all files/dirs under /data/$PROJ/data
 sudo chgrp $group -R /data/$PROJ/data/
 sudo chmod -R g+s /data/$PROJ/data/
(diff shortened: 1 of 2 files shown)


imaging_install.sh
```diff
@@ -148,6 +148,9 @@ else
     read -p "Cannot find the apache group name for your installation. Please provide? " group
 fi
 
+# Making lorisadmin part of the apache group
+sudo usermod -a -G $group $USER
+
```
MounaSafiHarab: @christinerogers ,
this was already in your Readme but the installation script did not do it.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -578,7 +578,23 @@ sub move_minc {
     ### figure out where to put the files ######################
     ############################################################
     $dir = $this->which_directory($subjectIDsref,$data_dir);
-    `mkdir -p -m 770 $dir/native`;
+
```
MounaSafiHarab: I know the next changes are not pretty but the line we had before, albeit compact, only sets the permissions properly on the last directory only (native in this case). the -p does not propagate the permissions set in -m option to the earlier subdirectories (hence CandID/ and /VisitLabel/ and mri/ all had permissions other than 770).
Below I invoke mkdir with -m option for each of the directories in the hierarchy

PR #135, opened 2016-06-16
MRI Readme updates re environment, usergroups, version

christinerogers:
Since we're now on release branches, it's not enough to just clone the repo, which will get you the master branch.  
Users attempting to install Loris-MRI 16.0 must clone the v16.0.0 tag specifically.
Updated Readme to reflect this.

Also removed spurious minc path export instruction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ See [aces/Loris README.md](https://github.com/aces/loris) for further informatio
 
   If the imaging install script reports errors in creating directories (due to /data/ mount permissions), manually execute mkdir/chmod/chown commands starting at [imaging_install.sh:L90](https://github.com/aces/Loris-MRI/blob/16.0-dev/imaging_install.sh#L90)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,7 +23,7 @@ See [aces/Loris README.md](https://github.com/aces/loris) for further informatio
    sudo mkdir -p /data/$projectname/bin/mri
    sudo chown -R lorisadmin:lorisadmin /data/$projectname
    cd /data/$projectname/bin
-   git clone https://github.com/aces/Loris-MRI.git mri
+   git clone -b 16.0-dev https://github.com/aces/Loris-MRI.git mri
    ```
```
MounaSafiHarab: @christinerogers 

why are we cloning the development branch? shouldn't this be 16.0?

christinerogers: Updated: now cloning from the (future) release tag, instead of the branch

PR #140, opened 2016-07-05
new converter can use dcm2mnc command structure

gluneau:
This change allows project to modify their prod files to use a different version of dcm2mnc and still use the same command format. For example dcm2mnc4ng does match the proper command string now.


MounaSafiHarab:
@gluneau ,

Useful for running dcm2mnc using a modified binary of dcm2mnc...
code changes make the script more general and should not break existing projects.

PR #141, opened 2016-07-05
allow the insertion of violated scans

gluneau:
This command will insert a violated scan since providing the acquisition protocol bypasses the protocol checks.  If you also need to bypass the extra_file_checks use bypass_extra_file_checks
Please note the -create_minc_pics option as well since it is usually created by the tarchiveloader script.

```
minc_insertion.pl -acquisition_protocol t2w -bypass_extra_file_checks -create_minc_pics -profile prod -globLocation -force  -tarchivePath /Users/xxxproject/dataTransfer/library/2009/DCM_2009-09-25_project_20110214_185904581.tar -mincPath /data/project/data/trashbin/TarLoad-3-34-pVzGC5/xxx0067_703739_v12_20090925_222403_18e1_mri.mnc
```

### DIFF
--- a/uploadNeuroDB/minc_insertion.pl
+++ b/uploadNeuroDB/minc_insertion.pl
@@ -470,7 +470,11 @@ unless ($no_nii) {
 ################################################################
 if ($create_minc_pics) {
     print "\nCreating Minc Pics\n" if $verbose;
-    NeuroDB::MRI::make_minc_pics(\$tarchiveInfo{TarchiveID});
+    NeuroDB::MRI::make_minc_pics(\$dbh,
+                                  $tarchiveInfo{TarchiveID},
+                                  $profile,
+                                  1);
+    # Set minFileID to 1: minFileID $row[1] & maxFileID $row[1]
 }
 
 ################################################################
--- a/uploadNeuroDB/minc_insertion.pl
+++ b/uploadNeuroDB/minc_insertion.pl
@@ -46,6 +46,7 @@ my $no_jiv      = 0;           # Should bet set to 1, if jivs should not be
 my $NewScanner  = 1;           # This should be the default unless you are a 
                                # control freak
 my $xlog        = 0;           # default should be 0
+my $bypass_extra_file_checks=0;# If you need to bypass the extra_file_checks, set to 1.
 my $acquisitionProtocol;       # Specify the acquisition Protocol also bypasses the checks
 my $acquisitionProtocolID;     # acquisition Protocol id
 my @checks      = ();          # Initialise the array
@@ -105,6 +106,9 @@ my @opt_table = (
 
                  ["-create_minc_pics", "boolean", 1, \$create_minc_pics,
                   "Creates the minc pics."],
+
+                 ["-bypass_extra_file_checks", "boolean", 1, \$bypass_extra_file_checks,
+                  "Bypasses extra_file_checks."],
 );
 
 
@@ -395,7 +399,7 @@ $file->setFileData('Caveat', 0);
       \%tarchiveInfo,$center_name,
       $minc,
       $acquisitionProtocol,
-      $force
+      $bypass_extra_file_checks
     );
 
 
--- a/uploadNeuroDB/minc_insertion.pl
+++ b/uploadNeuroDB/minc_insertion.pl
@@ -388,15 +388,16 @@ $file->setFileData('Caveat', 0);
 ################################################################
 ## Get acquisition protocol (identify the volume) ##############
 ################################################################
-if(!defined($acquisitionProtocol)) {
-  ($acquisitionProtocol,$acquisitionProtocolID,@checks)
-    = $utility->getAcquisitionProtocol(
-        $file,
-        $subjectIDsref,
-        \%tarchiveInfo,$center_name,
-        $minc
-      );
-}
+($acquisitionProtocol,$acquisitionProtocolID,@checks)
+  = $utility->getAcquisitionProtocol(
+      $file,
+      $subjectIDsref,
+      \%tarchiveInfo,$center_name,
+      $minc,
+      $acquisitionProtocol,
+      $force
+    );
+
 
 if($acquisitionProtocol =~ /unknown/) {
    $message = "\n  --> The minc file cannot be registered ".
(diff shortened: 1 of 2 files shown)


uploadNeuroDB/minc_insertion.pl
```diff
@@ -455,6 +464,40 @@
     NeuroDB::MRI::make_nii(\$file, $data_dir);
 }
 
+############################################################
+############# Create minc-pics #############################
+############################################################
+unless ($create_minc_pics)
+{
```
MounaSafiHarab: @gluneau,

you want to create pics if the user spoecifies this option; this will do the opposite
(I just confirmed when running the command line: pics generated withOUT specifying -create_minc_pics)

MounaSafiHarab:
@gluneau,

when running the command we should not specify -create_minc_pics 1; it should simply be -create_minc_pics (without the 1)

uploadNeuroDB/minc_insertion.pl
```diff
@@ -380,13 +387,15 @@
 ################################################################
 ## Get acquisition protocol (identify the volume) ##############
 ################################################################
-my ($acquisitionProtocol,$acquisitionProtocolID,@checks)
-  = $utility->getAcquisitionProtocol(
+if(!defined($acquisitionProtocol)) {
+  my ($acquisitionProtocol,$acquisitionProtocolID,@checks)
```
MounaSafiHarab: @gluneau,

$acquisitionProtocol is no longer accessible for the next check $utility->registerScanIntoDB() (line 416)

uploadNeuroDB/minc_insertion.pl
```diff
@@ -96,7 +100,11 @@
                  ["General options","section"],
                  ["-verbose", "boolean", 1, \$verbose, "Be verbose."],
            
+                 ["-acquisition_protocol","string", 1, \$acquisitionProtocol,
+                  "Suggest the acquisition protocol to use."],
 
+                 ["-create_minc_pics", "boolean", 1, \$create_minc_pics,
+                  "Use this if invoked directly witouth tarchiveloader."],
```
driusan: The help should describe what the flag does, not when to use it. ("without" is also spelled wrong.)

uploadNeuroDB/minc_insertion.pl
```diff
@@ -455,6 +465,40 @@
     NeuroDB::MRI::make_nii(\$file, $data_dir);
 }
 
+############################################################
+############# Create minc-pics #############################
+############################################################
+if ($create_minc_pics)
```
MounaSafiHarab: @gluneau,

this can be taken outside to a utility file, this way tarchiveLoader and minc_insertion can reuse the same function

uploadNeuroDB/minc_insertion.pl
```diff
@@ -46,6 +46,10 @@
 my $NewScanner  = 1;           # This should be the default unless you are a 
                                # control freak
 my $xlog        = 0;           # default should be 0
+my $acquisitionProtocol;       # Specify the acquisition Protocol also bypasses the checks
+my $acquisitionProtocolID;     # acquisition Protocol id
+my @checks      = ();          # Initialise the array
+my $create_minc_pics    = 0;   # Default is 0, set the option to overide.
```
on line 63, can you please fix the typo that was already there (casic should be basic)!

MounaSafiHarab:
@gluneau;

I think:
1) you forgot to add the MRI.pm file to your last commit (which has the main create_minc_pics()); 
and 
2) tarchiveLoader should be updated as well by removing the call to mass_pic.pl

gluneau:
@MounaSafiHarab 

1) The MRI.pm call was already present in the file see other make_ function calls. (line 16)

2) The tarchieveLoader modification is in this request: https://github.com/aces/Loris-MRI/pull/149

MounaSafiHarab:
@gluneau 

This has to be tested together with the changes in: https://github.com/aces/Loris-MRI/pull/149
otherwise "Undefined subroutine &NeuroDB::MRI::make_minc_pics called at uploadNeuroDB/minc_insertion.pl line 473."

Otherwise, it looks good.

I would modify the comment as to how to run it as follows:
in the Loris-MRI code base, i.e. in /data/$PROJ/bin/mri/ run as follows:
$ uploadNeuroDB/minc_insertion.pl -acquisition_protocol t2w -create_minc_pics -profile prod -globLocation -force  -tarchivePath /Users/xxxproject/dataTransfer/library/2009/DCM_2009-09-25_project_20110214_185904581.tar -mincPath /data/project/data/trashbin/TarLoad-3-34-pVzGC5/xxx0067_703739_v12_20090925_222403_18e1_mri.mnc

gluneau:
@MounaSafiHarab running instruction updated to not have the perl program in front.

driusan:
Acquisition_protocol as a parameter should only specify the acqusition_protocol, not also bypass the protocol enforcement.

There should be two flags:
1. -acquisition_protocol (to specify the acquisition protocol if the mri_protocol table can't determine it)
2. -force (to skip the checks and insert it regardless of the study protocol)

MounaSafiHarab:
@driusan @gluneau 

the -force flag is already used when checking if the isTarchiveValidated flag is set in the mri_upload table, so I suggested to Greg to use another flag that is indicative of forcing specifically no extra_file_check. I assume you are Okay with this @driusan ?

driusan:
Do the flags get passed between the scripts? If not, I'm not sure it would be a problem and `-force` is a pretty standard name across different software packages. But if there's a conflict or you think it would be confusing, I have no problem using a different name (`-forceinsert`, maybe?)

MounaSafiHarab:
Currently our imaging_upload_file.pl which runs all the insertion scripts calls tarchiveLoader withOUT the -force option. So . As we have it now, the -force is used to bypass tarchive validation checks in tarchiveLoader and bypass the isTarchiveValidated flag check in mri_upload table in the minc_insertion script. 
I think we have two options:
1) If it is to be used as is (i.e. -force), then a) the help section which describes the use of this flag needs to be updated to say that we are bypassing tarchiveValidation and extra_file_checks(), and b) projects that might have customized this flag for their study need to be made aware of this
2) Use a flag name that is more specific to what we are doing: namely bypassing extra_file_checks() (as suggested by @gluneau) or -forceinsert (as syuggested by @driusan)

I personally prefer option (2)! it is more flags, but it is clearer I think!

MounaSafiHarab:
@gluneau 

tested it on a file that would violate the mri_protocol_checks entry, with and without the -bypass_extra_file_checks flag, and it seems to be NOT inserting without the flag and inserting with the flag on. So I think this is all good now.

PR #144, opened 2016-07-19
Readme for project directory permissions

MounaSafiHarab:

PR #147, opened 2016-08-01
Proper permission in installation2

MounaSafiHarab:
I think this is better than adding a note (as I did in this pull request: https://github.com/aces/Loris-MRI/pull/144) because:

1) We change group in the installation on subdirectories of /data/$PROJ/ , so we might as well do it in the installation script as well to be consistent (as opposed to adding a note, and do it on all subdirectories, not just some; more specifically the bin/mri/ path)
2) Most importantly, changes to the group ownership should happen on the prod file after its creation 

NOTE: The note I added in PR144 is more appropriate if I added it after point 4. or 5. (after having run .imaging_install.sh); but in all cases, it might be a good note to keep somewhere in the Wiki perhaps as a way of debugging permission-related errors

***
MAC install updates in this pull request are UNTESTED (Mac OSX no longer supported as of 15.10 but incorporated changes from PR #112, #129, #134 here
***

### DIFF
--- a/imaging_install.sh
+++ b/imaging_install.sh
@@ -151,12 +151,10 @@ fi
 # Making lorisadmin part of the apache group
 sudo usermod -a -G $group $USER
 
-#Setting group permissions to apache and group ID for all files/dirs under /data/$PROJ/data
-sudo chgrp $group -R /data/$PROJ/data/
+#Setting group permissions and group ID for all files/dirs under /data/$PROJ/data
 sudo chmod -R g+s /data/$PROJ/data/
 
-#Setting group permissions to apache and group ID for all files/dirs under /data/$PROJ/incoming
-sudo chgrp $group -R /data/incoming/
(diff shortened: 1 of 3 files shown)


imaging_install.sh
```diff
@@ -148,15 +138,25 @@ else
...
-#Setting group permissions to apache and group ID for all files/dirs under /data/$PROJ/data
-sudo chgrp $group -R /data/$PROJ/data/
+#Setting group permissions for all files/dirs under /data/$PROJ/ now that the prod file is created
+sudo chgrp $group -R /data/$PROJ/
+
+#Setting group ID for all files/dirs under /data/$PROJ/data
 sudo chmod -R g+s /data/$PROJ/data/
 
-#Setting group permissions to apache and group ID for all files/dirs under /data/$PROJ/incoming
-sudo chgrp $group -R /data/incoming/
```
driusan: I don't think this is supposed to be removed.

PR #148, opened 2016-08-01
Adding mnc2nii version to the parameter_file table (Redmine #5959 )

MounaSafiHarab:
- Running the insertion scripts form the command line now prints:
mnc2nii -version
program: 2.3.01
libminc: 2.3.01
netcdf : 4.3.3 of Feb  9 2016 13:43:57 $
HDF5   : 1.8.15

- Getting the same info from the database can be obtained by running the following mysql query :
mysql> select * from parameter_file as pf where pf.ParameterTypeID=(select pt.ParameterTypeID from parameter_type as pt where pt.Name = 'mnc2nii_version');

- Note that the version of the minc (Value: 2.3.01) is already in the database when querying for: 
where pt.Name = ':minc_version'

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -1203,8 +1203,14 @@ sub make_nii {
                     $data_dir . "/" . $nifti;
     system($m2n_cmd);
 
+    #  write mnc2nii version into parameter_file 
+    my $m2nv_cmd = "mnc2nii -version";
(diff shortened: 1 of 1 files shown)


MounaSafiHarab:
this is sent to 16.1 but the milestone is 17.0 as the 17.0-dev branch is not created yet.

MounaSafiHarab:
based on the imaging meeting of August 4th, 2016; this pull request and the associated task will be closed (we do store the minc_version used for dcm2mnc conversion in the database)

PR #149, opened 2016-08-01
Move minc pics creation to MRI.pm

gluneau:
This proto function was moved out of tarchiveLoader and into MRI.pm because it will be used also in minc_insertion.pl

`uploadNeuroDB/tarchiveLoader -globLocation -profile prod ../../../../xxx_test/DCM_2016-02-17_ImagingUpload-11-34-xxxxxx.tar`

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -1216,7 +1216,7 @@ sub make_minc_pics {
     my $output = undef;
     my @row = $sth->fetchrow_array();
     if (@row) {
-        $script = "mass_pic.pl -minFileID $minFileID -maxFileID $row[1] ".
+        $script = "mass_pic.pl -minFileID $row[$minFileID] -maxFileID $row[1] ".
                      "-profile $profile";
         print "Running mass_pic as follows : " .$script . "\n";
 
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -1201,6 +1201,36 @@ sub make_nii {
 }
 
 =pod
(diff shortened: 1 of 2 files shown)


uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -1201,6 +1201,36 @@ sub make_nii {
...
+=cut
+sub make_minc_pics {
+    my ($dbhr, $TarchiveSource, $profile, $minFileID) = @_;
+    my $where = "WHERE TarchiveSource = ? ";
+    my $query = "SELECT Min(FileID) AS min, Max(FileID) as max FROM files ";
+    $query    = $query . $where;
+    my $sth   = $${dbhr}->prepare($query);
+    $sth->execute($TarchiveSource);
+    print "TarchiveSource is " . $TarchiveSource . "\n";
+
```
MounaSafiHarab: @gluneau 

I think the print statement was added for debugging purposes (original file did not have it)

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -1201,6 +1201,36 @@ sub make_nii {
...
+    my $script = undef;
+    my $output = undef;
+    my @row = $sth->fetchrow_array();
+    if (@row) {
+        $script = "mass_pic.pl -minFileID $row[$minFileID] -maxFileID $row[1] ".
+                     "-profile $profile";
+        print "Running mass_pic as follows : " .$script . "\n";
+
```
MounaSafiHarab: @gluneau 

1) same comment re. "print" statement
2) original script had -verbose option appended to the $script command if $verbose is set (you will probably need to pass this variable along with the function call), so it is good to keep this so that it is consistent with what we had before

PR #150, opened 2016-08-10
More informative success message upon insertion scripts completion - …

MounaSafiHarab:
…See PR #122 which is in master but not in 16.0.


christinerogers:
💯

PR #145, opened 2016-07-20
Allow identifier to be PatientID or PatientName

spaiva:
- Created $identifier variable to allow parsing for PatientID or PatientName
- No longer using $lookupCenterNameUsing
- Regular Pattern of Patient ID coming from Prod file. 
- Dependencies: Dicom-Archive [PR#38](https://github.com/aces/dicom-archive-tools/pull/38)

PR #152, opened 2016-08-10
Allow identifier to be PatientID or PatientName

spaiva:
(Previously from PR #145) 
- Created `$identifier` variable to allow parsing for **PatientID** or **PatientName**
- With this `$identifier` variable, we no longer need `$lookupCenterNameUsing`
- Regular Pattern (`$PatientID_regex`) of Patient ID is configurable on ProfileTemplate. 
- Removed duplicate method `getSubjectIDs()` from Profile Template, using the one from MRI.pm instead, as it is more robust. 
- Dependencies: Dicom-Archive [PR#38](https://github.com/aces/dicom-archive-tools/pull/38)


uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -975,30 +954,30 @@ sub createNewCandID {
...
     }
 
-    ###########################################################################	
+    ###########################################################################
     ###Get the centerID using the PSCID########################################
     ###########################################################################
 
     #extract the PSCID from $patientName
-    $subjectIDsref = getSubjectIDs($patientName,null,$dbhr);
+    $subjectIDsref = &Settings::getSubjectIDs($identifier,null,$dbhr);
```
cmadjar: I am not sure to follow. The function getSubjectIDs is removed from the profile but still called from the profile here?

cmadjar: Forget it... I just realize I was not looking at the profile template.. Haha. All good :)

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -50,35 +50,14 @@ $VERSION = 0.2;
...
-	$subjectID{'CandID'} = my_trim(getScannerCandID($scannerID, $dbhr));
-	$subjectID{'visitLabel'} = my_trim($patientName);
-# subject data       	
-# old versions of this
-# qnts /([A-Z-]{3,4}\s+\d+)_(\d+)_([^_ ]+)/) or nihpd =~ /(\w{3}\d+)_(\d+)_([^_ ]+)/)
-    } elsif ($patientName =~ /([^_]+)_(\d+)_([^_ ]+)/) {
-	$subjectID{'PSCID'} = my_trim($1);
-	$subjectID{'CandID'} = my_trim($2);
-	$subjectID{'visitLabel'} = my_trim($3);
-	if(!subjectIDIsValid($subjectID{'CandID'}, $subjectID{'PSCID'}, $subjectID{'visitLabel'}, $dbhr)) {
```
MounaSafiHarab: @spaiva, @cmadjar:
We just need to be careful that this duplicate getSubjectID() function that got removed did not also remove some additional functionalities... for example subjectIDisValid() which itself also calls other functions... this is not used in the getSubjectID() in the prod file. So we need to investigate this further.

spaiva: I agree. I will look into this.

MounaSafiHarab:
@spaiva @cmadjar 

I think you need to update also uploadNeuroDB/registerFile.pl because there is a call to getSubjectID().

spaiva:
@MounaSafiHarab Sounds good. Thanks!

MounaSafiHarab:
Move to 17.0

MounaSafiHarab:
@spaiva @cmadjar 
One more thing that occurred to me now, the pipeline calls IsCandidateInfoValid() which greps the DICOM header for the PatientName field in cases of non-phantom entries to validate this field being PSCID_DCCID_VisitLabel, but since now the study can choose PatientName or PatientID as the identifier, then this function should also be made configurable as to which DICOM header to validate.

MounaSafiHarab:
@spaiva 
I will set the milestone to 17.1 to be in sync with redmine I guess.

spaiva:
Ready for testing @MounaSafiHarab and @cmadjar. Thanks so much for your time.

MounaSafiHarab:
I added the needs work label just to make sure that the changes are still valid and do not conflict with other requests that have been merged, since it has been a while (I am no longer sure which release this is going to)

PR #153, opened 2016-08-17
mysql error on duplicate entries

MounaSafiHarab:
This pull request re-computes the SNR basing it on setParameter() and getParameter() built-in functions... It also checks if the new SNR is different from the old/stored SNR (happening in the case where the noise_estimate algorithm from the MincToolKit is updated) and stores the new SNR value.

(The computeSNR() in the 16.1 branch could throw an exception (of duplicate entries; If the file already had an SNR entry, i.e. FileID, ParameterTypeID exists, and we re-run tarchiveLoader for example).

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
+++ b/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
@@ -1219,7 +1219,7 @@ sub computeSNR {
 		    $paramID = $sth->fetchrow_array;
  		}
 
-                $query = "INSERT INTO parameter_file SET Value=?, ".
+                $query = "INSERT IGNORE INTO parameter_file SET Value=?, ".
 			 "FileID=?, ParameterTypeID=?";
                 if ($this->{debug}) {
                     print $query . "\n";
@@ -1231,7 +1231,7 @@ sub computeSNR {
 		$this->spool($message, 'N', $upload_id, $notify_detailed);
             }
             else {
-                $message = "The SNR was not be computed for $base. ".
+                $message = "The SNR can not be computed for $base. ".
                            "Either the getSNRModalities is not defined in your ".
                            "$profile file, or the imaging modality does not ".
                            "support SNR computation. \n";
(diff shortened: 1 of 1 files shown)


MounaSafiHarab:
@driusan,

I took the liberty to assign to you as this is straightforward. But feel free to un-assign if you don't have time!

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1213,7 +1213,7 @@ sub computeSNR {
 		    $paramID = $sth->fetchrow_array;
  		}
 
-                $query = "INSERT INTO parameter_file SET Value=?, ".
+                $query = "INSERT IGNORE INTO parameter_file SET Value=?, ".
 			 "FileID=?, ParameterTypeID=?";
```
MounaSafiHarab: replace into
so it inserts the new updated value;

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1192,52 +1192,36 @@ sub computeSNR {
...
+            && Settings::getSNRModalities($base)) {
+                $cmd = "noise_estimate --snr $fullpath";
+                $SNR = `$cmd`;
+                $SNR =~ s/\n//g;
+                print "$cmd \n" if ($this->{verbose});
+                print "SNR is: $SNR \n" if ($this->{verbose});
+                my $file = NeuroDB::File->new($this->{dbhr});
+                $file->loadFile($fileID);
+                $SNR_old = $file->getParameter('SNR');
+                if ((defined ($SNR_old)) && ($SNR_old != $SNR)) {
```
gluneau: SNR is:  
Argument "" isn't numeric in numeric ne (!=) at /data/ibis/bin/mri/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm 

May need to break it into two if.

MounaSafiHarab: @gluneau 
should be fixed now in the last commit (changed != to ne)

PR #155, opened 2016-09-12
Command Line Tool to obtain FileID of inserted mincs using a lists of SeriesUID as STDIN

gluneau:
This tool will help figure out if the dicom files you have are inserted in loris.  

Example of use:
`for i in $(find -type f); do dcmdump $i | grep SeriesInstanceUID; done | sort | uniq | cut -d "[" -f2 | cut -d "]" -f1 | /data/ibis/bin/mri/tools/seriesuid2fileid`

image

### DIFF
--- a/tools/seriesuid2fileid
+++ b/tools/seriesuid2fileid
@@ -1,9 +1,37 @@
 #!/usr/bin/perl
 use strict;
 use warnings;
+no warnings 'once';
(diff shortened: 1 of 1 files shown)


spaiva:
@gluneau Tested it. Very good :)

spaiva:
@gluneau Ready to be tested?

PR #158, opened 2016-09-20
better error messages to help find why scans do not go in

gluneau:
Explains what is happening when errors happens

### DIFF
--- a/uploadNeuroDB/minc_insertion.pl
+++ b/uploadNeuroDB/minc_insertion.pl
@@ -405,7 +405,8 @@ $file->setFileData('Caveat', 0);
 
 if($acquisitionProtocol =~ /unknown/) {
    $message = "\n  --> The minc file cannot be registered ".
-		"since the AcquisitionProtocol is unknown \n";
+		"since the AcquisitionProtocol is unknown ".
+                "or it did not pass the extra_file_checks\n";
 
    print LOG $message;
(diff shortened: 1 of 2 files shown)


uploadNeuroDB/minc_insertion.pl
```diff
@@ -405,7 +405,8 @@
 
 if($acquisitionProtocol =~ /unknown/) {
    $message = "\n  --> The minc file cannot be registered ".
-		"since the AcquisitionProtocol is unknown \n";
+		"since the AcquisitionProtocol is unknown ".
+                "or it did not pass the extra_file_checks\n";
```
If I remember correctly the error you were getting (acquisition protocol does not exist in prod file), I believe this extra message should have been appended to the message on line 432

uploadNeuroDB/minc_insertion.pl
```diff
@@ -430,7 +430,8 @@
    ) {
    $message = "\n  --> The minc file cannot be registered ".
                 "since $acquisitionProtocol ".
-                "does not exist in $profile \n";
+                "does not exist in $profile ". 
+                "or it did not pass the extra_file_checks\n";
```

looks good to me. This puts it in line with the logic in the function registerScanIntoDB().

It would have been better to perhaps split the message as follows:
if ($checks->[0] !~ /exclude/) {keep the original message before your changes}
else { add the new part of your message}

But in all cases, your change is necessary as the initial message was misleading when the acquisition was in the prod file but extra_file_checks failed.

PR #161, opened 2016-10-06
To_log set to 1 for determineScannerID

MounaSafiHarab:
To be in synch with determinePSC() and determineSubjectID()
which write to log in tarchive_validation and NOT in minc_insertion and tarchiveLoader. I think the same should apply to determineScannerID().


samirdas:
Mouna is merging

PR #163, opened 2016-10-13
Readme and install script tweaks for 17.0

christinerogers:
Updating version numbers, plus: 

Loris-MRI Readme updates: 
* more explicit mention for CentOS of how DICOM toolkit can be installed

imaging_install.sh:
* will no longer attempt to install (via apt-get) DICOM toolkit when run on CentOS
As a result, this script should no longer abort when trying to apt-get install the DICOM toolkit, leaving the user to manually execute the rest of the code.  For the first time, CentOS installs will actually run the last half of the script. 

Mac install script:  added a note that it's not supported

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-This Readme covers release 16.1 of the LORIS Imaging Insertion Pipeline for Ubuntu or CentOS systems
+This Readme covers release 17.0 of the LORIS Imaging Insertion Pipeline for Ubuntu or CentOS systems
 
-This repo accompanies the [LORIS neuroimaging data platform main repo](https://github.com/aces/Loris/releases)</b>, release 16.1.*.<br>
+This repo accompanies the [LORIS neuroimaging data platform main repo](https://github.com/aces/Loris/releases)</b>, release 17.0.*.<br>
 For documentation and detailed setup information, please see the [LORIS wiki](https://github.com/aces/Loris/wiki/Imaging-Database)</b>.
 
 This repo can be installed on either the same VM as the main LORIS codebase, or on a different machine such as a designated fileserver where large imaging filesets are to be stored. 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -13,6 +13,7 @@ This repo can be installed on either the same VM as the main LORIS codebase, or
 On <u>Ubuntu</u>, DICOM toolkit will be installed by the imaging install script (step 4 below). This script will _apt-get install dcmtk_.
```
MounaSafiHarab: @christinerogers 
this is minor, but can we swap steps 4 and 3 in the system requirements? I think step 3 should come before step 4.

README.md
```diff
@@ -13,6 +13,7 @@ This repo can be installed on either the same VM as the main LORIS codebase, or
 On <u>Ubuntu</u>, DICOM toolkit will be installed by the imaging install script (step 4 below). This script will _apt-get install dcmtk_.   
 
 For <u>CentOS</u>: Dependency installation notes are included in the [LORIS wiki](https://github.com/aces/Loris/wiki/Imaging-Database) Imaging Setup page, Section 1 (installing codebase)</b>.
+As of the release date, this includes a transcript for [CentOS installation](https://github.com/aces/Loris/wiki/CentOS-Imaging-installation-transcript) and notes on dependencies including [DICOM toolkit](https://github.com/aces/Loris/wiki/CentOS-Imaging-installation-transcript#7-install-dicom-toolkit).
```
MounaSafiHarab: It is greta to add this here.
My only comment is as I go to the Wiki, the DICOM section, I would not keep the two lines on _export_ statement or would make it clearer that below are two possible examples

MounaSafiHarab:
@christinerogers 

from before (i..e not from your pull request), but I just saw this: line 43 in the Readme.md; a small formatting issue with _imagemagick_

README.md
christinerogers: Is there no need to install Minc toolkit before the imaging install script runs?

README.md
christinerogers: noted.  I'll clarify this in the wiki.

README.md
MounaSafiHarab: @christinerogers 
I only meant in the listing (the 3 bullets you have), just list MINC toolkit before DICOM toolkit; i.e. swapping lines 10 and 11, that's all.

samirdas:
to be tested thoroughly on Ubuntu and CentOS

PR #164, opened 2016-10-14
Uncommented an exit statement that was commented by mistake...

cmadjar:
An exit statement was commented by mistake during testing of a pull request in October 2015... Commenting this exit prevailed from quitting the minc_insertion.pl pipeline in case the file was already registered therefore allowing to insert the same file several times in the files table. 

This bug has been present since October 2015. OOOOOOOPSIE


MounaSafiHarab:
@cmadjar 

it is all good now. I no longer get duplicate entries of the same file in the files table/imaging browser with exit 8 UNcommented out.

PR #165, opened 2016-10-14
When tarchiveID exists, instruct the user how to launch tarchiveLoader at the command line: Redmine 11157

MounaSafiHarab:
As discussed in the imaging meeting, the user is now told what command to run to launch tarchiveLoader if the tarchiveID already exists (i.e. dicomTar.pl has completed).

### DIFF
--- a/uploadNeuroDB/NeuroDB/ImagingUpload.pm
+++ b/uploadNeuroDB/NeuroDB/ImagingUpload.pm
@@ -138,9 +138,9 @@ sub IsCandidateInfoValid {
     }
 
     ############################################################
-    ####Check to see if the scan has been ran ##################
+    ####Check to see if the scan has been run ##################
     ####if the tarchiveid or the number_of_mincCreated is set ##
-    ####it means that has already been ran. ####################
+    ####it means that has already been run. ####################
     ####So the user can continue the insertion by running ######
     ####tarchiveLoader exactly as the error message indicates ##
     ############################################################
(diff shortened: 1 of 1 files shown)


gluneau:
It works like it should:


```
imaging_upload_file.pl -profile prod -verbose -upload_id 371 /tmp/JedZFk/SEA1042_152174_V24.zip  

Spool message is: 
The Scan for the uploadID 371 has already been ran with tarchiveID: 5000184. 
To continue with the rest of the insertion pipeline, please run tarchiveLoader from a terminal as follows: /data/ibis/bin/mri/uploadNeuroDB/tarchiveLoader -globLocation -profile prod /data/ibis/data/tarchive/2016/DCM_2016-08-13_ImagingUpload-15-16-u7PhbY.tar -verbose
 

The candidate info validation has failed
```

christinerogers:
Looks great!  I noticed a small typo: "has already been ran with" (ran->run).

Would it help to specify how the candidate info validation failed - the nature of the error?

MounaSafiHarab:
@christinerogers 

typo fixed; thanks a lot in the second commit :)
as to the message, I think it is self explanatory: when we validate candidate we do:
1) tarchive not run before
2) files are all type dicom
3) patient name in mri_upload matches dicom header

any of the errors above will cause the candidate info validation to fail. In this case, it is step 1) that failed which is why you get both messages (that tarchive has been run and therefore the candidate info validation has failed).

PR #166, opened 2016-10-21
make sure the session is active

gluneau:
Session can be inactive, but should not be chosen for insertion

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -172,7 +172,7 @@ sub getSessionID {
     my $dbh = $$dbhr;
     
 # find a matching timepoint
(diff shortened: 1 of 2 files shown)


MounaSafiHarab:
@gluneau 

In the MRIProcessingUtility.pm, there are lines 92 (lookupNextVisitLabel) and 513 (in the update_mri_acquisition_dates), shouldn;t these also be set to search/update session with Active='Y'?

gluneau:
@MounaSafiHarab Good catch, I've updated those as well.

MounaSafiHarab:
tested it and it is all good.
I made a session Inactive in the backend and reproduced a new session that is identical but active.
Reinserting files now go to the active session one.

My concern though is on the Loris side, why is the NDB_Menu_Filter_imaging_browser.class.inc not querying on Active='Y'? as it stands, the inactive session gets displayed but when we click on it, it says 'Failed to retrieve data for timepoint' (I think we should add: AND s.Active = 'Y' on line 105).

PR #168, opened 2016-10-25
Removal of the deprecated array definition warning in mincpik: Redmine 11164

MounaSafiHarab:
this is on behalf of @gluneau

### DIFF
--- a/uploadNeuroDB/bin/mincpik
+++ b/uploadNeuroDB/bin/mincpik
@@ -432,7 +432,7 @@ if(defined($opt{'anot_bar'})){
    $bw = int($bh/10);
    
    # get range if not defined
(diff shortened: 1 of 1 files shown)


MounaSafiHarab:
_Before_:

FileID from mass_pic.pl is: 159
defined(@array) is deprecated at /data/ccna/bin/mri/uploadNeuroDB/bin/mincpik line 435.
	(Maybe you should just omit the defined()?)
FileID from mass_pic.pl is: 160
defined(@array) is deprecated at /data/ccna/bin/mri/uploadNeuroDB/bin/mincpik line 435.
	(Maybe you should just omit the defined()?)


_After_:

FileID from mass_pic.pl is: 159
FileID from mass_pic.pl is: 160

PR #138, opened 2016-06-29
remove minc, add a minc based on SeriesUID

gluneau:
New script to delete minc file based on SeriesUID

`uploadNeuroDB/minc_deletion.pl -profile prod -seriesuid 1.3.12.2.1107.5.2.32.35045.201507232316035928639007.0.0.0 confirm`

```
./uploadNeuroDB/minc_deletion.pl -help                                                                                              
*******************************************************************************
Minc Deletion 
*******************************************************************************

Author  :   Gregory Luneau
Date    :   July 2016
Version :   1 revision  2


The program does the following:

Deletes minc files from the web interface by:
  - Moving the existing files to an archive directory.
    .mnc .jpg .header .raw_byte.gz
  - Deleting all related data from 4 database tables.
    parameter_file, files_qcstatus, feedback_mri_comments, files


Summary of options:

-- Basic options ---------------------------------------------------------------
   -profile       config file in ../dicom-archive/.loris_mri [default: ""]
   -seriesuid     Only deletes this SeriesUID [default: ""]
   -fileid        Only deletes this FileID [default: ""]

usage: ./uploadNeuroDB/minc_deletion.pl [options] select|confirm
       ./uploadNeuroDB/minc_deletion.pl -help to list options
```

Also, modification of tarchiveLoader so that a unique file can be inserted based on SeriesUID

`uploadNeuroDB/tarchiveLoader -globLocation -profile null_grads -seriesuid 1.3.12.2.1107.5.2.32.35045.201507232316035928639007.0.0.0 DCM_2015-07-23_ImagingUpload-10-38-WiBmLY.tar`

PR #171, opened 2016-11-17
set caveat if acquisitionProtocol is overridden, undef seriesuid

gluneau:
Could go in 17.0 or 17.1 when it gets updated.


uploadNeuroDB/minc_insertion.pl
```diff
@@ -387,7 +390,11 @@
 $file->setFileData('OutputType', 'native');
 $file->setFileData('FileType', 'mnc');
 $file->setFileData('TarchiveSource', $tarchiveInfo{'TarchiveID'});
-$file->setFileData('Caveat', 0);
+if (defined($seriesuid)) {
+    $file->setFileData('Caveat', 1);
```

the seriesUID is set when we are deleting a file (as in this pull request: https://github.com/aces/Loris-MRI/pull/138/files)
I was thinking we should set Caveat to 1 when we are insering a violated scan (referring to this pull request: https://github.com/aces/Loris-MRI/pull/141)

PR #173, opened 2016-11-22
Remove the .DS_Store file from the candidate info validation process

MounaSafiHarab:
This will make the CandidateInfoValidation() ignore .DS_Store files which are commonly included with scans zipped on a MAC.
Subsequently, dicomTar.pl will remove this file (line 105 of: dicom-archive/dicomTar.pl)

### DIFF
--- a/uploadNeuroDB/NeuroDB/ImagingUpload.pm
+++ b/uploadNeuroDB/NeuroDB/ImagingUpload.pm
@@ -187,11 +187,11 @@ sub IsCandidateInfoValid {
         #1) Check to see if the file is of type DICOM###########
         #2) Check to see if the header matches the patient-name#
         ########################################################
-        if ( ( $_ ne '.' ) && ( $_ ne '..' ) ) {
+        if ( ( $_ ne '.' ) && ( $_ ne '..' ) && (basename($_) ne '.DS_Store')) {
             if ( !$this->isDicom($_) ) {
                 $files_not_dicom++;
             }
-	    else {
+    	    else {
          #######################################################
          #Validate the Patient-Name, only if it's not a phantom#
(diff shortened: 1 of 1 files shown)


uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -187,11 +187,11 @@ sub IsCandidateInfoValid {
         #1) Check to see if the file is of type DICOM###########
         #2) Check to see if the header matches the patient-name#
         ########################################################
-        if ( ( $_ ne '.' ) && ( $_ ne '..' ) ) {
+        if ( ( $_ ne '.' ) && ( $_ ne '..' ) && (basename($_) ne '.DS_Store')) {
```
MounaSafiHarab: I do not know what the first '.' and '..' are checking for.
anyone knows?

gluneau: Yes, the one dot and two dot are file system convention refer to the current directory and the parent directory respectively.

that's why when you do cd ..  
it brings you to the parent directory.

gluneau: I would also like to see && (basename($_) !~ '^\._')

This would exclude all files starting with dot underscore.

MounaSafiHarab: Thanks @gluneau 

but:
1) we are looping over files (not directories), I think
and
2)  what is returned  from $File::Find::name is the fullpath (including /tmp/ImagingUpload..../ at the beginning of the filename) 
so I feel we are always going to satisfy the if statement for the conditions on '.' and '..'

PR #178, opened 2017-02-10
[ImagingUpload.pm] Look inside file type and not file path for the word dicom

gluneau:
Previous file_type content:
/tmp/ImagingUpload-12-48-Rv9qJY/LEGO_PHANTOM_HOS_20160830/DICOM/.test: ASCII text

New file_type content:
DICOM medical imaging data

### DIFF
--- a/uploadNeuroDB/NeuroDB/ImagingUpload.pm
+++ b/uploadNeuroDB/NeuroDB/ImagingUpload.pm
@@ -417,7 +417,8 @@ sub isDicom {
     my $this         = shift;
     my ($dicom_file) = @_;
     my $cmd    = "file $dicom_file";
-    my $file_type    = `$cmd`;
+    my $cmd_log = `$cmd`;
+    my ($file_path, $file_type) = split /:/, $cmd_log;
     if ( !( $file_type =~ /DICOM/ ) ) {
         print "\n $dicom_file is not of type DICOM \n";
         return 0;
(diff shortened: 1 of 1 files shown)


uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -417,7 +417,8 @@ sub isDicom {
     my $this         = shift;
     my ($dicom_file) = @_;
     my $cmd    = "file $dicom_file";
-    my $file_type    = `$cmd`;
+    my $cmd_log = `$cmd`;
+    my ($file_path, $file_type) = split /:/, $cmd_log;
```
MounaSafiHarab: match on *last* DICOM medical imaging data

driusan: `DICOM medical imaging data$`

MounaSafiHarab:
@gluneau 

works; scripts fail now at the isDicom check step as opposed to the PatientName extraction step (see messages I get below; although I feel that either way the script would fail with Okay messages). 

but can you please change the description above? a .DS_Store file would not go through this check in the first place (check line 190). I am guessing you tested this before merging the pull request that excludes the .DS_Store from the check.

With my tests, here is what I get withOUT your changes:
`E: DcmElement: CommandField (0000,0100) larger (828667202) than remaining bytes in file
E: dcmdump: I/O suspension or premature end of stream: reading file: /data/not_backed_up/ImagingUpload-12-38-aAfrn0/DICOM/test/.test
Spool message is: 
The patient name cannot be extracted `

and WITH your change
`/data/not_backed_up/ImagingUpload-12-36-SfJ97X/DICOM/test/.test is not of type DICOM 
Spool message is: 
ERROR: There are 1 file(s) which are not of type DICOM `

MounaSafiHarab:
try `file -b` as per @driusan

PR #180, opened 2017-02-22
Fix file count in mri_upload when deleting a minc file and then re-inserting: Redmine 12008

MounaSafiHarab:


MounaSafiHarab:
@gluneau 

not sure if minc_insertion based on seriesuid needs to be updated as well now that minc_deletion is subtracting 1.
Not even sure we still need the minc_insertion based on serisuid as an argument given that now tarchiveLoader inserts only unique files (with exit 8 UNcommented out)...

gluneau:
image

MounaSafiHarab:
@cmadjar commented that the file count should account for multi-echo files as well (fix the new count to be minus the number of total minc files with that seriesUID)

PR #185, opened 2017-04-21
Fixed minc_deletion for FileID use case

davidblader:
Redmine 11947


MounaSafiHarab:
@davidblader 

all good!

PR #187, opened 2017-06-22
Run dicomtar.pl without the clobber option anymore to allow more than one tar per studyUID: Redmine 12171

MounaSafiHarab:
No more -clobber so as not to overwrite the existing .tar info in the tarchive table when the user needs to upload more than one upload with the same studyUID.
Related to:
https://github.com/aces/dicom-archive-tools/pull/48


MounaSafiHarab:
Closing as this will be somehow integrated with the implementation of the tar per modality (Redmine 9985)

PR #192, opened 2017-08-17
[SQL] Fixing MySQL 5.7 group by during visit creation

Jkat:
@MounaSafiHarab do you think this is fine?

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -234,7 +234,7 @@ sub getSessionID {
 	# fixme ask Jon ... is this still useful?
     # determine the centerID and new visit number (which is now deprecated) if getPSC() failed.
 	if($centerID == 0) {
-            $query = "SELECT IFNULL(MAX(VisitNo), 0)+1 AS newVisitNo, CenterID FROM session WHERE CandID=".$dbh->quote($subjectIDref->{'CandID'})." GROUP BY CandID newVisitNo, CenterID";
+            $query = "SELECT IFNULL(MAX(VisitNo), 0)+1 AS newVisitNo, CenterID FROM session WHERE CandID=".$dbh->quote($subjectIDref->{'CandID'})." GROUP BY CandID, newVisitNo, CenterID";
             $sth = $dbh->prepare($query);
             $sth->execute();
             if($sth->rows > 0) {
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -234,7 +234,7 @@ sub getSessionID {
 	# fixme ask Jon ... is this still useful?
     # determine the centerID and new visit number (which is now deprecated) if getPSC() failed.
 	if($centerID == 0) {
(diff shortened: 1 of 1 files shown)


uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -234,7 +234,7 @@ sub getSessionID {
 	# fixme ask Jon ... is this still useful?
     # determine the centerID and new visit number (which is now deprecated) if getPSC() failed.
 	if($centerID == 0) {
-            $query = "SELECT IFNULL(MAX(VisitNo), 0)+1 AS newVisitNo, CenterID FROM session WHERE CandID=".$dbh->quote($subjectIDref->{'CandID'})." GROUP BY CandID";
+            $query = "SELECT IFNULL(MAX(VisitNo), 0)+1 AS newVisitNo, CenterID FROM session WHERE CandID=".$dbh->quote($subjectIDref->{'CandID'})." GROUP BY CandID newVisitNo, CenterID";
```
MounaSafiHarab: @Jkat 
I think you have a typo in this query!
but out of curiosity, when would you exercise this part of the loop? is it when you create the session from the back end? because this is not the default setting (projects like CCNA and IBIS do not use this).

Jkat: Thanks! I fixed it. I'm not sure actually, I was setting up MRI on my sandbox in combination with what PreventAD has currently. Maybe Cecile would understand better why I was hitting this condition. If I remember correctly yes she did mention something about a new visit being created...and I do believe our scripts create new sessions from the back end, not sure if that's normal or not.

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -234,7 +234,7 @@ sub getSessionID {
 	# fixme ask Jon ... is this still useful?
     # determine the centerID and new visit number (which is now deprecated) if getPSC() failed.
 	if($centerID == 0) {
-            $query = "SELECT IFNULL(MAX(VisitNo), 0)+1 AS newVisitNo, CenterID FROM session WHERE CandID=".$dbh->quote($subjectIDref->{'CandID'})." GROUP BY CandID";
+            $query = "SELECT IFNULL(MAX(VisitNo), 0)+1 AS newVisitNo, CenterID FROM session WHERE CandID=".$dbh->quote($subjectIDref->{'CandID'})." GROUP BY CandID, newVisitNo, CenterID";
```
MounaSafiHarab: I get this error: `can't group on newVisitNo`
I think you meant `GROUP BY CandID, CenterID`?

Jkat: Strange, not sure how my scripts ran fine despite having newVisitNo in the GROUP BY...

MounaSafiHarab: I ran the query directly in the backend, so maybe running it from the script 'masks' the error?
I am testing the query directly form the backend because my project configuration is such that the visit has to be created to be able to upload, so this was the fastest way to tets this.

MounaSafiHarab:
@jkat,
is the CenterID really required? if not, can we remove it from the query altogether?

MounaSafiHarab:
@Jkat 

what is the status? still needs work?

Jkat:
Can be merged for now. We can relook at the query in the future. Maybe when Cecile is back too.

PR #194, opened 2017-09-14
New installs now also have their Imaging Pipeline ConfigSettings populated

MounaSafiHarab:
… based on the user's answers during the installation process: Redmine 13199.

Bug pointed out by @christinerogers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Ensure the _project/config.xml_ file (in the main LORIS codebase) contains the f
 
 #### 7. Verify filesystem permissions 
 
-As a final step, ensure that permissions on /data/$projectname and /data/incoming and their subdirectories are set such that lorisadmin and the Apache linux user can read, write _and_ execute all contents.
+Ensure that permissions on /data/$projectname and /data/incoming and their subdirectories are set such that lorisadmin and the Apache linux user can read, write _and_ execute all contents.
 
 The following must be recursively owned by the lorisadmin user and by Apache group:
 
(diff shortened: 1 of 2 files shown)


MounaSafiHarab:
@christinerogers 
can you please take a look since you pointed this out (thanks!)

@Jkat 
do you think you will have a few minutes to run a quick install/review this to see if this makes sense? (this is mainly because the `prod` file now has less entries, as those are moved to the front-end/database, and so new installs did not benefit from the users' answers to update those values directly in the database.

Jkat:
Did a quick test, it's good

README.md
```diff
@@ -110,6 +110,10 @@ The following must be recursively owned by the lorisadmin user and by Apache gro
    /data/incoming/
    /data/$projectname/bin/mri/dicom-archive/.loris_mri/prod
    ```
+8. Verify in the front-end some 'Imaging Pipeline' Configuration values
+
+As a final verification step, ensure that Config table entries accessible via the Configuration module in the front-end (Admin menu) under the `Imaging Pipeline` section are properly populated for ConfigSetting `Loris-MRI Data Directory` , `Study Name`, `User to notify when executing the pipeline`, `Full path to get_dicom_info.pl script`, and `Path to Tarchives`.
```
christinerogers: This is great.  I'd further shorten the wording here, and list the specific settings with bullets. 
break these down into bullet points, to make the text more readable.

christinerogers:
Comment:  the last link in the readme isn't working -- i tried a few syntax tweaks, not sure why I can't get it to work.  It's not the worst so I suggest we leave it, awkward looking as it is, for now.

MounaSafiHarab:
@driusan, @cmadjar 

this was not merged yesterday although it was discussed with @samirdas because I needed to rebase it (after we merged two other pull request just before this one, conflicts arose). So I rebased it and it now needs to be merged. I think @samirdas got too busy. Can you and/or @cmadjar merge?

PR #196, opened 2017-09-27
Corrected invalid code used to compare float numbers

nicolasbrossard:


MounaSafiHarab:
link from @nicolasbrossard 
https://docstore.mik.ua/orelly/perl4/cook/ch02_04.htm
for more background

MounaSafiHarab:
@nicolasbrossard 

is it possible to send this from you own branch?

PR #197, opened 2017-09-28
Corrected code that tests if two float numbers are equal

nicolasbrossard:

Reference [https://docstore.mik.ua/orelly/perl4/cook/ch02_04.htm](url)

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -47,6 +47,9 @@ use FindBin;
 $VERSION = 0.2;
 @ISA = qw(Exporter);
 
+# Number of decimals considered when testing if two floats are equal
+$FLOAT_EQUALS_NB_DECIMALS = 4;
+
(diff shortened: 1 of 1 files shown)


MounaSafiHarab:
@nicolasbrossard 

I tagged this for the next release so 1) we can discuss it at the next imaging meeting with others, and 2) I really will not be able to test it before this release (which is imminent).

MounaSafiHarab:
discussed at the Loris Imaging meeting of october 13th; no need for a bugfix branch; minor and major might be enough for now

MounaSafiHarab:
> Could you rebase your branch to aces:minor? When I checkout your branch, I still have the config settings fetched from the prod file instead of the Config table. Not sure why it does not show up on the Files changed front end of the PR...
> Thanks!

or to aces/major (they are all the same at this point in Loris-MRI), but since we will be releasing major (like Loris), please rebase on aces/major.

nicolasbrossard:
I'll rebase on aces/major then. I'll let you know when it's done

PR #201, opened 2017-11-15
Do not store PatientName header in notification_spool table:Redmine 13506

MounaSafiHarab:
If the patientname header in the DICOM files does not match the PSCID_CandID_VisitLabel naming convention (as entered using Imaging Uploader GUI), the error message stored in the notification_spool table (and also in the Log Viewer in the front-end) will show both the actual header info and what it is supposed/expected to be.
That was supposed to be informative for the user so they know how to fix the problem before the pipeline exits, however, if the site did not anonymize the PatientName, then our database is storing this info which it should not.
Simply remove the PatientName in the DICOM header from the stored message in the database.

### DIFF
--- a/uploadNeuroDB/NeuroDB/ImagingUpload.pm
+++ b/uploadNeuroDB/NeuroDB/ImagingUpload.pm
@@ -418,7 +418,7 @@ sub PatientNameMatch {
     }
     my ($l,$pname,$t) = split /\[(.*?)\]/, $patient_name_string;
     if ($pname !~ /^$this->{'pname'}/) {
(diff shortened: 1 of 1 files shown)


MounaSafiHarab:
@sruthymathew123 @llevitis 

this fix is due to the recently encountered issues with scans on IBIS. Can you please test?

sruthymathew123:
@MounaSafiHarab . I was about to ask about this and you sent a pull request then. Thanks Mouna for this.
I tested with un-anonymized data and seems to be perfect. 
screenshot from 2017-11-16 10-55-09
screenshot from 2017-11-16 11-20-08

PR #205, opened 2017-11-29
MySQL 5.7 - Mapping empty date of birth and scan to undef 

cmadjar:
If a scan for some reason does not contain a date of birth (for example when scanning a phantom), the dicom archive scripts would try to insert it as "". However, in MySQL 5.7, this creates an error and the tarchive entry is not created.

In order to fix this, reassigned date of birth and date of scan to undef if its value = ""

### DIFF
--- a/dicom-archive/DICOM/DCMSUM.pm
+++ b/dicom-archive/DICOM/DCMSUM.pm
@@ -187,6 +187,11 @@ QUERY
         }
     }
 
+    # If DoB is not set, $self->{header}->{birthdate} = '' which will
+    # not be allowed anymore in MySQL 5.7 for date fields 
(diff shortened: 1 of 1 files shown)


cmadjar:
Nicolas, I assigned it to you for review if you have time. I also put Mouna in the assignee so that she is aware of the PR. Would be good to merge before 19.0 release as this fixes a bug for MySQL 5.7 users.

Thank you!!

nicolasbrossard:
@cmadjar , I pulled your branch on my sandbox (running MySQL 5.7) and uploaded a scan with empty DoB and AcquisitionDate. Works with your fix. Fails without.

We have a winner!   :)

PR #206, opened 2017-11-30
Corrected code that tests if two float numbers are equal

nicolasbrossard:

Reference [https://docstore.mik.ua/orelly/perl4/cook/ch02_04.htm](url)

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -47,6 +47,9 @@ use FindBin;
 $VERSION = 0.2;
 @ISA = qw(Exporter);
 
+# Number of decimals considered when testing if two floats are equal
+$FLOAT_EQUALS_NB_DECIMALS = 4;
+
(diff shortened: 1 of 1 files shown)


uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -633,19 +636,36 @@ sub in_range
...
     foreach $range (@ranges) {
-	chomp($range);
-	if($range=~/^[0-9.]+$/) { ## single value element
-	    return 1 if $value==$range;
-	} else { ## range X-Y
-	    $range =~ /([0-9.]+)-([0-9.]+)/;
-	    return 1 if ($1 <= $value && $value <= $2);
-	}
+       chomp($range);
+       if($range=~/^[0-9.]+$/) { ## single value element
```
mohadesz: @nicolasbrossard is this the only instance where floating numbers are compared?

MounaSafiHarab:
tested on one scan and it inserts without any issues.

PR #207, opened 2017-11-30
[Cleanup]: Remove duplicate DBI.pm; Redmine 13613

MounaSafiHarab:
This removes the duplicate DBI.pm class we had (now that both repos are merged into one).
It will also facilitate the work of the working group of @mohadesz that is looking into building the DBI class.

Caveat for existing projects in case projects use customized scripts that point to DB/DBI.pm they will need to point to NeuroDB/DBI.pm instead.


MounaSafiHarab:
Tagged it as 'needs work' because I did not get to test it yet

cmadjar:
Hi Mouna, I could test it if you want and give you feedbacks if there are any problems. I'll be happy to :).

MounaSafiHarab:
> Hi Mouna, I could test it if you want and give you feedbacks if there are any problems. I'll be happy to :).

YESSS! especially or the DTIPrep part. But I just wanted to test it once myself with one scan then I will remove the needs work tag, then yes please, you could test it then.

PR #210, opened 2017-12-08
Add a TarchiveID column to mri_protocol_violated_scans

cmadjar:
Make sure to source the patch in [PR #3343](https://github.com/aces/Loris/pull/3343) to create and populate the column TarchiveID in the mri_protocol_violated_scans table.

Updates to the insertion scripts so that the new TarchiveID field of mri_protocol_violated_scans table can be populated when inserting a violated scan.


cmadjar:
Sent PR to the wrong branch...

PR #211, opened 2017-12-08
Add TarchiveID to mri_protocol_violated_scans

cmadjar:
Make sure to source the patch in [PR #3354](https://github.com/aces/Loris/pull/3354) to create and populate the column TarchiveID in the mri_protocol_violated_scans table.

Updates to the insertion scripts so that the new TarchiveID field of mri_protocol_violated_scans table can be populated when inserting a violated scan.

NOTE: this is for the next minor release of LORIS

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -398,12 +398,13 @@ Returns: Textual name of scan type
 
 sub identify_scan_db {
 
-    my  ($psc, $subjectref, $tarchiveInfo, $fileref, $dbhr,$minc_location) = @_;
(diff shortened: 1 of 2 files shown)


cmadjar:
Thank you @nicolasbrossard. I made the changes and renamed $tarchiveInfo to $tarchiveInfoRef

MounaSafiHarab:
@cmadjar 
please rebase on 19.0-dev so I can test it (might not make a difference anyway as both branches are very similar), but just to make sure.
Since the Loris one is merged, and Nicolas reviewed it, once I test it, I will merge it if all is good.

MounaSafiHarab:
tested it on a regular scan with no violated scans and one with a violated scan and the TarchiveID is properly populated in the latter case.
works perfectly as intended!

since the Loris part is merged, I will merge this one as well so they both go to 19.0.

PR #215, opened 2017-12-14
Refactor smart matching code in DTI.pm

nicolasbrossard:
Since the ~~ operator is now deprecated in Perl, I refactored DTI.pm so it does its job without it. Basically, instead of trying to find a 'smart match' in the DTI_frames list, I use the == operator to find a strict _numerical_ match.


cmadjar:
Tested when dimension in the minc file is the same as the expected dimension in the configuration section of the imaging pipeline => ran DTIPrep

Tested when dimension in the minc file is different that the one expected in the configuration section of the imaging pipeline => did not run DTIPrep

Hence: PASS MANUAL TESTING :)

PR #231, opened 2017-12-20
[Documentation] DICOM::Fields library perldocification

cmadjar:
Using perldoc/perlpod to document `Fields.pm` so that users can type in the terminal
`perldoc Fields.pm` and get the documentation.

Using the `pod2markdown` tool from perl `Pod::Markdown` module, the `Fields.md` file has been created so that we have a web displayed version of the documentation.
`pod2markdown Fields.pm > Fields.md`

### DIFF
--- a/dicom-archive/DICOM/Fields.pm
+++ b/dicom-archive/DICOM/Fields.pm
@@ -27,8 +27,8 @@ DICOM::Fields -- Definitions of fields of DICOM headers.
 
 =head1 DESCRIPTION
 
-Definitions of fields of DICOM headers. Not for medical use. This file is
-provided purely for experimental use.
+Definitions of fields of DICOM headers. This file is provided purely for
+experimental use.
 
 Simply creating an array of:
 
--- a/dicom-archive/DICOM/Fields.pm
+++ b/dicom-archive/DICOM/Fields.pm
@@ -4,6 +4,25 @@
 
 package DICOM::Fields;
 
+=pod
+
+=head1 NAME
+
+DICOM::Fields
+
+=head1 SYNOPSIS
+
+A few lines of code on how to use the package or script
+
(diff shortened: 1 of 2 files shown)


dicom-archive/DICOM/Fields.pm
```diff
@@ -1,9 +1,62 @@
...
+      my ($group, $elem, $code, $numa, $name) = split(/\s+/, $line);
+      my @lst = ($code, $name);
+      $dict{$group}{$elem} = [@lst];
+    }
+  }
+
+
+=head1 DESCRIPTION
+
+Definitions of fields of DICOM headers. Not for medical use. This file is
```
nicolasbrossard: Not for medical use? What do you mean by that?

dicom-archive/DICOM/Fields.pm
```diff
@@ -1,9 +1,62 @@
...
+=head1 SYNOPSIS
+
+  use DICOM::Element;
+  use DICOM::Fields;	# Standard header definitions.
+  use DICOM::Private;	# Private or custom definitions.
+
+  # Initialize dictionary.
+  # Read the contents of the DICOM dictionary into a hash by group and element.
+  # dicom_private is read after dicom_fields, so overrides common fields.
+  BEGIN {
```
nicolasbrossard: Not a big fan of this BEGIN block. I think this should be refactored (see review of #232 ) at a later time.

cmadjar: agreed

dicom-archive/DICOM/Fields.pm
cmadjar: I did not write this. It was the available description before formatting so I did not edit it. Not sure it is really relevant so I'll remove it.

PR #232, opened 2017-12-20
[Documentation] DICOM::Private library perldocification

cmadjar:
Using perldoc/perlpod to document `Private.pm` so that users can type in the terminal
`perldoc Private.pm` and get the documentation.

Using the `pod2markdown` tool from perl `Pod::Markdown` module, the `Private.md` file has been created so that we have a web displayed version of the documentation.
`pod2markdown Private.pm > Private.md`

### DIFF
--- a/docs/scripts_md/Private.md
+++ b/docs/scripts_md/Private.md
@@ -1,0 +1,47 @@
+# NAME
+
+DICOM::Private -- Definitions of (optional) private fields of DICOM headers.
+
+# SYNOPSIS
+
+    use DICOM::Element;
+    use DICOM::Fields;    # Standard header definitions.
+    use DICOM::Private;   # Private or custom definitions.
+
+    # Initialize dictionary.
+    # Read the contents of the DICOM dictionary into a hash by group and element.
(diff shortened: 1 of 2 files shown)


docs/scripts_md/Private.md
```diff
@@ -0,0 +1,47 @@
...
+# SYNOPSIS
+
+    use DICOM::Element;
+    use DICOM::Fields;    # Standard header definitions.
+    use DICOM::Private;   # Private or custom definitions.
+
+    # Initialize dictionary.
+    # Read the contents of the DICOM dictionary into a hash by group and element.
+    # dicom_private is read after dicom_fields, so overrides common fields.
+    BEGIN {
```
nicolasbrossard: If I understand correctly, in order to use the private fields one has to type this `BEGIN { .... }` block at the start of the script? If so, I think there are better strategies for implementing a DICOM dictionary. Let's leave it like this for now and come back at a later time.

cmadjar: yeah. I found that a little weird. We can definitely improve this!

PR #237, opened 2017-12-21
[Documentation] minc_insertion.pl

cmadjar:
Using perldoc/perlpod to document `minc_insertion.pl` so that users can type in the terminal
`perldoc minc_insertion.pl` and get the documentation.

Using the `pod2markdown` tool from perl `Pod::Markdown` module, the `minc_insertion.pl` file has been created so that we have a web displayed version of the documentation.
`pod2markdown minc_insertion.pl > minc_insertion.md`

### DIFF
--- a/uploadNeuroDB/minc_insertion.pl
+++ b/uploadNeuroDB/minc_insertion.pl
@@ -170,6 +170,8 @@ The program does the following:
   parameters
 - Finally sets the series notification
 
+Documentation: perldoc minc_insertion.pl
+
 HELP
 my $Usage = <<USAGE;
 usage: $0 </path/to/DICOM-tarchive> [options]
--- a/uploadNeuroDB/minc_insertion.pl
+++ b/uploadNeuroDB/minc_insertion.pl
@@ -13,10 +13,10 @@ perl minc_insertion.pl C<[options]>
 Available options are:
 
 -profile     : name of the config file in
-                C<../dicom-archive/.loris_mri>
+               C<../dicom-archive/.loris_mri>
 
 -reckless    : uploads data to database even if study protocol
-                is not defined or violated
+               is not defined or violated
 
 -force       : forces the script to run even if validation failed
 
@@ -27,8 +27,8 @@ Available options are:
 -tarchivePath: the absolute path to the tarchive file
 
 -globLocation: loosens the validity check of the tarchive allowing
-                 for the possibility that the tarchive was moved
-                 to a different directory
+               for the possibility that the tarchive was moved
+               to a different directory
 
 -newScanner  : if set [default], new scanner will be registered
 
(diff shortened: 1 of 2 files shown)


uploadNeuroDB/minc_insertion.pl
```diff
@@ -502,6 +574,15 @@
 ################################################################
 exit 0;
 
+
+=pod
+
+=head3 logHeader()
```
MounaSafiHarab: notice the same function in the tarchiveLoader pull request here:
https://github.com/aces/Loris-MRI/pull/246/files

I added a consistency item to the redmine ticket. I think we should follow what you have HERE and not have inputs to the function, but we need to make it consistently

uploadNeuroDB/minc_insertion.pl
```diff
@@ -1,4 +1,74 @@
...
+
+minc_insertion.pl -- Insert MINC files into the LORIS database system
+
+=head1 SYNOPSIS
+
+perl minc_insertion.pl C<[options]>
+
+Available options are:
+
+-profile     : name of the config file in
```
MounaSafiHarab: minor indentation issues here and the 2 or 3 inputs below
(see screenshot)
minc_insertion_indentation

uploadNeuroDB/minc_insertion.pl
cmadjar: @MounaSafiHarab The problem is that in here, no variables are passed to the function, while in tarchiveLoader, there are 3 arguments directly passed to the function. I agree that we should make the logHeader() functions consistent across scripts but this should be a different ticket than documentation. 
Actually, I end up creating the same function for other scripts so maybe there is a way to have that function only once in one of the library instead of duplicating it across all the scripts. 
What do you think?

uploadNeuroDB/minc_insertion.pl
cmadjar: @MounaSafiHarab done! Thank you :)

uploadNeuroDB/minc_insertion.pl
MounaSafiHarab: yes, definitely a separate pull request, but why I bring it as part of the documentation stuff, is because in tarchiveLoader, when you run it now (after it has been `perlified`!), the logHeader function gives a warning (I noticed it the other day) and it has to do with passing the inputs to logHeader. I know you tested it but it is easy to miss this message on the screen as everything progresses normally afterwards... so this should be addressed before we release the documentation branch.
but by all means, the same function used in many scripts should being centralized, if possible, should be another ticket.

MounaSafiHarab:
all good, merging

uploadNeuroDB/minc_insertion.pl
MounaSafiHarab: @cmadjar 
I just verified, I do not think that the logHeader() function in tarchiveLoader is used any differently (i.e. I do not see any arguments passed directly).

PR #243, opened 2017-12-22
[Documentation] database_files_update.pl perldocification 

cmadjar:
First, moved this script under the tools directory as it is an (old) tool scripts.

Second, using perldoc/perlpod to document `database_files_update.pl` so that users can type in the terminal
`perldoc database_files_update.pl` and get the documentation.

Using the `pod2markdown` tool from perl `Pod::Markdown` module, the `database_files_update.pl` file has been created so that we have a web displayed version of the documentation.
`pod2markdown database_files_update.pl > database_files_update.md`

Dependencies added: perldoc, Pod::Markdown

PR #244, opened 2017-12-22
[Documentation] MakeArchiveLocationRelative.pl perldocification

cmadjar:
First, moved this script under the tools directory as it is an (old) tool scripts.

Second, using perldoc/perlpod to document `MakeArchiveLocationRelative.pl` so that users can type in the terminal
`perldoc MakeArchiveLocationRelative.pl` and get the documentation.

Using the `pod2markdown` tool from perl `Pod::Markdown` module, the `MakeArchiveLocationRelative.pl` file has been created so that we have a web displayed version of the documentation.
`pod2markdown MakeArchiveLocationRelative.pl > MakeArchiveLocationRelative.md`

Dependencies added: perldoc, Pod::Markdown

### DIFF
--- a/docs/scripts_md/MakeArchiveLocationRelative.md
+++ b/docs/scripts_md/MakeArchiveLocationRelative.md
@@ -1,0 +1,53 @@
+# NAME
+
+MakeArchiveLocationRelative.pl -- Removes the root directory from the
+`ArchiveLocation` field in the `tarchive` table to make the path to the
+tarchive relative.
+
+# SYNOPSIS
+
+perl MakeArchiveLocationRelative.pl `[options]`
+
+Available option is:
+
+\-profile: name of the config file in ../dicom-archive/.loris\_mri
+
+# DESCRIPTION
+
+This script will remove the root directory from the ArchiveLocation field
+in the tarchive table to make path to the tarchive relative. This should
+be used once, when updating the LORIS-MRI code.
+
+## Methods
+
+### getTarchiveList($dbh, $tarchiveLibraryDir)
+
+This function will grep all the TarchiveID and associated ArchiveLocation
+present in the tarchive table and will create a hash of this information
+including new ArchiveLocation to be inserted into the DB.
+
+INPUT: database handle, tarchives location
+
+RETURNS: hash with tarchive information and new archive location
+
+### updateArchiveLocation($dbh, %tarchive\_list)
+
+This function will update the tarchive table with the new ArchiveLocation.
+
+INPUT: database handle, hash with tarchive information.
+
+# TO DO
+
+Nothing planned.
+
+# BUGS
+
+None reported.
+
+# LICENSING
+
+License: GPLv3
+
+# AUTHORS
+
+LORIS community <loris.info@mcin.ca> and McGill Centre for Integrative Neuroscience
--- a/docs/scripts_md/MakeArchiveLocationRelative.md
+++ b/docs/scripts_md/MakeArchiveLocationRelative.md
@@ -14,27 +14,28 @@ Available option is:
 
 # DESCRIPTION
 
-This script will remove the root directory from the ArchiveLocation field
-in the tarchive table to make path to the tarchive relative. This should
-be used once, when updating the LORIS-MRI code.
+This script will remove the root directory from the `ArchiveLocation` field
+in the `tarchive` table to make the `.tar` path a relative one. This should
+be used once to remove the root directory if the `tarchive` table still has
+some `ArchiveLocation` paths stored from the root directory.
 
 ## Methods
 
 ### getTarchiveList($dbh, $tarchiveLibraryDir)
 
-This function will grep all the TarchiveID and associated ArchiveLocation
-present in the tarchive table and will create a hash of this information
-including new ArchiveLocation to be inserted into the DB.
+This function will grep all the `TarchiveID` and associated `ArchiveLocation`
+present in the `tarchive` table and will create a hash of this information
+including new `ArchiveLocation` to be inserted into the database.
 
-INPUT: database handle, tarchives location
+INPUTS database handle, location of the `tarchive` directory
 
 RETURNS: hash with tarchive information and new archive location
 
 ### updateArchiveLocation($dbh, %tarchive\_list)
 
-This function will update the tarchive table with the new ArchiveLocation.
+This function will update the `tarchive` table with the new `ArchiveLocation`.
 
-INPUT: database handle, hash with tarchive information.
+INPUTS: database handle, hash with `tarchive` information.
 
 # TO DO
 
--- a/tools/update_to_relative_path/MakeArchiveLocationRelative.pl
+++ b/tools/update_to_relative_path/MakeArchiveLocationRelative.pl
@@ -1,0 +1,218 @@
+#! /usr/bin/perl
+
+=pod
+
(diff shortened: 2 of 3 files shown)


tools/update_to_relative_path/MakeArchiveLocationRelative.pl
```diff
@@ -0,0 +1,218 @@
+#! /usr/bin/perl
```
MounaSafiHarab: @cmadjar 

I agree the script should me boved to tools/ but why create a sub-directory inside the tools/ directory that is super specific?

cmadjar: hehe, boved, do you have a cold? ;)

Because there is a second one that will go in the same directory. Two scripts were created to make the paths relative in the DB, one on the dicom-archive side and one on the LORIS-MRI side. The other is in PR #243

MounaSafiHarab: > @MounaSafiHarab hehe, boved, do you have a cold? ;)

can't believe you saw this; I fixed it in a second, yet you saw old typo!!!

re. path, I still think it might be an overkill to create a sub-directory for every script or couple of scripts; especially these were done to accommodate a release change. But am open to whatever the group decides!

cmadjar: Sounds good! How about we mention it during our next meeting and I make the change on whatever we decide. I am not married to any options. BTW, in the redmine ticket, I marked all the files that should probably go into the tools directory. Need more investigation though, but it is a lot of files!

MounaSafiHarab: sure, already added an item to the meeting agenda; and yes many can go to the tools/ directory and the LDZ one should be removed I think :)

driusan: Wouldn't it make more sense to just have a single directory for all one time release update scripts?

cmadjar: ooooh, very good point! I guess we'll have a lot of pre-19.0 tools cause I believe this is the case for most of them :P.

docs/scripts_md/MakeArchiveLocationRelative.md
```diff
@@ -0,0 +1,53 @@
...
+`ArchiveLocation` field in the `tarchive` table to make the path to the
+tarchive relative.
+
+# SYNOPSIS
+
+perl MakeArchiveLocationRelative.pl `[options]`
+
+Available option is:
+
+\-profile: name of the config file in ../dicom-archive/.loris\_mri
```
MounaSafiHarab: I think this path is no longer valid if the file is moved to the tools/ directory; don't you think?

docs/scripts_md/MakeArchiveLocationRelative.md
```diff
@@ -0,0 +1,53 @@
...
+
+# DESCRIPTION
+
+This script will remove the root directory from the ArchiveLocation field
```
MounaSafiHarab: use C<>

docs/scripts_md/MakeArchiveLocationRelative.md
```diff
@@ -0,0 +1,53 @@
...
+in the tarchive table to make path to the tarchive relative. This should
```
MounaSafiHarab: here too for the `tarchive`

docs/scripts_md/MakeArchiveLocationRelative.md
```diff
@@ -0,0 +1,53 @@
...
+be used once, when updating the LORIS-MRI code.
```
MounaSafiHarab: Isn't this ambiguous? It was needed for the 15.10 release or even before :)

docs/scripts_md/MakeArchiveLocationRelative.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Methods
+
+### getTarchiveList($dbh, $tarchiveLibraryDir)
+
+This function will grep all the TarchiveID and associated ArchiveLocation
+present in the tarchive table and will create a hash of this information
+including new ArchiveLocation to be inserted into the DB.
```
MounaSafiHarab: same comment as above, please use C<>

docs/scripts_md/MakeArchiveLocationRelative.md
```diff
@@ -0,0 +1,53 @@
...
+
+INPUT: database handle, tarchives location
```
MounaSafiHarab: INPUTS

tools/update_to_relative_path/MakeArchiveLocationRelative.pl
MounaSafiHarab: I still think we should not have this very specific sub-directory (as otherwise, every script will be in its own directory).
I would suggest maybe moving to the `tools' for now, then maybe in another PR, re-organize the scripts within the tools/ once we know more how to logically split them

docs/scripts_md/MakeArchiveLocationRelative.md
MounaSafiHarab: I would re-word the sentence to say: "to make the .tar path a relative one".

docs/scripts_md/MakeArchiveLocationRelative.md
cmadjar: Since I am moving the script in the tools directory based on your suggestion below, it becomes valid now so I leave it like this.

docs/scripts_md/MakeArchiveLocationRelative.md
cmadjar: done

docs/scripts_md/MakeArchiveLocationRelative.md
cmadjar: Nice! Done!

docs/scripts_md/MakeArchiveLocationRelative.md
cmadjar: Changed it to: 

"This should
be used once to remove the root directory if the C<tarchive> table still has 
some C<ArchiveLocation> paths stored from the root directory."

docs/scripts_md/MakeArchiveLocationRelative.md
cmadjar: done

docs/scripts_md/MakeArchiveLocationRelative.md
cmadjar: done

MounaSafiHarab:
@cmadjar 

what are the permissions on these files in your directory?
the same applies to PR#243
I had to change the permissions to make these files executable, otherwise I could not run the script...
is it the case for you?

cmadjar:
@MounaSafiHarab I have a tendency to run perl scripts like that "perl script.pl" so I did not catch this.

MounaSafiHarab:
> @MounaSafiHarab I have a tendency to run perl scripts like that "perl script.pl" so I did not catch this.
> 
> But yes, they don't look like files with execute permission on my sandbox either. I think it was always the case.

Okay; I will merge, but maybe eventually we can make those permissions uniform (nothing to do with your pull request here but outstanding since a while ago I think)

PR #253, opened 2018-01-10
[Documentation] deletemincsqlwrapper.pl

MounaSafiHarab:
Generated using:
tools/example_scripts/deletemincsqlwrapper.pl docs/scripts_md/deletemincsqlwrapper.md

User can type:
perldoc tools/example_scripts/deletemincsqlwrapper.pl
to get the documentation at the terminal


MounaSafiHarab:
@cmadjar 

do you think then we should remove the `Readme.md` file that is in the same directory?
it was a recommendation by @samirdas. It made sense at the time, but since we are documenting everything in the .md form in docs/scripts_md/, do you think we need to keep both?

tools/example_scripts/deletemincsqlwrapper.pl
```diff
@@ -1,4 +1,49 @@
 #!/usr/bin/perl
+
+=pod
+
+=head1 NAME
+
+deletemincsqlqrapper.pl -- This script is a wrapper for deleting multiple MINC
+files at a time and optionally re-inserting them. It will pause for confirmation
+before deleting. B< Projects should modify the query as needed to suit their
```
cmadjar: Remove space between "B<" and "Projects".

tools/example_scripts/deletemincsqlwrapper.pl
```diff
@@ -1,4 +1,49 @@
...
+-insertminc   : Re-insert the deleted MINC
+
+
+
+=head1 DESCRIPTION
+
+This is an B<example> script that does the following:
+ - Deletes multiple MINC files fitting a common criterion from the database.
+ - Provides the option to re-insert deleted scans with their series UID when
+   using the `-insertminc` flag.
```
cmadjar: I am not sure to understand why ` character is used around -insertminc. If they should appear as code, then, should use C<> tag. Also present multiple times in the Notes section.

cmadjar:
I agree with you @MounaSafiHarab Since we are doing the perldoc, no need to repeat in the readme file. Thanks!!

cmadjar:
I tried to manually test the script and it does something weird. It deletes properly the data in the mysql tables but I get stuck in a weird loop... See attached screenshot. I've never used this script until now but this does not sound right.

screen shot 2018-01-10 at 3 22 34 pm

MounaSafiHarab:
@cmadjar 

just ran the script; and it runs fine for me (did you change the example query to match something in your database? when prompted, I selected the option 'Enter').
See screenshot; it tells me that a file is found, it displays the seriesUID for it, then it also displays the new count in the mri_upload table (since when we delete one, the count needs to be updated).

Now if I run the same script again 9with the same query), it tells me no files found (since I just deleted the only entry that it did match the step before).

deletemincsqlwrapper

MounaSafiHarab:
otherwise, all your changes are done!

PR #255, opened 2018-01-12
[Documentation] - batch_uploads_tarchive

nicolasbrossard:
Added perldoc for script `batch_uploads_tarchive`. Also ran `pod2markdown batch_uploads_tarchive` to generate `batch_uploads_tarchive.md`, the web-based documentation for the script.

### DIFF
--- a/batch_uploads_tarchive
+++ b/batch_uploads_tarchive
@@ -1,4 +1,90 @@
 #!/usr/bin/perl -w
+
+=pod
+
+=head1 NAME
+
+batch_uploads_tarchive - upload a batch of tarchives using script tarchiveLoader
+
+=head1 SYNOPSIS
+
+./batch_uploads_tarchive
+
+=head1 DESCRIPTION
+
+This script uploads a list of tarchive files to the database by calling script 
+C<tarchiveLoader> on each file in succession. The list of files to process is read 
+from STDIN, one file name per line. Each file name is assumed to be a path relative 
+to C<tarchiveLibraryDir> (see below).
+
+The following settings of file F<$ENV{LORIS_CONFIG}/.loris-mri/prod> affect the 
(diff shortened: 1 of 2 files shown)


batch_uploads_tarchive
```diff
@@ -1,4 +1,90 @@
...
+
+=pod
+
+=head1 NAME
+
+batch_uploads_tarchive - upload a batch of tarchives using script tarchiveLoader
+
+=head1 SYNOPSIS
+
+./batch_uploads_tarchive
```
MounaSafiHarab: @cmadjar @nicolasbrossard 
shouldn't this have more details?:)

docs/scripts_md/batch_uploads_tarchive.md
```diff
@@ -0,0 +1,58 @@
...
+  $mail\_user cointaining the list of files processed by batch\_upload\_tarchive
+
+File prod should also contain the information needed to connect to the database in an
+array `@db` containing four elements:
+
+- The database name
+- The SQL user name used to connect ot the database
+- The password for the user identified above
+- The database hostname
+
```
MounaSafiHarab: since the lines on the `@db` is applicable to all the scripts, should we put it everywhere?

batch_uploads_tarchive
nicolasbrossard: The script actually does not take any arguments, switches or parameters...Once run, it takes its input from STDIN so you have to type the parameters *along the way*. The script supports -D and -v but if you look at the code, these switches are not used so I did not document them. So no I don't think this needs more details

docs/scripts_md/batch_uploads_tarchive.md
nicolasbrossard: I suppose we could. Up to you. 😄

PR #259, opened 2018-01-19
Remove unused scripts and files

driusan:
This deletes the files that it was decided on removing in the LORIS imaging meeting. They are old, unused, and unsupported.


cmadjar:
@driusan could you rebase your PR to the major branch please? Thanks!!

MounaSafiHarab:
@driusan 
added the Caveat tag because some projects might be using a variation of one of those files (such as registerFile.pl), so maybe worth mentioning it to them?
maybe Document for release is the proper tag in this case; not sure...

MounaSafiHarab:
> @driusan could you rebase your PR to the major branch please? Thanks!!

@cmadjar 
should this go to 20.0: is that why you are asking for a rebase? or it is because his branch still has the dicom archive submodule so we can not test properly?
I was thinking 19.0 is the better option since it is major and it would allow us to not have to maintain them IF they are unused.

cmadjar:
@MounaSafiHarab I just thought that at this stage of the game we were not supposed to send additional PR to 19.0 and that this should go to another release. I you are OK with having that in 19.0, no big deal.

cmadjar:
@MounaSafiHarab I just changed the branch of the PR to go to aces/minor. I'll let you merge it if that is OK with you.

PR #260, opened 2018-01-23
Better MINC toolkit sourcing command in README and updated VERSION file

MounaSafiHarab:
as I was testing with Shen the MRI install, I realized that the source command for the MINC toolkit points to /opt/minc/ whereas more recent releases install automatically in /opt/minc/1.9.15/ (or /opt/minc/VERSION/), so this pull request:
1) updates the path to be more generic
and
2) update the version to 19.0.0


README.md
```diff
@@ -36,7 +36,11 @@ Download the pre-compiled package for your operating system.  Install required d
    sudo dpkg -i minc-toolkit<version>.deb
    ```
 
-  Then source the MINC toolkit environment by running (for bash) `source /opt/minc/minc-toolkit-config.sh` or (tcsh) `source /opt/minc/minc-toolkit-config.csh`.
+  Then source the MINC toolkit environment by running (for bash) 
+  `source $mincToolsDirectory/minc-toolkit-config.sh` or (tcsh) 
+  `source $mincToolsDirectory/minc-toolkit-config.csh`,
+  
+  where `$mincToolsDirectory` is the path where the MINC toolkit is installed (e.g. `/opt/minc/` or `opt/minc/1.9.15/` for more recent installs)
```
cmadjar: - missing / in front of opt of the version path
- instead of a version number, shouldn't it say /opt/minc/VERSION/ instead of the version number?

MounaSafiHarab: agree with both.
I thought about the second point, but was hesitant as I am not sure the Minc Toolkit install directory will always be the version number (there are 3 different paths so far depending on whether it is before 1.0.8, between 1.0.8 and 1.0.9 or after 1.0.9:))
but I think your proposal is the most generic.

MounaSafiHarab:
> @MounaSafiHarab One little typo and comment. Thanks!!

done! thanks

MounaSafiHarab:
> Looks great to me, and I agree with Cecile's comments.

done @christinerogers.
Is this a good time to update the version number ( as I do it in this pull request so I wanted to make sure from you).

MounaSafiHarab:
> (However, Dave/Rida have removed the Version file in Loris core Loris#3292 in favour of auto-generating one so I'm not sure how we should negotiate that with the MRI repo.) -CR

good point. Yes they did. But for the MRI side, we need to investigate if there is an auto-generation capability (on the Loris side, I believe they do it during the npm step which we do not do here).  Maybe worth bringing up in an upcoming imaging meeting.

MounaSafiHarab:
can anyone of @cmadjar, @christinerogers squash and merge if you are both Okay with the changes?

cmadjar:
Done!

MounaSafiHarab:
@christinerogers @cmadjar 

check this: https://github.com/aces/Loris/pull/3445
looks like even on the Loris side they still update the version file!

christinerogers:
@MounaSafiHarab so they did!  
(As you said, something to be discussed for Loris-MRI)

PR #266, opened 2018-01-30
Grep the different file types from the DB instead of hardcoding it in the code

cmadjar:
In this PR, instead of hardcoding the different file types in the File.pm file, reads this information from the ImagingFileType table.

The only exception left in the if statements are if a file matches '.mnc' or '.mnc.gz', it will determine the file type to be 'mnc'.

Redmine ticket associated with this PR: https://redmine.cbrain.mcgill.ca/issues/13839

### DIFF
--- a/uploadNeuroDB/NeuroDB/File.pm
+++ b/uploadNeuroDB/NeuroDB/File.pm
@@ -261,7 +261,7 @@ QUERY
     $sth->execute();
 
     # else, loop through the different values from ImagingFileTypes table
-        # and see if $file matches one of the file types.
+    # and see if $file matches one of the file types.
     while (my $fileTypeRow = $sth->fetchrow_hashref()) {
         if ($file =~ /\.$fileTypeRow->{'type'}(\.gz)?$/) {
             $fileType = $fileTypeRow->{'type'};
--- a/uploadNeuroDB/NeuroDB/File.pm
+++ b/uploadNeuroDB/NeuroDB/File.pm
@@ -260,17 +260,11 @@ QUERY
     my $sth = ${$this->{'dbhr'}}->prepare($query);
     $sth->execute();
 
-    # determine file type of $file
-    if($file =~ /\.mnc(\.gz)?$/) {
-        # if file type is .mnc or .mnc.gz, then file type will be 'mnc'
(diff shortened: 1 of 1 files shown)


cmadjar:
@nicolasbrossard I am assigning this to you in case you have some time to review this.

cmadjar:
@nicolasbrossard Good point. I updated the code so that it is possible to .gz after the main extension determining the file type.

Note: in the `ImagingFileTypes` table, there is '.nii' and '.nii.gz'. However, we should probably have the imaging script behave the same way for '.mnc.gz' and '.nii.gz' (meaning it does not take into account the .gz extension and just return 'mnc' or 'nii' for the file type) @MounaSafiHarab what do you think?

If everyone is OK with this plan, I'll update a PR on the LORIS side for the `ImagingFileTypes` table so that we can map '.nii.gz' file type to '.nii' in the `files` table and remove '.nii.gz' from the `ImagingFileType` table (for projects already using LORIS). Looking at the MRI script, I am pretty confident that no '.nii' and '.nii.gz' are present in the `files` table anyway, but better safe than sorry.

cmadjar:
@nicolasbrossard I think it is ready for review now. Thanks!

cmadjar:
@nicolasbrossard Great suggestion that I did just apply!! Thank you!! You can look at it one more time when you get a chance.

nicolasbrossard:
👍

nicolasbrossard:
Tried to insert a tarball via the front end with autolaunch and it worked perfrectly. I also added a trace in `File.pm` to make sure that it returned the right file type and it did. I believe it's the only thing that I can test for now. Tagging as 'PassedManualTest'.

:+1:

MounaSafiHarab:
all good on my end as well (only can test for one of the types; that is `mnc` type), but if this type works, the rest should work as well!
Tagging as passed manual test here as @nicolasbrossard  beat me to that!

Added Update PerDoc tag eventhough the .pl is already done that way, just so we remember to update the .md file later on...

MounaSafiHarab:
merging now

cmadjar:
Perldoc updated in PR #299 sent to aces/minor so removing the flag "Update Perldoc" from this PR.

PR #272, opened 2018-02-14
Installer: Populate the default LORIS config settings for the MRI paths (Redmine 13915)

MounaSafiHarab:
On the MRI side, since the user provides the projectname during the interactive install; it is better to auto-populate/change the default values that LORIS sets.
It is already done for some setting (Imaging Pipeline section), but makes sense to do it also for the Path section in the configuration module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,12 +102,20 @@ The following must be recursively owned by the lorisadmin user and by Apache gro
    ```
 #### 7. Verify Configuration module settings for Imaging Pipeline
   
-In the LORIS front-end, under the Admin menu, go to the `Config` module.  Under the section `Imaging Pipeline`, verify/set the following config settings: 
+In the LORIS front-end, under the Admin menu, go to the `Config` module.  Verify/set the following config settings 
+
+under the section `Imaging Pipeline`:
  * `Loris-MRI Data Directory`
  * `Study Name`
  * `User to notify when executing the pipeline`
  * `Full path to get_dicom_info.pl script`
  * `Path to Tarchives`
+
+under the section `Path`:
+ * `Imaging Data`
+ * `LORIS-MRI Code`
+ * `MINC files`
+ * `Images`
 
 Click 'Submit' at the end of the Configuration page to save any changes. 
 
--- a/README.md
+++ b/README.md
@@ -102,20 +102,20 @@ The following must be recursively owned by the lorisadmin user and by Apache gro
    ```
 #### 7. Verify Configuration module settings for Imaging Pipeline
   
-In the LORIS front-end, under the Admin menu, go to the `Config` module.  Verify/set the following config settings 
+In the LORIS front-end, under the Admin menu, go to the `Config` module.  Verify/set the following config settings (examples below illustrated for a project name `demo`):
 
-under the section `Imaging Pipeline`:
- * `Loris-MRI Data Directory`
- * `Study Name`
+Under the section `Imaging Pipeline`:
+ * `Loris-MRI Data Directory` (typically `/data/demo/data/')
+ * `Study Name` (`exampleStudy`; this name will be appended as a prefix to the filenames in LORIS' Imaging Browser)
  * `User to notify when executing the pipeline`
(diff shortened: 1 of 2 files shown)


MounaSafiHarab:
@christinerogers @cmadjar 

I think this change makes sense for new installs; what do you think?

imaging_install.sh
```diff
@@ -187,10 +187,14 @@ fi
...
-mysql $mysqldb -h$mysqlhost --user=$mysqluser --password="$mysqlpass" -A -e "UPDATE Config SET Value='/data/$PROJ/data/tarchive' WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='tarchiveLibraryDir')"
+mysql $mysqldb -h$mysqlhost --user=$mysqluser --password="$mysqlpass" -A -e "UPDATE Config SET Value='/data/$PROJ/data/tarchive/' WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='tarchiveLibraryDir')"
+mysql $mysqldb -h$mysqlhost --user=$mysqluser --password="$mysqlpass" -A -e "UPDATE Config SET Value='/data/$PROJ/data/' WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='imagePath')"
+mysql $mysqldb -h$mysqlhost --user=$mysqluser --password="$mysqlpass" -A -e "UPDATE Config SET Value='/data/$PROJ/data/' WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='data')"
+mysql $mysqldb -h$mysqlhost --user=$mysqluser --password="$mysqlpass" -A -e "UPDATE Config SET Value='/data/$PROJ/data/' WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='mincPath')"
+mysql $mysqldb -h$mysqlhost --user=$mysqluser --password="$mysqlpass" -A -e "UPDATE Config SET Value='/data/$PROJ/bin/mri/' WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='MRICodePath')"
```
MounaSafiHarab: add the AND condition if the value is at its default

MounaSafiHarab: @cmadjar @christinerogers 

I added the condition as per @driusan 's feedback during the imaging meeting

MounaSafiHarab:
@cmadjar @christinerogers 
let me know if you got the chance to look at this since it was decided in the imaging meeting we send this to 19.0?

README.md
```diff
@@ -102,13 +102,21 @@ The following must be recursively owned by the lorisadmin user and by Apache gro
...
+
+under the section `Imaging Pipeline`:
  * `Loris-MRI Data Directory`
  * `Study Name`
  * `User to notify when executing the pipeline`
  * `Full path to get_dicom_info.pl script`
  * `Path to Tarchives`
 
+under the section `Path`:
+ * `Imaging Data`
```
cmadjar: I would tend to give example of what the config settings would look like for a typical LORIS project. What do you think? I have no strong opinion about this so either way would work.

README.md
```diff
@@ -102,13 +102,21 @@ The following must be recursively owned by the lorisadmin user and by Apache gro
    ```
 #### 7. Verify Configuration module settings for Imaging Pipeline
   
-In the LORIS front-end, under the Admin menu, go to the `Config` module.  Under the section `Imaging Pipeline`, verify/set the following config settings: 
+In the LORIS front-end, under the Admin menu, go to the `Config` module.  Verify/set the following config settings
```
cmadjar: I would put a dot after "settings" and have upper case U for "Under" of the next 2 subsections

README.md
MounaSafiHarab: check the latest commit; I added more explanation... is this better?

README.md
MounaSafiHarab: agreed. I just however put `:` instead of the `.`

nicolasbrossard:
Works for me. Approved.   :+1:

PR #275, opened 2018-02-16
Adding -uploadID option to minc_insertion.pl

cmadjar:
Add an `-upload_id` option that can be used to call the `minc_insertion.pl` script. Will keep the `-tarchivePath` option as well.

Running minc_insertion.pl now requires that either `-tarchivePath` or `-upload_id` options are set (but not both at the same time). 
=> If `-tarchivePath` is given, the corresponding `UploadID` is fetched at the beginning of the script. The `UploadID` is then passed to the different functions of `MRIProcessingUtility.pm` for logging purposes instead of querying the `UploadID` based on the `-tarchivePath` in each function.

With this change, no dependant anymore on the tarchive location for logging purposes.

Linked Redmine ticket: https://redmine.cbrain.mcgill.ca/issues/13936

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
+++ b/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
@@ -149,13 +149,11 @@ extracts it so data can actually be uploaded
 =cut
 sub extract_tarchive {
     my $this = shift;
-    my ($tarchive, $tarchive_srcloc, $seriesuid) = @_;
-    my $upload_id = undef;
+    my ($tarchive, $upload_id, $seriesuid) = @_;
     my $message = '';
     my $tarnames = '';
     # get the upload_id from the tarchive source location
     # to pass to the notification_spool
-    $upload_id = getUploadIDUsingTarchiveSrcLoc($tarchive_srcloc);
     $message = "\nExtracting tarchive $tarchive in $this->{TmpDir} \n";
     $this->spool($message, 'N', $upload_id, $notify_detailed);
     my $cmd = "cd $this->{TmpDir} ; tar -xf $tarchive";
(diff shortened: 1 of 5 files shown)


uploadNeuroDB/minc_insertion.pl
```diff
@@ -204,6 +216,11 @@
 print "\nlog dir is $LogDir and log file is $logfile \n" if $verbose;
 open LOG, ">>", $logfile or die "Error Opening $logfile";
 LOG->autoflush(1);
+# strings needed for the logHeader, if not set, as an argument, use empty string
+my $source_data_for_log = $tarchive;
```
nicolasbrossard: You are essentially doing 

`my $source_data_for_log = (defined $tarchive ? $tarchive :  "")`

which can be more concisely written as

`my $source_data_for_log = $tarchive // "";`

Same thing for the definition of variable `$upload_id_for_log`

cmadjar: cool! Thanks!!

cmadjar:
@nicolasbrossard  Ready for another round of review/testing.

MounaSafiHarab:
@cmadjar 
can you please change the base branch to minor?

cmadjar:
> @cmadjar
can you please change the base branch to minor? just so that we make sure we do not merge accidentally into 19.0-dev

@MounaSafiHarab just changed the base branch. Thanks for catching it!!

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -220,9 +220,7 @@ sub extractAndParseTarchive {
 sub determineSubjectID {
 
     my $this = shift;
-    my ($scannerID,$tarchiveInfo,$to_log) = @_;
-    my $tarchive_srcloc = $tarchiveInfo->{'SourceLocation'};
-    my $upload_id = getUploadIDUsingTarchiveSrcLoc($tarchive_srcloc);
+    my ($scannerID,$tarchiveInfo,$to_log, $upload_id) = @_;
```
MounaSafiHarab: what will happen to the scripts that currently call this function while passing the $tarchive_srcloc? 
i.e. tarchiveLoader and tarchive_validation.pl scripts.

cmadjar: oups... Very valid point that I oversaw. Fixing it for all the functions that were changed in MRIProcessingUtility.pm.

uploadNeuroDB/minc_insertion.pl
```diff
@@ -160,8 +163,14 @@
     exit 3; 
 }
 
+if ( ($tarchive && $upload_id) || (!$tarchive && !$upload_id) ) {
```
nicolasbrossard: Use the `xor` operator here. It tests that one of the two values is set but not both. So the above test could be written as:

`if (!($tarchive xor $upload_id))`

cmadjar:
> I am guessing we can delete subroutine getUploadIDUsingTarchiveSrcLoc now, right? The MRI upload object broker will be used anyway instead of this when major is merged.

@nicolasbrossard Actually, no, it is still being used by several scripts to determine the uploadID that corresponds to the archivelocation given as an argument (for example line 183 of tarchive_validation.pl). When the object broker are in place and all the scripts mapped, then yes, we can remove it.

uploadNeuroDB/minc_insertion.pl
cmadjar: ouhh, nice!

PR #283, opened 2018-03-05
MRI protocol table can no longer have comma separated ranges: Redmine 14021

MounaSafiHarab:
this is to be merged at the same time (same release) as:
https://github.com/aces/Loris/pull/3523


MounaSafiHarab:
@cmadjar @nicolasbrossard 

I added a tag 'Update perlDoc' because we are updating those fucntions that had already done 'perldocification'! this tag is to make us keep track of those for later when the loris-mri-docs branch is merged onto minor.

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -664,19 +664,20 @@ sub in_range
...
-       if($range=~/^[0-9.]+$/) { ## single value element
-           return 1 if &floats_are_equal($value, $range, $FLOAT_EQUALS_NB_DECIMALS);
-       } else { ## range X-Y
-           $range =~ /([0-9.]+)-([0-9.]+)/;
-           return 1 if ($1 <= $value && $value <= $2) 
-               || &floats_are_equal($value, $1, $FLOAT_EQUALS_NB_DECIMALS) 
-               || &floats_are_equal($value, $2, $FLOAT_EQUALS_NB_DECIMALS);
-       }
+    if($range_string=~ /,/) {
+        print "Comma separated ranges, as found in $range_string, are no longer supported\n";
```
cmadjar: Should it be a print, a print LOG? Should the script exit? Should we also use the notification?

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -664,19 +664,20 @@ sub in_range
...
+        return 0;
```
cmadjar: Should it be `return 0` or `return undef`? How do the scripts handle this return 0 ?

MounaSafiHarab:
> @MounaSafiHarab Need to clarify how the scripts will handle a bad protocol entry. Other than that

perfectly valid point. I had done 2 commits, specifically so we can discuss this :)
Same applies to my other pull request... we somehow should have a smooth transition to the user and am not sure the best way to do this.

MounaSafiHarab:
Discussed at the Imaging meeting of March 16, 2018;
we can exit so projects do not re-instate comma values in the protocol table (eventually the columns in the table would be decimal so this will no longer be possible, but the column type change on the Loris side might be a separate PR in a separate release).

uploadNeuroDB/NeuroDB/MRI.pm
nicolasbrossard: This is the perfect place to throw an exception, but since we do not have access to them (yet...), I would go with "die". Maybe also put a "TODO" that says that this should be refactored to throw an exception when the Moose PR is merged.

nicolasbrossard: Usually I would not be in favor of a function call "dying" on us but in this specific case I think it makes sense.

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -763,19 +762,24 @@ sub in_range
     return 0 unless $range_string;
     return 0 unless defined($value);
 
-    my @ranges = split(/,/, $range_string);
+    ## mri_protocol table is being normalized in steps. 
+    ## Until this is completed in full and/or the database 
+    ## prevents such a setting, do not allow projects
+    ## to put comma-separated values in the table columns.
+    ## If they do, LOG the specific scan type as a violated scan
```
MounaSafiHarab: @cmadjar @nicolasbrossard 

the comment above summarizes how the comma separated values are being handled. I can basically return 0 or 1 in this function, so I return a 0 as if it is a violated scan (so projects can act on this).
The pull request in Loris #3528 will prevent such a setting as the column type will no longer allow strings

uploadNeuroDB/NeuroDB/ExitCodes.pm
```diff
@@ -189,4 +189,4 @@ our $NO_VALID_MINC_CREATED = 190; # if no valid MINC file was created
 
 our $REGISTER_PROGRAM_FAILURE = 200; # if MNI::Spawn::RegisterPrograms failed
 
-## -- FROM
\ No newline at end of file
+## -- FROM
```
cmadjar: should probably remove that `## -- FROM` comment from the last line of the ExitCodes.pm library

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -763,19 +762,24 @@ sub in_range
...
     return 0 unless defined($value);
 
-    my @ranges = split(/,/, $range_string);
+    ## mri_protocol table is being normalized in steps. 
+    ## Until this is completed in full and/or the database 
+    ## prevents such a setting, do not allow projects
+    ## to put comma-separated values in the table columns.
+    ## If they do, LOG the specific scan type as a violated scan  
+    if($range_string=~ /,/) {
+        print STDERR "Comma separated ranges, as found in '$range_string', are no longer supported. Please remove. Logging as a violated scan. \n";
```
cmadjar: I would modify the message to something like that:
"Comma separated ranges, as found in '$range_string', are no longer supported. Please modify your mri_protocol table accordingly. Logging the scan as a violated scan for now.\n"

Also, code should probably not exceed 80 characters per line so that we can view it on Github without having to scroll to the right. Very nitpicky, I know ;).

docs/scripts_md/ExitCodes.md
```diff
@@ -54,11 +54,11 @@ Below is a list of the possible exit codes:
 9\. Exit codes from uploadNeuroDB/NeuroDB/ImagingUpload.pm (exit codes from
 170 to 179)
 
-11\. Exit codes from uploadNeuroDB/minc\_insertion.pl (exit codes from 180 to 189)
+10\. Exit codes from uploadNeuroDB/minc\_insertion.pl (exit codes from 180 to 189)
```
cmadjar: I am not sure to understand why these appears in your PR. Maybe it needs rebase?

uploadNeuroDB/NeuroDB/MRI.pm
MounaSafiHarab: I took your feedback into account, so I used your message but in a different place.

docs/scripts_md/ExitCodes.md
MounaSafiHarab: NO CLUE. I kept wondering and wondering. I merged aces/minor and it shows no changes... anyway I did hard reset it so it no longer shows in the diff.

MounaSafiHarab: Ohh I think I know. The last one that got merged onto minor did not have the updated `.md` regenerated, so now that I removed the `## -- FROM` as per your earlier comment and regenerated the `/.md` it shows up again; so it is a legitimate change

MounaSafiHarab:
@cmadjar 
after discussion with @nicolasbrossard 
what I was doing was probably not the best place to throw this error (it was being done initially where for each MINC file we check against the mri_protocol table, so depending on where the comma was present, projects might or might not see the error, and the scan being logged as violated might not always correspond to the row that has the comma in it...); for that, we thought it is best I check for the protocol table BEFORE entering the loop on each MINC file; and log the error in the notification_spool table as well; then `die` as per Nick's suggestion!

This also makes it easier when the database no longer supports the varchar on those columns to simply remove this added function altogether.

cmadjar:
@MounaSafiHarab Sorry, another comment. Shouldn't the code of the `in_range` function be modified to remove the split per comma and the loop? It looks like it is not part anymore of your PR but I think it is still relevant to remove that part of the code since we say we don't support it anymore.

uploadNeuroDB/tarchiveLoader
```diff
@@ -484,6 +484,25 @@ if ($mcount < 1) {
     exit $NeuroDB::ExitCodes::NO_VALID_MINC_CREATED;
 }
 
+################################################################
+################################################################
+# Make sure the mri_protocol table no longer has comma-separated
+########################## values ##############################
+################################################################
+my ($isMRIProtocolValid) = 
+    $utility->isValidMRIProtocol();
```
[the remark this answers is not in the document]
MounaSafiHarab: okay! done

uploadNeuroDB/tarchiveLoader
```diff
@@ -484,6 +484,25 @@ if ($mcount < 1) {
...
+
+if ( !($isMRIProtocolValid) ) {
+    my $message = "Comma separated ranges or values in the mri_protocol "
+                  . "table are no longer supported. Please modify your "
+                  . "mri_protocol table accordingly. Exiting now. \n";
+    $notifier->spool('tarchive loader', $message, 0,
+		    'tarchiveLoader', $upload_id, 'Y',
+		    $notify_notsummary);
+    print STDERR $message;
+    die;
```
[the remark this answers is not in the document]
MounaSafiHarab: as you see from  my commits, I went for a new exit code to die as per @nicolasbrossard 's comment/suggestion. I do not have any strong feeling about either/or. I changed it back to be an exit code, so I hope @nicolasbrossard does not mind.

MounaSafiHarab:
> @MounaSafiHarab Sorry, another comment. Shouldn't the code of the in_range function be modified to remove the split per comma and the loop? It looks like it is not part anymore of your PR but I think it is still relevant to remove that part of the code since we say we don't support it anymore.

perfect point, thanks for catching. It is my bad, I rebased by copying the file from minor and forgot to remove the extra lines, and when tetsing, we never exercise this anymore as we exit much earlier if there are commas...

uploadNeuroDB/NeuroDB/ExitCodes.pm
```diff
@@ -140,7 +140,7 @@ our $TARGET_EXISTS_NO_CLOBBER  = 83; # if tarchive already exists but option
 our $UNKNOWN_PROTOCOL          = 84; # if could not find acquisition protocol
                                      # of the file to be inserted
 our $NOT_A_SINGLE_STUDY        = 85; # if the upload regroups multiple studies
-
+our $BAD_PROTOCOL_SETTING      = 86; # unacceptable mri protocol table entry
```
cmadjar: instead of creating a new exit code, I would tend to use PROJECT_CUSTOMIZATION_FAILURE (setting a proper mri_protocol table is part of the customization). Also I guess that eventually the new function you added to check the mri_protocol table will be removed once the second part of the mri_protocol revamp will be merged, right?

MounaSafiHarab: Okay. I changed the exit code to the existing one.
And yes, the intention is to remove this validation function once the other PR is merged (perhaps a release later) which will then automatically prevent data entry to be non-conform. I added this to the TODO as well. Updated the perldoc as well.

Let me know if any more feedback!

uploadNeuroDB/tarchiveLoader
```diff
@@ -484,6 +484,24 @@ if ($mcount < 1) {
...
+
+if ( !($isMRIProtocolValid) ) {
+    my $message = "\nComma separated ranges or values in the mri_protocol "
+                  . "table are no longer supported. Please modify your "
+                  . "mri_protocol table accordingly. Exiting now. \n";
+    $notifier->spool('tarchive loader', $message, 0,
+		    'tarchiveLoader', $upload_id, 'Y',
+		    $notify_notsummary);
+    print STDERR $message;
+    exit $NeuroDB::PROJECT_CUSTOMIZATION_FAILURE;
```
cmadjar: oups. I missed that earlier and noticed it when testing the code. It should be `$NeuroDB::ExitCodes::PROJECT_CUSTOMIZATION_FAILURE;` and not `$NeuroDB::PROJECT_CUSTOMIZATION_FAILURE;`

MounaSafiHarab: yep, my bad again!

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -763,19 +763,14 @@ sub in_range
...
-           return 1 if &floats_are_equal($value, $range, $FLOAT_EQUALS_NB_DECIMALS);
-       } else { ## range X-Y
-           $range =~ /([0-9.]+)-([0-9.]+)/;
-           return 1 if ($1 <= $value && $value <= $2) 
-               || &floats_are_equal($value, $1, $FLOAT_EQUALS_NB_DECIMALS) 
-               || &floats_are_equal($value, $2, $FLOAT_EQUALS_NB_DECIMALS);
-       }
+    chomp($range_string);
+    if($range_string=~/^[0-9.]+$/) { ## single value element
+        return 1 if &floats_are_equal($value, $range, $FLOAT_EQUALS_NB_DECIMALS);
```
cmadjar: in the function `&floats_are_equal`, `$range` should be replaced by `$range_string`

cmadjar:
@MounaSafiHarab I removed the blocked label since the associated PR on the LORIS side and this PR are approved and ready to be merged into minor when the 19.2 branch will be created.

MounaSafiHarab:
> Tested with a comma in the MRI protocol table and without the comma. All good! :)

FINALLY I got it right.

MounaSafiHarab:
@cmadjar @nicolasbrossard 
is this ready and can now be merged on minor here?
if so please do so the LORIS one can be merged as well.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1781,6 +1781,41 @@ sub spool  {
...
+                OR TE_range LIKE '%,%'
+                OR TI_range LIKE '%,%'
+                OR slice_thickness_range LIKE '%,%'
+                OR xspace_range LIKE '%,%'
+                OR yspace_range LIKE '%,%'
+                OR zspace_range LIKE '%,%'
+                OR xstep_range LIKE '%,%'
+                OR ystep_range LIKE '%,%'
+                OR zstep_range LIKE '%,%'
+                OR time_range LIKE '%,%'";
```
MounaSafiHarab: I should add `mri_protocol_checks` and `mri_violations_log` to this check as well.
I can do it now for the first table, but for the second, I need to wait on @llevitis to see her PR on the Loris side first.

MounaSafiHarab: I did `mri_protocol_checks` in PR #318

PR #284, opened 2018-03-06
Change hyphen separated ranges in mri_protocol to MIN/MAX columns: Redmine 14022

MounaSafiHarab:
this is to be merged at the same time (same release) as:
aces/Loris#3528


MounaSafiHarab:
Discussion required to address the 'return 0' case and how we should properly deprecate the other function.

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -511,19 +511,19 @@ sub identify_scan_db {
...
-
-                && (!$rowref->{'xspace_range'} || &in_range($xspace, $rowref->{'xspace_range'}))
-                && (!$rowref->{'yspace_range'} || &in_range($yspace, $rowref->{'yspace_range'}))
-                && (!$rowref->{'zspace_range'} || &in_range($zspace, $rowref->{'zspace_range'}))
-
-                && (!$rowref->{'xstep_range'} || &in_range($xstep, $rowref->{'xstep_range'}))
-                && (!$rowref->{'ystep_range'} || &in_range($ystep, $rowref->{'ystep_range'}))
-                && (!$rowref->{'zstep_range'} || &in_range($zstep, $rowref->{'zstep_range'}))
-                && (!$rowref->{'time_range'} || &in_range($time, $rowref->{'time_range'}))) {
+         	if ((!$rowref->{'TR_range'} || &in_min_max($tr, $rowref->{'TR_min'}, $rowref->{'TR_max'}))
```
cmadjar: Instead of creating a new function, could you simply recreate the range based on MIN-MAX and use the in_range function on this created range?

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -661,6 +661,8 @@ sub in_range
     my ($value, $range_string) = @_;
     chomp($value);
 
+    print "This function is deprecated and is replaced with in_min_max to identify the mri_protocol\n";
```
cmadjar: in_range function is being called in other instances where people want to check the range of something so I would not call it deprecated...

MounaSafiHarab: I did grep the codebase, and there is no occurrence of it except here (unless I missed something?)
There is actually **another** in_range function in `MNI/NumericUtilities.pm` which takes in min max as its arguments, but returns different codes...

but either way is fine for me. I was thinking maybe we send a deprecation messagr to the user now, and use a new function. Then in the next release, really remove the old in_range function...
but am open to whatever everyone thinks is the best way to proceed :)

uploadNeuroDB/NeuroDB/MRI.pm
MounaSafiHarab: I don't mind at all; not sure what is a better practice.
@nicolasbrossard @driusan ?

uploadNeuroDB/NeuroDB/MRI.pm
cmadjar: in_range is being used in the following places:
- [MRIProcessingUtility.pm line 514](https://github.com/aces/Loris-MRI/blob/9cf1f5e0f8d6f315a0fb321cf228b6fb6fed6dd4/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm#L514)  by the extra file checks
- [MRI.pm line 595](https://github.com/aces/Loris-MRI/blob/9cf1f5e0f8d6f315a0fb321cf228b6fb6fed6dd4/uploadNeuroDB/NeuroDB/MRI.pm#L595), although, for this one it is within the function `debug_inrange()` that does not seem to be used anywhere in the code...

My guess is that you probably grepped with the & in front of the in_range, which is probably why you did not see those two instances.

MounaSafiHarab: mmm, I actually did not use `&`, and I knew about the debug_in_range (which I think can be ignored), BUT the important one is the extra file checks... I did not see this one, so great catch, but now I wonder if this table too should not have ranges then!!!!

cmadjar: yes, it would make sense to normalize these ranges too then. Maybe in a separate PR?

MounaSafiHarab: yes I can do the changes, but it all depends on how it will be done on the Loris side...

MounaSafiHarab:
discussed at the Imaging meeting of March 16, 2018:
keep the function as `in_range`;

changes here are still pending on the `ValidRange` column on the Loris side, and the `mri_protocol_violated_scans` table changes, so the pull request here can be updated accordingly.

MounaSafiHarab:
@cmadjar @nicolasbrossard 
this can be tested now if you would like...

3 things we need to decide on:

1) waiting to see if @llevitis will split the ValidRange column in the protocol_checks table; in which case I need to incorporate this in the proper place
2) The new Min and Max columns allow NULL values now in the schema change of @llevitis, we need to decide if we should start allowing one OR the other to be set, in which case, I will need to augment the in_range function logic there. The other option would be if projects require no MIn we expect a value of `0` and if there is no Max check, we expect the highest value of `int(4)` to be put there in those min-max columns.
3) after discussions with @davidblader, maybe we are reconstructing the hack by adding the `-` back to the function now that @llevitis removed it from the database! do you think it is more proper to have the in_range() function takes in min and max as separate input variables? either way is fine for me. What I have in this pull request is less risky for sure but I feel is less clean...

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -529,8 +529,8 @@ sub identify_scan_db {
 
     
     # get the list of protocols for a site their scanner and subproject
-    $query = "SELECT Scan_type, ScannerID, Center_name, TR_range, TE_range, TI_range, slice_thickness_range, xspace_range, yspace_range, zspace_range,
-              xstep_range, ystep_range, zstep_range, time_range, series_description_regex
+    $query = "SELECT Scan_type, ScannerID, Center_name, TR_min, TR_max, TE_min, TE_max, TI_min, TI_max, slice_thickness_min, slice_thickness_max, xspace_min, xspace_max, yspace_min, yspace_max, zspace_min, zspace_max,
+              xstep_min, xstep_max, ystep_min, ystep_max, zstep_min, zstep_max, time_min, time_max, series_description_regex
```
cmadjar: Would be nice to have an error message if the query fails to tell the user that they need to update their `mri_protocol` table. What do you think?
screen shot 2018-05-28 at 4 22 04 pm

cmadjar:
> 1. waiting to see if @llevitis will split the ValidRange column in the protocol_checks table; in which case I need to incorporate this in the proper place

It would make sense to do the two tables at the same time.

> 2. The new Min and Max columns allow NULL values now in the schema change of @llevitis, we need to decide if we should start allowing one OR the other to be set, in which case, I will need to augment the in_range function logic there. The other option would be if projects require no MIn we expect a value of 0 and if there is no Max check, we expect the highest value of int(4) to be put there in those min-max columns.

The only thing I am worried about is that we really need to document correctly how to create the `mri_protocol` table to avoid any confusion (like a user setting only the MIN for a fix value (old TR_range=2300) instead of MIN and MAX to the same value (tr_min=2300, tr_max=2300). It has to be clear that:
- if MIN is set to null but not MAX, then, all values from 0 until the value in MAX are allowed (the value in MAX is included in the search)
- if MAX is set to null but not MIN, then, all values from MIN to infinity are allowed (the value in MIN is included in the search)
- if both MIN and MAX are null, then don't use any constraint on this header
- setting the MIN and MAX to the same value means look for the exact match
Probably very bad wording on my explanation above. This is how I understand the new table would work. Hope I am not too far from it ;).

> 3. after discussions with @davidblader, maybe we are reconstructing the hack by adding the - back to the function now that @llevitis removed it from the database! do you think it is more proper to have the in_range() function takes in min and max as separate input variables? either way is fine for me. What I have in this pull request is less risky for sure but I feel is less clean...

If @llevitis does the change in the mri_protocol_checks as well, then I agree, it would be cleaner to have MIN and MAX as argument to the function. However, if it is done separately, then your current fix seems the more appropriate until mri_protocol_checks is fixed.

MounaSafiHarab:
> waiting to see if @llevitis will split the ValidRange column in the protocol_checks table; in which case I need to incorporate this in the proper place
> It would make sense to do the two tables at the same time.
> 
Okay, will follow up with @llevitis on this in her LORIS PR then (I already commented on this before), but until hers is finalized, I can not finalize this :(

> The new Min and Max columns allow NULL values now in the schema change of @llevitis, we need to decide if we should start allowing one OR the other to be set, in which case, I will need to augment the in_range function logic there. The other option would be if projects require no MIn we expect a value of 0 and if there is no Max check, we expect the highest value of int(4) to be put there in those min-max columns.
> The only thing I am worried about is that we really need to document correctly how to create the mri_protocol table to avoid any confusion (like a user setting only the MIN for a fix value (old TR_range=2300) instead of MIN and MAX to the same value (tr_min=2300, tr_max=2300). It has to be clear that:
> 
> if MIN is set to null but not MAX, then, all values from 0 until the value in MAX are allowed (the value in MAX is included in the search)
I need to make sure this option works now (as it is a new 'feature').
> if MAX is set to null but not MIN, then, all values from MIN to infinity are allowed (the value in MIN is included in the search)
It is not infinity, it is the max of int(4) because this is the column type in the DB (I think). I need to make sure this option works now (as it is a new 'feature').

> if both MIN and MAX are null, then don't use any constraint on this header
> setting the MIN and MAX to the same value means look for the exact match
> Probably very bad wording on my explanation above. This is how I understand the new table would work. Hope I am not too far from it ;).

Okay, I can update the documentation section to make sure this is clarified, and at release time, in the release notes, we can also have a small section on this for existing projects. Where else can we document this?

> after discussions with @davidblader, maybe we are reconstructing the hack by adding the - back to the function now that @llevitis removed it from the database! do you think it is more proper to have the in_range() function takes in min and max as separate input variables? either way is fine for me. What I have in this pull request is less risky for sure but I feel is less clean...
> If @llevitis does the change in the mri_protocol_checks as well, then I agree, it would be cleaner to have MIN and MAX as argument to the function. However, if it is done separately, then your current fix seems the more appropriate until mri_protocol_checks is fixed.

Agreed. If the protocol checks table does not change, then I will keep it as is, if it does, then I will change to split MIN and MAX as two different input arguments.

uploadNeuroDB/NeuroDB/MRI.pm
cmadjar: sounds good!

cmadjar:
Discussed during the imaging meeting of August 16th, 2018.

With the Min Max replacing the hiphens in mri_protocol, we'll need to account for having Min or Max set as NULL. If Min or Max is set as NULL, then do an exact value match. If both Min and Max are set to NULL, then we ignore that parameter.

cmadjar:
All these changes have been integrated in the follow PR: https://github.com/aces/Loris-MRI/pull/339

PR #285, opened 2018-03-14
Improve minc_insertion.pl command line validation

nicolasbrossard:
This PR adds validation to the arguments passed on the command line and also gets rid of some warnings that were popping up when I tested this script. It also makes sure that the `globArchiveLocation` option behaves like it should by correcting the MySQL statement used to fetch a `tarchive` based on its `ArchiveSource`'s base name

### DIFF
--- a/uploadNeuroDB/minc_insertion.pl
+++ b/uploadNeuroDB/minc_insertion.pl
@@ -263,10 +263,16 @@ if ($tarchive) {
 
     if(!defined $is_valid) {
 		my $errorMessage = $globArchiveLocation
-		    ? "No mri_upload with the same archive location basename as '$tarchive'"
-		    : "No mri_upload with archive location '$tarchive'";
-		die "$errorMessage\n";
-	} 
+		    ? "No mri_upload with the same archive location basename as '$tarchive'\n"
+		    : "No mri_upload with archive location '$tarchive'\n";
+        $utility->writeErrorLog(
+                $errorMessage, $NeuroDB::ExitCodes::INVALID_ARG, $logfile
+        );
+        $notifier->spool('tarchive validation', $errorMessage, 0,
+                         'minc_insertion.pl', $upload_id, 'Y', 
+                         $notify_notsummary);
+        exit $NeuroDB::ExitCodes::INVALID_ARG;
+ 	} 
 	
     ## Setup  for the notification_spool table ##
     # get the tarchive_srcloc from $tarchive
@@ -278,10 +284,17 @@ if ($tarchive) {
     
     if(!defined $tarchive_srcloc) {
 		my $errorMessage = $globArchiveLocation
-		    ? "No tarchive with the same source location basename as '$tarchive'"
-		    : "No tarchive with source location '$tarchive'";
-		die "$errorMessage\n";
-	} 
+		    ? "No tarchive with the same source location basename as '$tarchive'\n"
+		    : "No tarchive with source location '$tarchive'\n";
+		    
+        $utility->writeErrorLog(
+                $errorMessage, $NeuroDB::ExitCodes::INVALID_ARG, $logfile
+        );
+        $notifier->spool('tarchive validation', $errorMessage, 0,
+                         'minc_insertion.pl', $upload_id, 'Y', 
+                         $notify_notsummary);
+        exit $NeuroDB::ExitCodes::INVALID_ARG;
+    } 
 
     # get the $upload_id from $tarchive_srcloc
     $query =
--- a/uploadNeuroDB/minc_insertion.pl
+++ b/uploadNeuroDB/minc_insertion.pl
@@ -149,7 +149,7 @@ USAGE
 
 if (!$ENV{LORIS_CONFIG}) {
 	print STDERR "\n\tERROR: Environment variable 'LORIS_CONFIG' not set\n\n";
-	exit 1;  # TODO replace by appropriate exit code from ExitCodes once it is defined
+	exit $NeuroDB::ExitCodes::INVALID_ENVIRONMENT_VAR; 
 }
 
 if (!defined $profile || !-e "$ENV{LORIS_CONFIG}/.loris_mri/$profile") {
--- a/uploadNeuroDB/minc_insertion.pl
+++ b/uploadNeuroDB/minc_insertion.pl
@@ -145,43 +145,46 @@ usage: $0 </path/to/DICOM-tarchive> [options]
 
 USAGE
 &Getopt::Tabular::SetHelp($Help, $Usage);
-&Getopt::Tabular::GetOptions(\@opt_table, \@ARGV) || exit 1;
+&Getopt::Tabular::GetOptions(\@opt_table, \@ARGV) || exit $NeuroDB::ExitCodes::GETOPT_FAILURE;
 
+if (!$ENV{LORIS_CONFIG}) {
+	print STDERR "\n\tERROR: Environment variable 'LORIS_CONFIG' not set\n\n";
+	exit 1;  # TODO replace by appropriate exit code from ExitCodes once it is defined
+}
+
+if (!defined $profile || !-e "$ENV{LORIS_CONFIG}/.loris_mri/$profile") {
+    print $Help; 
+    print STDERR "$Usage\n\tERROR: You must specify a valid and existing profile.\n\n";  
+    exit $NeuroDB::ExitCodes::PROFILE_FAILURE;
+}
+
+if (defined $tarchive && !(-e $tarchive) ) {
+    print STDERR "\nERROR: Could not find archive $tarchive. \nPlease, make sure "
+        . " the path to the archive is correct. Upload will exit now.\n\n\n";
+    exit $NeuroDB::ExitCodes::ARG_FILE_DOES_NOT_EXIST;
+}
+
+if ( !($tarchive xor $upload_id) ) {
+    print STDERR "\nERROR: You should either specify an upload ID or a tarchive "
+        . "path. Make sure that you set only one of those options. "
+        . "Upload will exit now.\n\n\n";
+    exit $NeuroDB::ExitCodes::MISSING_ARG;
+}
+
+if (!defined $minc || !-e $minc) {
+    print STDERR "$Usage\n\tERROR: You must specify a valid and existing "
+        . "MINC file with -minc.\n\n";  
+    exit $NeuroDB::ExitCodes::ARG_FILE_DOES_NOT_EXIST;
+}
 
 # input option error checking
 { package Settings; do "$ENV{LORIS_CONFIG}/.loris_mri/$profile" }
 
-if ( !$profile ) {
-    print $Help;
-    print STDERR "$Usage\n\tERROR: missing -profile argument\n\n";
-    exit $NeuroDB::ExitCodes::PROFILE_FAILURE;
-}
 if ( !@Settings::db ) {
     print STDERR "\n\tERROR: You don't have a \@db setting in the file "
                  . "$ENV{LORIS_CONFIG}/.loris_mri/$profile \n\n";
     exit $NeuroDB::ExitCodes::DB_SETTINGS_FAILURE;
 }
-
-if ( !($tarchive xor $upload_id) ) {
-    print "\nERROR: You should either specify an upload ID or a tarchive ".
-        "path. Make sure that you set only one of those options. ".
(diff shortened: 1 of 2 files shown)


uploadNeuroDB/minc_insertion.pl
```diff
@@ -145,44 +145,44 @@
...
-{ package Settings; do "$ENV{LORIS_CONFIG}/.loris_mri/$profile" }
-
-
-if ($profile && !@Settings::db) { 
-    print "\n\tERROR: You don't have a ".
-    "configuration file named '$profile' in:  $ENV{LORIS_CONFIG}/.loris_mri/ \n\n";
-    exit 2; 
+if(!$ENV{LORIS_CONFIG}) {
+	print STDERR "\n\tERROR: Environment variable 'LORIS_CONFIG' not set\n\n";
+	exit 1;  # TODO replace by appropriate exit code from ExitCodes once it is defined
```
MounaSafiHarab: do you think it is worth it to change exit codes on the user given that they will be changed again in the Harmonize Exit codes PR?

uploadNeuroDB/minc_insertion.pl
```diff
@@ -145,44 +145,44 @@
...
 }
 
-if (!$profile) { 
+if (!defined$profile || !-e "$ENV{LORIS_CONFIG}/.loris_mri/$profile") {
```
MounaSafiHarab: do we need a space between `defined` and `$profile`?
just out of curiosity, I would tend to put `()` around `$profile` because it makes it more readable to me, but is this not needed or not desirable?

uploadNeuroDB/minc_insertion.pl
```diff
@@ -145,44 +145,44 @@
...
     print $Help; 
-    print "$Usage\n\tERROR: You must specify a valid ".
-    "and existing profile.\n\n";  
-    exit 3; 
+    print STDERR "$Usage\n\tERROR: You must specify a valid and existing profile.\n\n";  
+    exit 2; #TODO 1: use exit code from ExitCodes once it is merged
```
MounaSafiHarab: same comment here

uploadNeuroDB/minc_insertion.pl
```diff
@@ -145,44 +145,44 @@
...
 
+# input option error checking
+{ package Settings; do "$ENV{LORIS_CONFIG}/.loris_mri/$profile" }
 
-
+if (!@Settings::db) { 
+    printf STDERR
+        "\n\tERROR: no database setting '\@db' found in %s\n",   
+        "$ENV{LORIS_CONFIG}/.loris_mri/$profile or file unreadable";
+    exit 4; #TODO 1: use exit code from ExitCodes once it is merged
```
MounaSafiHarab: are theswe changes all going to conflict with the Exit code harmonization?!

uploadNeuroDB/minc_insertion.pl
nicolasbrossard: I usually put a space after `defined` but I guess I forgot to put one this time. I am actually surprised that this even works! But I tested again and this is valid. Adding parentheses around the argument is also valid and is entirely a matter of programming style. I'll add a space.

uploadNeuroDB/minc_insertion.pl
nicolasbrossard: Actually, the numerical values used in the code that I corrected are in line with the intended usage of each code defined in `ExitCode.pm`:

```perl
our $PROFILE_FAILURE         = 2; # if no profile file specified
our $MISSING_ARG             = 3; # if missing script's argument(s)
our $DB_SETTINGS_FAILURE     = 4; # if DB settings in profile file are not set
our $ARG_FILE_DOES_NOT_EXIST = 5; # if file given as an argument does not exist in the file system
```
The only one I added was error code `1` for the missing environment variable. But we can add it in `ExitCode.pm`.

uploadNeuroDB/minc_insertion.pl
nicolasbrossard: See my comment on exit codes

uploadNeuroDB/minc_insertion.pl
nicolasbrossard: See my comment on exit codes

uploadNeuroDB/minc_insertion.pl
```diff
@@ -145,44 +145,45 @@
...
-    print "\n\tERROR: You don't have a ".
-    "configuration file named '$profile' in:  $ENV{LORIS_CONFIG}/.loris_mri/ \n\n";
-    exit 2; 
+if(!$ENV{LORIS_CONFIG}) {
+	print STDERR "\n\tERROR: Environment variable 'LORIS_CONFIG' not set\n\n";
+	exit 1;  # TODO replace by appropriate exit code from ExitCodes once it is defined
 }
 
-if (!$profile) { 
+if (!defined $profile || !-e "$ENV{LORIS_CONFIG}/.loris_mri/$profile") {
```
cmadjar: 👍 so much better!!!

MounaSafiHarab:
> @nicolasbrossard Looks good! The only thing is that it might be worth to wait that the ExitCode one is merged so you can rebase your PR as some of the changes you made have also been done in the exitcode PR (some print STDERR for example)

now that it is finally merged, might be worth rebasing!

uploadNeuroDB/minc_insertion.pl
```diff
@@ -145,44 +145,45 @@
 &Getopt::Tabular::SetHelp($Help, $Usage);
 &Getopt::Tabular::GetOptions(\@opt_table, \@ARGV) || exit 1;
```
cmadjar: @nicolasbrossard I just realize I missed one exit code here. 
`exit 1` should actually be `exit $NeuroDB::ExitCodes::GETOPT_FAILURE`

Do you mind making the change at the same time as you PR?

uploadNeuroDB/minc_insertion.pl
cmadjar: @nicolasbrossard Now that the ExitCode has been merged, maybe you could use them in this PR?

uploadNeuroDB/minc_insertion.pl
nicolasbrossard: Done.

uploadNeuroDB/minc_insertion.pl
```diff
@@ -250,20 +252,35 @@
 
     my $where = "WHERE t.ArchiveLocation='$tarchive'";
     if ($globArchiveLocation) {
-        $where = "WHERE t.ArchiveLocation LIKE '%" . basename($tarchive) . "'";
+        $where = "WHERE t.ArchiveLocation LIKE '%/" . basename($tarchive) . "'";
```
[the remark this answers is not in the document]
nicolasbrossard: Good point. I corrected this so it handles all cases.

@MounaSafiHarab @cmadjar ready for another round of review/testing

uploadNeuroDB/minc_insertion.pl
```diff
@@ -145,43 +145,46 @@
 
 USAGE
 &Getopt::Tabular::SetHelp($Help, $Usage);
-&Getopt::Tabular::GetOptions(\@opt_table, \@ARGV) || exit 1;
+&Getopt::Tabular::GetOptions(\@opt_table, \@ARGV) || exit $NeuroDB::ExitCodes::GETOPT_FAILURE;
 
+if (!$ENV{LORIS_CONFIG}) {
+	print STDERR "\n\tERROR: Environment variable 'LORIS_CONFIG' not set\n\n";
+	exit 1;  # TODO replace by appropriate exit code from ExitCodes once it is defined
```
MounaSafiHarab: why not define it now?

nicolasbrossard: The $ENV{'LORIS_CONFIG'} check should be done in *all* scripts if we want to be consistent. I was thinking of issuing a PR that does just that (along with creating a new entry for the exit code in ExitCode.pm). Agreed?

MounaSafiHarab: sounds good to me. It might be a good idea to just define it here as a variable in the ExitCodes.pm class, and re-use that code in other scripts at a later time
(simply to avoid telling the users at release time that exit 1 again changed meaning)....

nicolasbrossard: My next commit covers this.

uploadNeuroDB/minc_insertion.pl
```diff
@@ -249,20 +253,36 @@
 
     my $where = "WHERE t.ArchiveLocation='$tarchive'";
     if ($globArchiveLocation) {
-        $where = "WHERE t.ArchiveLocation LIKE '%" . basename($tarchive) . "'";
+        $where = "WHERE t.ArchiveLocation LIKE '%/" . quotemeta(basename($tarchive)) . "' "
+            .    "OR t.ArchiveLocation LIKE '" . quotemeta(basename($tarchive)) . "'";
```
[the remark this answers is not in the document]
nicolasbrossard: Both parts are needed actually. With `ArchiveLocation = '2011/loc'` and `$tarchive = '9999/loc'`, the first part matches but the second one does not. With `ArchiveLocation = 'loc'` and `$tarchive ='9999/loc'`, the firsrt part does not match but the second one does.

uploadNeuroDB/minc_insertion.pl
```diff
@@ -249,20 +253,36 @@
...
     my $query = "SELECT m.IsTarchiveValidated FROM mri_upload m " .
         "JOIN tarchive t on (t.TarchiveID = m.TarchiveID) $where ";
     print $query . "\n" if $debug;
     $is_valid = $dbh->selectrow_array($query);
 
+    if(!defined $is_valid) {
+		my $errorMessage = $globArchiveLocation
+		    ? "No mri_upload with the same archive location basename as '$tarchive'"
+		    : "No mri_upload with archive location '$tarchive'";
+		die "$errorMessage\n";
```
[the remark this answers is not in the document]
nicolasbrossard: I see your point. I'll change this.

uploadNeuroDB/minc_insertion.pl
```diff
@@ -249,20 +253,36 @@
...
         " FROM tarchive t $where";
     my $sth = $dbh->prepare($query);
     $sth->execute();
     $tarchive_srcloc = $sth->fetchrow_array;
+    
+    if(!defined $tarchive_srcloc) {
+		my $errorMessage = $globArchiveLocation
+		    ? "No tarchive with the same source location basename as '$tarchive'"
+		    : "No tarchive with source location '$tarchive'";
+		die "$errorMessage\n";
```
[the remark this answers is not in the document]
nicolasbrossard: Ditto

uploadNeuroDB/minc_insertion.pl
```diff
@@ -291,12 +311,11 @@
     $ArchiveLocation = $array[1];
 }
 
-if (($is_valid == 0) && ($force==0)) {
+if ((!defined $is_valid || $is_valid == 0) && !$force) {
     $message = "\n ERROR: The validation has failed. ".
```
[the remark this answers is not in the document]
nicolasbrossard: The check on line 264 is done only when you use the -tarchivePath option. You still need this check for when the -uploadId option is used.

uploadNeuroDB/minc_insertion.pl
```diff
@@ -145,43 +145,46 @@
...
-    print "\nERROR: Could not find archive $tarchive. \nPlease, make sure ".
-          " the path to the archive is correct. Upload will exit now.\n\n\n";
-    exit $NeuroDB::ExitCodes::ARG_FILE_DOES_NOT_EXIST;
-}
-unless (-e $minc) {
-    print STDERR "\nERROR: Could not find minc $minc.\n"
-                 . "Please, make sure the path to the MINC file is valid.\n\n";
+if (!defined $minc || !-e $minc) {
+    print STDERR "$Usage\n\tERROR: You must specify a valid and existing "
+        . "MINC file with -minc.\n\n";
```
MounaSafiHarab: is this -minc or -mincPath? just a minor detail

uploadNeuroDB/minc_insertion.pl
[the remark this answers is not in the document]
MounaSafiHarab: @nicolasbrossard @cmadjar 
I am not sure the second case is possible; i.e. having the ArchiveLocation not having the year in it, while the tarchive has it?

PR #287, opened 2018-04-04
Refactoring of MriProcessingUtility::getUploadIDUsingTarchiveSrcLoc()

nicolasbrossard:
Refactored method `getUploadIDUsingTarchiveSrcLoc()` in module `MriProcessingUtility` so it uses the new `Database.pm` module. All scripts that create an instance of `MriProcessingUtility` were also updated: they now create a NeuroDB::Database object and pass it to the `MriProcessingUtility` constructor, which can then use it for specific method calls. The old (DBI style) database handle object is kept for now but will eventually disappear once all modules/scripts have been updated.


nicolasbrossard:
Oh no...This PR makes no sense then...Closing.

PR #288, opened 2018-04-05
Remove links to scripts in uploadNeuroDB/bin and update environment file

cmadjar:
Sent a new PR for the 20.0 release. See #320

### Spring cleaning!

Instead of having links in the main directory to the scripts stored in uploadNeuroDB/bin, add the path to uploadNeuroDB/bin in the environment file.

### Caveat for existing projects:

Need to add in your environment file the following path at the beginning of the export PATH line:
/data/%PROJECT%/bin/mri/uploadNeuroDB/bin
where %PROJECT% should be replaced by the name of your project

### Note: 

Need to add this uploadNeuroDB/bin path at the beginning of the PATH variable to make sure your project is using the mincpik from LORIS-MRI instead of the mincpik from the MINC tools.

PR #289, opened 2018-04-05
Adding exit codes to uploadNeuroDB/bin scripts

cmadjar:
### Added the exit codes to the following scripts in uploadNeuroDB/bin:
- concat_mri.pl
- minc2jiv.pl

### Notes: 
- no exits are present in Mincinfo_wrapper
- did not do mincpik as its future might be compromised once we discuss PR #288. Can always do it later on.

Link to redmine ticket: https://redmine.cbrain.mcgill.ca/issues/14054

### DIFF
--- a/uploadNeuroDB/NeuroDB/ExitCodes.pm
+++ b/uploadNeuroDB/NeuroDB/ExitCodes.pm
@@ -73,6 +73,9 @@ to 179)
 
 14. Exit codes from uploadNeuroDB/tarchiveLoader (exit codes from 260 to 279)
 
+15. Exit codes from uploadNeuroDB/NeuroDB/bin/minc2jiv.pl (exit codes from 280
+to 300)
+
 
 =head1 LICENSING
 
--- a/uploadNeuroDB/NeuroDB/ExitCodes.pm
+++ b/uploadNeuroDB/NeuroDB/ExitCodes.pm
@@ -105,8 +105,10 @@ our $ARG_FILE_DOES_NOT_EXIST = 5; # if file given as an argument does not exist
 # database related failure
 # called from minc_insertion.pl & register_processed_data.pl
 our $FILE_NOT_UNIQUE = 6; # if file to register is not unique & already
-                           # inserted
+                          # inserted
 
+# other generic failure
+our $FILE_OR_FOLDER_DOES_NOT_EXIST = 6; # if file or folder does not exist
 
 
 
@@ -313,3 +315,12 @@ our $NO_VALID_MINC_CREATED = 261; # if no valid MINC file was created
                                   # (non-localizers)
 our $NO_MINC_INSERTED      = 262; # if no MINC files was inserted (invalid
                                   # study)
+
+
+
+
+
+
+#### --- FROM uploadNeuroDB/NeuroDB/bin/minc2jiv.pl
+
+our $REGISTER_PROGRAM_FAILURE = 280; # if MNI::Spawn::RegisterPrograms failed
--- a/uploadNeuroDB/NeuroDB/ExitCodes.pm
+++ b/uploadNeuroDB/NeuroDB/ExitCodes.pm
@@ -111,7 +111,7 @@ our $FILE_NOT_UNIQUE = 6; # if file to register is not unique & already
                           # inserted
 
 # other generic failure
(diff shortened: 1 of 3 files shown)


uploadNeuroDB/NeuroDB/ExitCodes.pm
```diff
@@ -105,8 +108,10 @@ our $ARG_FILE_DOES_NOT_EXIST = 5; # if file given as an argument does not exist
 # database related failure
 # called from minc_insertion.pl & register_processed_data.pl
 our $FILE_NOT_UNIQUE = 6; # if file to register is not unique & already
-                           # inserted
+                          # inserted
 
+# other generic failure
+our $FILE_OR_FOLDER_DOES_NOT_EXIST = 6; # if file or folder does not exist
```
MounaSafiHarab: just curious at using exit code =6 when 2 lines before we have the same one? is this intentional?

uploadNeuroDB/bin/minc2jiv.pl
```diff
@@ -93,7 +94,8 @@
 my $compress= ($gzip ? "| gzip -c9 " : "") ;
 
 my %base_names_seen;
-MNI::FileUtilities::check_output_path("$TmpDir/") or exit 1;
+MNI::FileUtilities::check_output_path("$TmpDir/")
```
MounaSafiHarab: wouldn't exit code 5 also work here?

MounaSafiHarab:
@cmadjar 
I put 2 comments earlier
and more importantly, this needs rebasing now!

uploadNeuroDB/NeuroDB/ExitCodes.pm
cmadjar: Oups, you're right, it should have been 7. I did not use the exit 5 as the exit 5 is specific to files passed as an *argument* that do not exist and here, the file or folder was not provided as an argument. Does that make sense?

uploadNeuroDB/bin/minc2jiv.pl
cmadjar: exit code 5 is specific to files provided as argument to the script which is why I did not use it here.

cmadjar:
@MounaSafiHarab I had to rebase it and change the code number to 9 as there was a conflict with the minor branch. Fixed now. I might have to do the same for the other exit code PR.

nicolasbrossard:
Makes sense to me.

MounaSafiHarab:
> @nicolasbrossard @MounaSafiHarab I propose that we merge this PR if there is no other issue than the fact that exit code should be between 0-255.
> 
> In the meantime, I am going to create a spreadsheet that I will share with you to propose some regroupment of the exit codes to make them more concise. Since the exit code will go to 19.1, it might be good to have the final version in that release. What do you think?

I would for sure say we should aim for 19.1 as the main change is going in there.
I am not sure how much work is involved, but if it doable, for sure we should aim for it (we do not want to tell the users the exit codes changed and then changed again in a next release, if we can).
Just got in, so I will read your proposal!

cmadjar:
> I would for sure say we should aim for 19.1 as the main change is going in there.
I am not sure how much work is involved, but if it doable, for sure we should aim for it (we do not want to tell the users the exit codes changed and then changed again in a next release, if we can).
Just got in, so I will read your proposal!

I totally agree with you. As soon as we agree on the exit codes, I'll make the change and send a PR. I don't think it will be a long one to do. :)

MounaSafiHarab:
> I totally agree with you. As soon as we agree on the exit codes, I'll make the change and send a PR. I don't think it will be a long one to do. :)

sounds great!
I put a couple of comments on the excel sheet; just minor ones to make things even more generic; perhaps @nicolasbrossard can comment there too.
We can also meet briefly in the next day or 2 if you are here if it makes this faster for you!

MounaSafiHarab:
and yes, I agree we cna merge this and make the change in one other complete PR
@nicolasbrossard what do you think? you had requested changes, so we can not move forward until you dismiss/resolve those.

nicolasbrossard:
Ok, I'll approve this PR and we'll tackle the refactoring of the exit codes in another one.

MounaSafiHarab:
ok so merging and will wait for the new PR to hopefully merge it onto this same release.

PR #291, opened 2018-04-05
Modified dcmdump command to grep the patient information from the Config module

cmadjar:
In ImagingUpload.pm, the PatientName was hard coded during the validation of the DICOM study. 
Instead of hardcoding this, it is now using the lookupCenterNameUsing Config option when performing the validation of the DICOM study.

In addition, as @driusan suggested during an imaging meeting, modified the dcmdump command from `dcmdump $dicom | grep` to `dcmdump +P $lookupCenterNameUsing $dicom` which gives the same result but is a bit cleaner.

Associated redmine ticket: https://redmine.cbrain.mcgill.ca/issues/14088

### DIFF
--- a/uploadNeuroDB/NeuroDB/ImagingUpload.pm
+++ b/uploadNeuroDB/NeuroDB/ImagingUpload.pm
@@ -10,6 +10,7 @@ use File::Find;
 use NeuroDB::FileDecompress;
 use NeuroDB::Notify;
 use NeuroDB::ExitCodes;
+use NeuroDB::DBI;
 use File::Temp qw/ tempdir /;
 
 
@@ -410,7 +411,12 @@ Arguments:
 sub PatientNameMatch {
     my $this         = shift;
     my ($dicom_file) = @_;
-    my $cmd          = "dcmdump $dicom_file | grep PatientName";
+
+    my $lookupCenterNameUsing = NeuroDB::DBI::getConfigSetting(
+        $this->{'dbhr'},'lookupCenterNameUsing'
+    );
+
+    my $cmd = "dcmdump +P $lookupCenterNameUsing $dicom_file";
     my $patient_name_string =  `$cmd`;
     if (!($patient_name_string)) {
 	my $message = "\nThe patient name cannot be extracted \n";


uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -410,7 +411,12 @@ Arguments:
 sub PatientNameMatch {
     my $this         = shift;
     my ($dicom_file) = @_;
-    my $cmd          = "dcmdump $dicom_file | grep PatientName";
+
+    my $lookupCenterNameUsing = NeuroDB::DBI::getConfigSetting(
+        $this->{'dbhr'},'lookupCenterNameUsing'
+    );
+
+    my $cmd = "dcmdump +P $lookupCenterNameUsing $dicom_file";
```
nicolasbrossard: \<paranoid\>
`my $cmd = sprintf("dcmdump +p %s %s", quotemeta($lookupCenterNameUsing), quotemeta($dicom _file));`
\</paranoid\>

cmadjar: Cool! Thanks!! Always good to be paranoid for these kind of things!!

MounaSafiHarab:
putting this in 19.1 as well as it is a bugfix...
will test shortly

uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -410,7 +411,14 @@ Arguments:
...
     my $this         = shift;
     my ($dicom_file) = @_;
-    my $cmd          = "dcmdump $dicom_file | grep PatientName";
+
+    my $lookupCenterNameUsing = NeuroDB::DBI::getConfigSetting(
+        $this->{'dbhr'},'lookupCenterNameUsing'
+    );
+
+    my $cmd = sprintf("dcmdump +P %s %s",
+        quotemeta($lookupCenterNameUsing), quotemeta($dicom_file)
```
MounaSafiHarab: a small comment here:
since the identifier can be patient name or patient ID; then a few lines later, we are specifically telling the user that the patient name can not be extracted ; then 2 lines later, we also say the patient name read the header does not start with ...
do you think it makes sense to make the error message more generic?

cmadjar: Very good point!! Thank you for bringing that up. I just made it more generic. :)

uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -410,7 +411,14 @@ Arguments:
 sub PatientNameMatch {
     my $this         = shift;
     my ($dicom_file) = @_;
-    my $cmd          = "dcmdump $dicom_file | grep PatientName";
+
+    my $lookupCenterNameUsing = NeuroDB::DBI::getConfigSetting(
+        $this->{'dbhr'},'lookupCenterNameUsing'
+    );
```
MounaSafiHarab: @cmadjar @nicolasbrossard 

now that we are getting it from the database, shouldn't we check if this setting is defined before running dcmdump?

nicolasbrossard: I think we should. The new DB framework will check that by the way (it will throw an exception if the setting was not found).

MounaSafiHarab: if I put the word 'test' in that Config setting (which clearly does not exist in DICOMs), running the script would give me this error message:

```
W: unrecognized tag name: 'test'
Spool message is: 
The patient name cannot be extracted 
```

MounaSafiHarab: also in the Loris front end, this setting is called Lok UpCenterName using because the getPSC() uses it, and this is what the hover over says. Shouldn't it somehow also reflect that we are now going to use it to validate DICOMs based on the proper Loris identifier?

uploadNeuroDB/NeuroDB/ImagingUpload.pm
MounaSafiHarab: on another note; the fact u added `sprintf`; does that mean this command will print everytime it is executing?
it did not when tested it, but I was just curious...

uploadNeuroDB/NeuroDB/ImagingUpload.pm
cmadjar: oooooh, I did not think of testing a random string for the Config setting. Thanks!! 👍 

So, if I add the -q option to the dcmdump command, then you don't see the error from dcmdump but you can get into the if statement that checks if a value was returned. So now, if I put 'test' in the lookupCenterNameUsing config setting, it will print a nice error message:
```
The 'test' DICOM field cannot be extracted from the DICOM file /data/not_backed_up/ImagingUpload-11-47-w6aIf0/MTL0609_138233_PREBL00_20170410_154459088_adniT1/MR.1.3.12.2.1107.5.2.32.35442.2017041015351355915460270
```
But if I have a correct DICOM field, then it runs properly.

uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -419,7 +427,7 @@ sub PatientNameMatch {
     }
     my ($l,$pname,$t) = split /\[(.*?)\]/, $patient_name_string;
     if ($pname !~ /^$this->{'pname'}/) {
-        my $message = "\nThe patient-name read ".
+        my $message = "\nThe $lookupCenterNameUsing read ".
```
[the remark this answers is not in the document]
cmadjar: oups. Did not see that one...

cmadjar:
@nicolasbrossard @MounaSafiHarab When testing other things, I realized I did not test the rest of the pipeline so I will tackle this this PM and let you know when it is ready for another round of review. Stay tuned!

uploadNeuroDB/NeuroDB/ImagingUpload.pm
cmadjar: @MounaSafiHarab I just sent a PR on the LORIS side to change the description of the Config setting. Here is the PR: https://github.com/aces/Loris/pull/3585

Thanks!

cmadjar:
@MounaSafiHarab @nicolasbrossard I think I took care of all your comments. The bug I was talking about comes from another PR that I mixed up, so forget the comment I made 2 hours ago...

Let me know if you see something else.

cmadjar:
Added the "Needs work before merging" label as I will need to rebase the branch to get the ExitCode from PR #297 and add the MISSING_CONFIG_SETTING to the ExitCode library.

cmadjar:
> you only check if it is set with the unless clause, but isn't it also worthwhile to check that it IS PatientName or PatientID? this is what the hover over explains.. unless projects use other DICOM tags?

@MounaSafiHarab Now that I think of it, shouldn't this be a drop down between PatientID and PatientName in the Config module?

@cmadjar just so I remember, still need to implement $MISSING_CONFIG_SETTING and $BAD_CONFIG_SETTING to the exit code library once  PR #297 is merged.

uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -410,19 +411,46 @@ Arguments:
...
+        $this->spool($message, 'Y', $notify_notsummary);
+        #TODO: once refactoring of exit code merged to minor, add the following
+        #TODO: MISSING_CONFIG_SETTING exit code in the generic database code
+        exit $NeuroDB::ExitCodes::MISSING_CONFIG_SETTING;
+    }
+
+    unless ($lookupCenterNameUsing =~ /PatientName|PatientID/i) {
+        my $message = "\nConfig setting 'lookupCenterNameUsing' is set to "
+                      . "$lookupCenterNameUsing but should be set to "
+                      . "either to either PatientID or PatientName";
```
nicolasbrossard: Stuttering alert (to either to either...)!   :)

uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -410,19 +411,46 @@ Arguments:
...
+    unless ( $lookupCenterNameUsing ) {
+        my $message = "\nConfig Setting 'lookupCenterNameUsing' is not set in "
+                      . "the Config module under the Imaging Pipeline section.";
+        $this->spool($message, 'Y', $notify_notsummary);
+        #TODO: once refactoring of exit code merged to minor, add the following
+        #TODO: MISSING_CONFIG_SETTING exit code in the generic database code
+        exit $NeuroDB::ExitCodes::MISSING_CONFIG_SETTING;
+    }
+
+    unless ($lookupCenterNameUsing =~ /PatientName|PatientID/i) {
```
nicolasbrossard: Since we expected either 'PatientName' or 'PatientID', shouldn't the pattern check be `/^(PatientName|PatientID)$/i`?

uploadNeuroDB/NeuroDB/ImagingUpload.pm
MounaSafiHarab: lol

MounaSafiHarab:
@cmadjar 

#297 is now merged, could you please rebase this one and fix the exit code?

cmadjar:
> @cmadjar
> #297 is now merged, could you please rebase this one and fix the exit code?

@MounaSafiHarab @nicolasbrossard Done integrating the exit codes! 😄

uploadNeuroDB/NeuroDB/ExitCodes.pm
```diff
@@ -102,13 +102,14 @@ our $INVALID_ARG          = 6; # if one of the program arguments is invalid
...
-our $SELECT_FAILURE  = 23; # if a SELECT query did not return anything
-our $UPDATE_FAILURE  = 24; # if an UPDATE query failed
+our $FILE_NOT_UNIQUE    = 20; # if file to register is not unique & already
+                              # inserted
+our $INSERT_FAILURE     = 21; # if an INSERT query failed
+our $CORRUPTED_FILE     = 22; # if mismatch between the file's md5sum and the
+                              # hash stored in the database
+our $SELECT_FAILURE     = 23; # if a SELECT query did not return anything
+our $UPDATE_FAILURE     = 24; # if an UPDATE query failed
+our $BAD_CONFIG_SETTING = 25; # if bad config setting
```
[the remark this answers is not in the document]
cmadjar: I don't think so as I am not changing the documentation in this PR. I only added an exit code (25) that is still in the range of what was documented (see below)

`3. Common database related failures (exit codes from 20 to 39)` 

But thanks for pointing that out. :)

uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -410,19 +411,44 @@ Arguments:
...
+        $this->{'dbhr'},'lookupCenterNameUsing'
+    );
+
+    unless ( $lookupCenterNameUsing ) {
+        my $message = "\nConfig Setting 'lookupCenterNameUsing' is not set in "
+                      . "the Config module under the Imaging Pipeline section.";
+        $this->spool($message, 'Y', $notify_notsummary);
+        #TODO: once refactoring of exit code merged to minor, add the following
+        #TODO: MISSING_CONFIG_SETTING exit code in the generic database code
+        exit $NeuroDB::ExitCodes::MISSING_CONFIG_SETTING;
```
nicolasbrossard: This still has to be defined in ExitCode.pm

uploadNeuroDB/NeuroDB/ExitCodes.pm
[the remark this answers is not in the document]
MounaSafiHarab: Okay, just realized that the .pm only has the header not more details (i.e. no exit code numbers in it!); so it is all good thanks a lot!

uploadNeuroDB/NeuroDB/ImagingUpload.pm
MounaSafiHarab: good point @nicolasbrossard 
I did not realize this is what you have been telling me since yesterday!

otherwise, I tested it (by manually adding that missing exit code myself based on your comment), and it works...
so as soon as @cmadjar updates her PR, I will add the passed manual test tag

cmadjar:
@MounaSafiHarab @nicolasbrossard Sorry. Fixed now. Weird that I missed this one...

PR #292, opened 2018-04-05
Removing PendingStaging field from the files table

cmadjar:
This pull request removes the unused PendingStaging field from the files table as was discussed during the LORIS imaging meeting of February 2nd, 2018.

See also:
- associated redmine ticket: https://redmine.cbrain.mcgill.ca/issues/13876
- associate PR on the LORIS side: https://github.com/aces/Loris/pull/3569

Note for reviewers: 
I removed all the code associated to this PendingStaging field. I checked IBIS and PREVENT-AD, this field is always filled with 0. Maybe you could check your DB just to be sure but at this point, I am fairly confident this is not used.


cmadjar:
@nicolasbrossard Good catch! I updated the description of the getSessionID function. Ready for another round of review or for approval ;)

PR #293, opened 2018-04-06
Fix tarchive being named without the patient name

cmadjar:
When using the imaging uploader to upload a DICOM study, the `tarchive` used to be named as follows:
`DCM_2018-01-26_ImagingUpload-$hour-$minute-XXXXXX.tar`

Now the `tarchive` will include the patient name information as follows:
`DCM_2018-01-26_ImagingUpload-$hour-$minute-XXXXXX_$PatientName.tar`

Here are two examples of how the `tarchive` will be named depending on whether `dicomTar.pl` was run directly on the folder `MTL0021_917380_PREEN00_20120730_144819638` (first line) or via the imaging uploader (second line): 
```
DCM_2012-07-30_MTL0021_917380_PREEN00_20120730_144819638_MTL0021_917380_PREEN00.tar
DCM_2017-04-10_ImagingUpload-13-39-cTsG8e_MTL0609_138233_PREBL00.tar
```

Associated Redmine ticket: https://redmine.cbrain.mcgill.ca/issues/13933

I added the flag "add to release note" to let people know that the way we name the tarchive has changed.

### DIFF
--- a/dicom-archive/DICOM/DCMSUM.pm
+++ b/dicom-archive/DICOM/DCMSUM.pm
@@ -75,11 +75,11 @@ sub new {
 =cut
 
 sub database {
-    my $self   = shift;
-    my $dbh    = shift;
-    my $meta   = shift;
-    my $update = shift;
-    
+    my $self     = shift;
+    my $dbh      = shift;
+    my $metafile = shift;
+    my $update   = shift;
+
     # these are only available if you run dicomTar
     my $tarType     = shift;
     my $tarLog      = shift;
@@ -133,8 +133,7 @@ QUERY
 
     # INSERT or UPDATE 
     # get acquisition metadata
-    my $sfile = "$self->{tmpdir}/$meta.meta";
-    my $metacontent = &read_file($sfile);
+    my $metacontent = &read_file($metafile);
     
     (my $common_query_part = <<QUERY) =~ s/\n/ /gm;  
       tarchive SET  
--- a/dicom-archive/dicomTar.pl
+++ b/dicom-archive/dicomTar.pl
@@ -101,7 +101,7 @@ if (-d $dcm_source && -d $targetlocation) {
     exit $NeuroDB::ExitCodes::ARG_FILE_DOES_NOT_EXIST;
 }
 
-# The tar target 
+# The tar target
 my $totar = basename($dcm_source);
 print "Source: ". $dcm_source . "\nTarget: ".  $targetlocation . "\n\n" if $verbose;
 my $ARCHIVEmd5sum = 'Provided in database only';
@@ -137,10 +137,11 @@ my $studyUnique = $summary->{'studyuid'};
 my $creator         = $summary->{user};
 my $sumTypeVersion  = $summary->{sumTypeVersion}; 
 
-my $byDate;
 # Determine how to name the archive... by acquisition date or by today's date.
-if ($todayDate) { $byDate = $today; } else { $byDate = $summary->{header}->{scandate}; } # wrap up the archive 
-my $finalTarget = "$targetlocation/DCM_${byDate}_$summary->{metaname}.tar";
+my $byDate      = $todayDate ? $today : $summary->{header}->{scandate};
+my $pname       = $summary->{header}->{pname};
+my $tarbasename = "${metaname}_$pname";
+my $finalTarget = "$targetlocation/DCM_${byDate}_$tarbasename.tar";
 
 if (-e $finalTarget && !$clobber) {
     print STDERR "\nTarget exists. Use clobber to overwrite!\n\n";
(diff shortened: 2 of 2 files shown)


dicom-archive/DICOM/DCMSUM.pm
```diff
@@ -133,8 +133,7 @@ QUERY
 
     # INSERT or UPDATE 
     # get acquisition metadata
-    my $sfile = "$self->{tmpdir}/$meta.meta";
-    my $metacontent = &read_file($sfile);
+    my $metacontent = &read_file($metafile);
```
nicolasbrossard: Before your PR, the file that was read was `$self->{tmpdir}/$meta.meta` and now it is `$metafile`. The code is written as if it was right to assume that `$metafile` is already in `$self->{tmpdir}` and has extension `.meta`. Are you sure this is right? It probably is and I am most likely missing something since I am not super familiar with the code...

dicom-archive/dicomTar.pl
```diff
@@ -137,18 +137,19 @@
 my $creator         = $summary->{user};
 my $sumTypeVersion  = $summary->{sumTypeVersion}; 
 
-my $byDate;
 # Determine how to name the archive... by acquisition date or by today's date.
-if ($todayDate) { $byDate = $today; } else { $byDate = $summary->{header}->{scandate}; } # wrap up the archive 
-my $finalTarget = "$targetlocation/DCM_${byDate}_$summary->{metaname}.tar";
+my $byDate      = $todayDate ? $today : $summary->{header}->{scandate};
+my $pname       = $summary->{header}->{pname};
```
nicolasbrossard: Do you know if patient names can contain special characters, like `!`, `%`, etc...Might be worth doing

`my $pname = quotemeta($summary->{header}->{pname};`

in order to 100% sure that the file name generated with this is a valid Unix file name. But then this might be a bit paranoid...Up to you.

dicom-archive/DICOM/DCMSUM.pm
```diff
@@ -75,11 +75,11 @@ sub new {
...
-    my $self   = shift;
-    my $dbh    = shift;
-    my $meta   = shift;
-    my $update = shift;
-    
+    my $self     = shift;
+    my $dbh      = shift;
+    my $metafile = shift;
+    my $update   = shift;
+
```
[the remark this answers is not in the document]
cmadjar: agree with you! I just did not change the style that was not there but prefer using only one line too so changing it.

dicom-archive/DICOM/DCMSUM.pm
cmadjar: Yes, I changed the value $meta to be $metafile that is already in the tmp directory.

cmadjar: I also changed the value passed to the function in dicomTar.pl so that it is indeed the metafile in tmpdir. Hope my answer makes sense.

dicom-archive/dicomTar.pl
cmadjar: Good catch! It is unlikely but we never know. Adding the quotemeta function

dicom-archive/DICOM/DCMSUM.pm
nicolasbrossard: I looked at `dicomTar.pl` and `$metafile` is set to `$targetLocation/$tarbasename.meta`. But `$targetLocation` seems to be a command line argument, so there is no guarantee that it will be set to tmpdir. Am I missing something (most likely yes...)?

dicom-archive/dicomTar.pl
```diff
@@ -137,18 +137,19 @@
...
+my $finalTarget = "$targetlocation/DCM_${byDate}_$tarbasename.tar";
 
 if (-e $finalTarget && !$clobber) {
     print STDERR "\nTarget exists. Use clobber to overwrite!\n\n";
     exit $NeuroDB::ExitCodes::TARGET_EXISTS_NO_CLOBBER;
 }
 
 # read acquisition metadata into variable 
-my $metafile = "$targetlocation/$metaname.meta";
+my $metafile = "$targetlocation/$tarbasename.meta";
```
cmadjar: @nicolasbrossard I investigated a little more about the meta file and the tmpdir and here is what I found:
- in dicomTar.pl, the metafile was always created in $targetlocation (see line above)
- line 130 of dicomTar.pl: `my $summary = DICOM::DCMSUM->new($dcm_source,$targetlocation);`
- new constructor of DCMSUM is: 
```
# The constructor 
sub new {
    my $proto = shift;
    my $class = ref($proto) || $proto;
    my $self  = {};
    bless ($self, $class);

    my $dcm_dir = shift;
    my $tmp_dir = shift;
```

=> this means that actually, $self->{tmpdir} = $targetlocation...

So, I could either leave things as I did or change the call to the function DCMSUM->database and replace $metafile by $tarbasename and revert the code in DCMSUM. Not sure what makes more sense.

Does that clarify the enigma? Let me know what you think.

cmadjar:
Note to myself, it still needs work. When running the tarchiveLoader, the tarchive extraction do not contain the pname in the folder name. Need to fix this before this can be merged or re-reviewed.

PR #300, opened 2018-04-25
Fixed the perldoc for MRIProcessingUtility following PR #275

cmadjar:
Also noticed that a variable was not used anymore in a function so I removed it from the function and the calls to the function.

### DIFF
--- a/docs/scripts_md/MRIProcessingUtility.md
+++ b/docs/scripts_md/MRIProcessingUtility.md
@@ -79,30 +79,40 @@ INPUTS: seriesUID, list of tar files
 
 RETURNS: list of DICOM files corresponding to the seriesUID
 
-### extract\_tarchive($tarchive, $tarchive\_srcloc, $seriesuid)
+### extract\_tarchive($tarchive, $upload\_id, $seriesuid)
 
 Extracts the DICOM archive so that data can actually be uploaded.
 
-INPUTS: path to the DICOM archive, source location from the `tarchive` table,
-(optionally seriesUID)
+INPUTS:
+  - $tarchive : path to the archive
+  - $upload\_id: upload ID of the study
+  - $seriesuid: optionally a series UID
 
 RETURNS: the extracted DICOM directory
 
-### extractAndParseTarchive($tarchive, $tarchive\_srcloc, $seriesuid)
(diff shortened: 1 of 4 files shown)


docs/scripts_md/MRIProcessingUtility.md
```diff
@@ -261,13 +277,17 @@ Register programs.
...
 
-Moves and updates the tarchive table with the new location of the tarchive.
+Moves and updates the `tarchive` table with the new location of the
+`tarchive`.
 
-INPUTS: tarchive location, tarchive information hashref
+INPUTS:
+  - $tarchive\_location: `tarchive location`,
+  - $tarchiveInfo     : `tarchive` information hashref,
+  - $upload\_id        : upload ID of the study
```
MounaSafiHarab: not important but I know you care about these,  so just wanted to let you know, but ignore if you wish.
not sure why the `:` are not aligned (same for some functions below this as well).

cmadjar: weird. It looks like that only in the md code. However, the MD file in the browser looks OK (it always ignores spaces anyway). I checked the .pm and all looked aligned so should be OK. Thanks for checking, my inner OCD thanks you ;).

MounaSafiHarab:
@cmadjar 
reviewed it and all is good (I even counted that in PR #275, 16 functions had to be changed and I did find 16 here changed as well in MRIProcessingUtility.pm)

@nicolasbrossard 
since the changes are a bit more elaborate than the the other 2 I merged today, do you think you can look at this one, perhaps pass manual test it? and if all good, merge it?

nicolasbrossard:
I noticed that when documenting methods that take a lot (>5) of parameters, the first few are named and then '...' is added at the end of the parameter list. I think we should always list the complete set of parameters but that's just my preference. I leave it to you to decide whether you want to change that or not.

MounaSafiHarab:
> I noticed that when documenting methods that take a lot (>5) of parameters, the first few are named and then '...' is added at the end of the parameter list. I think we should always list the complete set of parameters but that's just my preference. I leave it to you to decide whether you want to change that or not.

indeed, and this is why counting the 16 functions took me forever, as 3 of those had the `(...`) in them and I could not find them!
But I think I asked @cmadjar that question in the past and I believe she was saying it is automatically done this way, non?

PR #304, opened 2018-05-09
Functions to interact with DeepQC flask app 

llevitis:
This PR adds the function `computeDeepQC` to the MRI Processing Utility. It makes a call to the DeepQC Flask app (to be merged and hosted on the same server). At the moment, I have hard-coded '44' as the `AcquisitionProtocolID` used to select those images that will be sent to the app for analysis. `tools/BackPopulateDeepQC` and the call from `tarchiveLoader` are modelled after @MounaSafiHarab's computeSNR function.


uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1652,6 +1653,81 @@ sub computeSNR {
...
+    my ($tarchiveID, $tarchive_srcloc, $profile)= @_;
+    my $data_dir = NeuroDB::DBI::getConfigSetting(
+                        $this->{dbhr},'dataDirBasepath'
+                        );
+    my $upload_id = getUploadIDUsingTarchiveSrcLoc($tarchive_srcloc);
+    my $ua = LWP::UserAgent->new;
+    my $json = 'JSON::PP'->new;
+    # note: url must be changed to production version later
+    # note 2: include DeepQC modality in prod file
+    my $acqIDForT1Scan = 44;
```
MounaSafiHarab: this `44` should be generic :)

llevitis: yeah I know...the thing is that Andrew's model only works for IBIS data at the moment anyway. But I can update it to be generic.

tools/BackPopulateDeepQC.pl
```diff
@@ -0,0 +1,128 @@
...
+
+# Query to grep all tarchive entries
+if (!defined($TarchiveID)) {
+    $query = "SELECT TarchiveID " .
+        "FROM tarchive";
+}
+# Selecting tarchiveID is redundant here but it makes the while() loop
+# applicable to both cases; when a TarchiveID is specified or not
+else {
+    $query = "SELECT TarchiveID " .
```
cmadjar: in the two queries, shouldn't you check that no deep QC were already done for the tarchives?

uploadNeuroDB/tarchiveLoader
```diff
@@ -580,6 +580,12 @@ foreach my $minc (@minc_files) {
 ############### Compute SNR on 3D modalities ###################
 ################################################################
 $utility->computeSNR($tarchiveInfo{TarchiveID}, $upload_id, $profile);
+################################################################
+############### Compute DeepQC on T1W MNC files ################
+################################################################
+$utility->computeDeepQC($tarchiveInfo{TarchiveID},
+                        $profile);
```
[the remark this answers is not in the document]
cmadjar: should not be $tarchiveInfo{TarchiveID} that is passed to computeDeepQC but $upload_id directly

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1652,6 +1653,81 @@ sub computeSNR {
...
+    my $query = "SELECT FileID, file, AcquisitionProtocolID from files f ";
+    my $where = "WHERE f.TarchiveSource=?";
```
cmadjar: why not simply write the complete query directly here instead of creating it and concatenating?

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1652,6 +1653,81 @@ sub computeSNR {
...
+
+    while ($row = $minc_file_arr->fetchrow_hashref()) {
+        $filename = $row->{'file'};
+        my $fileID = $row->{'FileID'};
+        my $acqID = $row->{'AcquisitionProtocolID'};
+        $base = basename($filename);
+        $fullpath = $data_dir . "/" . $filename;
+        if (-e $fullpath) {
+            if ($acqID == $acqIDForT1Scan) {
+                my $url = 'http://127.0.0.1:5000/deepqc/' . $fileID;
```
cmadjar: is this url specific for IBIS? If so, might want to find a way to make this more generic too. Is the code from Andrew generic enough to work on any T1W?

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
MounaSafiHarab: yes, but because it is making it to main Loris-MRi repo, it should be generic.
I believe you cna get this feidl from the ConfigSettings/Imaging Pipeline section. I believe the DTI PreProcessing pipeline needs the T1W modality as an input, so you can get it from there.
Am I not correct @cmadjar?

cmadjar: @MounaSafiHarab @llevitis indeed, you could grep it from the config setting called t1_scan_type :)

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
llevitis: @cmadjar To my understanding, he's only had success validating his model on IBIS data.

driusan: "127.0.0.1" is specific to the server that it's running on, not specific to ibis.

tools/BackPopulateDeepQC.pl
llevitis: @cmadjar - This check is done within the `computeDeepQC` in `MRIProcessingUtility.pm` function. If there is already an entry in the `parameter_file` table and the new DeepQC result is different, that entry gets updated. @crocodoyle  and I discussed a new database structure for the Flask app a few days ago, so I'll make other relevant changes this week as well.

PR #302, opened 2018-04-25
Used proper exit code constants in updateMRI_Upload.pl instead of numeric literals

nicolasbrossard:
As the title says.

PR #305, opened 2018-05-16
Used proper exit code constants in updateMRI_Upload.pl instead of numeric literals.

nicolasbrossard:
As the tile says.


MounaSafiHarab:
same as #https://github.com/aces/Loris-MRI/pull/302

MounaSafiHarab:
looks good to me too; only changing exit codes, and the new ones you used make sense based on the actual error, so merging.

PR #306, opened 2018-05-30
DeepQC interoperability 

llevitis:
This PR replaces #304 so as to be based off of the minor branch. I've addressed almost all the changes that were requested by @cmadjar and @MounaSafiHarab in the original PR. 

`computeDeepQC`:  
- [x] computeDeepQC takes as input the `profile`, `upload_id`, and `tarchive_id`. 
- [x] `computeDeepQC` determines the project specific `AcquisitionProtocolID` for T1 scans based on the `t1_scan_type` config setting. 
- [x] For any T1 MNC file associated with the `tarchive_id`, a POST request is made to the DeepQC app to compute the probability of passing QC. The response of this request is a JSON object with the prediction and confidence of the the most up to date algorithm. 
- [x] The JSON object is parsed, and the prediction value is used to update the `DeepQC` parameter in the `parameter_files` table. 
- [ ] Should the confidence also be added to the parameter_files table? or can it be added along with the prediction as a JSON object? 
- [ ] LWP::Agent & JSON::PP must be installed for existing projects. 

`BackPopulateDeepQC`: 
- [x] can either be run with a single `tarchive_id` passed or without `tarchive_id` as an argument in order to run DeepQC on every single T1 MNC the project has. 
- [ ] add Perldoc documentation

### DIFF
--- a/tools/BackPopulateDeepQC.pl
+++ b/tools/BackPopulateDeepQC.pl
@@ -126,9 +126,13 @@ if($sth->rows > 0) {
 	# Create tarchive list hash with old and new location
     while ( my $rowhr = $sth->fetchrow_hashref()) {
         my $TarchiveID = $rowhr->{'TarchiveID'};
+        my $SourceLocation = $rowhr->{'SourceLocation'};
 		print "Currently updating the DeepQC for applicable files in parameter_file table ".
             "for tarchiveID $TarchiveID\n";
-        $utility->computeDeepQC($TarchiveID, $profile);
+        my $upload_id = NeuroDB::MRIProcessingUtility::getUploadIDUsingTarchiveSrcLoc(
+                        $SourceLocation
+                    );
+        $utility->computeDeepQC($TarchiveID, $upload_id, $profile);
 		print "Finished updating DeepQC for for TarchiveID $TarchiveID\n";
 	}
 }
--- a/tools/BackPopulateDeepQC.pl
+++ b/tools/BackPopulateDeepQC.pl
@@ -97,7 +97,7 @@ my $logfile  = "$LogDir/$templog.log";
 ################## Instantiate MRIProcessingUtility ############
 ################################################################
 my $utility = NeuroDB::MRIProcessingUtility->new(
-                  $db, \$dbh,$debug,$TmpDir,$logfile,
+                  \$dbh,$debug,$TmpDir,$logfile,
                   $LogDir,$verbose
               );
 
--- a/tools/BackPopulateDeepQC.pl
+++ b/tools/BackPopulateDeepQC.pl
@@ -9,6 +9,7 @@ use File::Find;
 use Cwd;
 use NeuroDB::DBI;
 use NeuroDB::MRIProcessingUtility;
+use NeuroDB::ExitCodes;
 
 my $verbose = 1;
 my $debug = 1;
@@ -116,11 +117,11 @@ if (!defined($TarchiveID)) {
 else {
     $query = "SELECT TarchiveID " .
         "FROM tarchive ".
-        "WHERE TarchiveID = $TarchiveID ";
+        "WHERE TarchiveID = ?";
 }
 
 my $sth = $dbh->prepare($query);
-$sth->execute();
+$sth->execute($TarchiveID);
 
 if($sth->rows > 0) {
 	# Create tarchive list hash with old and new location
@@ -133,11 +134,11 @@ if($sth->rows > 0) {
                         $SourceLocation
                     );
         $utility->computeDeepQC($TarchiveID, $upload_id, $profile);
-		print "Finished updating DeepQC for for TarchiveID $TarchiveID\n";
+		print "Finished updating DeepQC for TarchiveID $TarchiveID\n";
 	}
 }
 else {
-	print "No tarchives to be updated \n";
+	print "No tarchive to be updated \n";
 }
 
 $dbh->disconnect();
--- a/tools/BackPopulateDeepQC.pl
+++ b/tools/BackPopulateDeepQC.pl
@@ -1,0 +1,140 @@
+#! /usr/bin/perl
+
+use strict;
+use warnings;
+use Getopt::Tabular;
+use File::Temp qw/ tempdir /;
+use File::Basename;
+use File::Find;
+use Cwd;
+use NeuroDB::DBI;
+use NeuroDB::MRIProcessingUtility;
+
+my $verbose = 1;
+my $debug = 1;
+my $profile = undef;
+my $TarchiveID = undef;
+my $query;
+
+my @opt_table = (
+    [ "-profile", "string", 1, \$profile,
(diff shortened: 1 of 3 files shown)


tools/BackPopulateDeepQC.pl
```diff
@@ -0,0 +1,144 @@
...
+
+use strict;
+use warnings;
+use Getopt::Tabular;
+use File::Temp qw/ tempdir /;
+use File::Basename;
+use File::Find;
+use Cwd;
+use NeuroDB::DBI;
+use NeuroDB::MRIProcessingUtility;
```
cmadjar: missing `use NeuroDB::ExitCodes;` in the imports

tools/BackPopulateDeepQC.pl
```diff
@@ -0,0 +1,144 @@
...
+if (!defined($TarchiveID)) {
+    $query = "SELECT TarchiveID " .
+        "FROM tarchive";
+}
+# Selecting tarchiveID is redundant here but it makes the while() loop
+# applicable to both cases; when a TarchiveID is specified or not
+else {
+    $query = "SELECT TarchiveID " .
+        "FROM tarchive ".
+        "WHERE TarchiveID = $TarchiveID ";
```
cmadjar: for the queries, should use the new Moose system that Nicolas developed.

General comment: variables should not be integrated in the query, instead, use placeholders in the query and execute the query with the variables. For this example, $TarchiveID should be replaced by a placeholder in the query (`?` in Perl) and the query should be executed with the `$TarchiveID` variable: `$sth-> execute($TarchiveID)`

tools/BackPopulateDeepQC.pl
```diff
@@ -0,0 +1,144 @@
...
+            "for tarchiveID $TarchiveID\n";
+        my $upload_id = NeuroDB::MRIProcessingUtility::getUploadIDUsingTarchiveSrcLoc(
+                        $SourceLocation
+                    );
+        $utility->computeDeepQC($TarchiveID, $upload_id, $profile);
+		print "Finished updating DeepQC for for TarchiveID $TarchiveID\n";
+	}
+}
+else {
+	print "No tarchives to be updated \n";
```
cmadjar: remove the 's' from 'tarchives'

tools/BackPopulateDeepQC.pl
```diff
@@ -0,0 +1,144 @@
...
+    while ( my $rowhr = $sth->fetchrow_hashref()) {
+        my $TarchiveID = $rowhr->{'TarchiveID'};
+        my $SourceLocation = $rowhr->{'SourceLocation'};
+		print "Currently updating the DeepQC for applicable files in parameter_file table ".
+            "for tarchiveID $TarchiveID\n";
+        my $upload_id = NeuroDB::MRIProcessingUtility::getUploadIDUsingTarchiveSrcLoc(
+                        $SourceLocation
+                    );
+        $utility->computeDeepQC($TarchiveID, $upload_id, $profile);
+		print "Finished updating DeepQC for for TarchiveID $TarchiveID\n";
```
cmadjar: remove duplicated 'for' in the sentence to be printed

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1641,6 +1643,90 @@ sub computeSNR {
...
+
+    while ($row = $minc_file_arr->fetchrow_hashref()) {
+        $filename = $row->{'file'};
+        my $fileID = $row->{'FileID'};
+        my $acqID = $row->{'AcquisitionProtocolID'};
+        $base = basename($filename);
+        $fullpath = $data_dir . "/" . $filename;
+        if (-e $fullpath) {
+            if ($acqID == $t1_scan_result) {
+                my $url = 'http://127.0.0.1:5000/deepqc/' . $fileID;
```
cmadjar: shouldn't the http address be in the Config? As I understand it, this is the url for IBIS but other projects might have different url for the deep QC algo since the algo would be different for each project (at least that is what I understood from our discussion).

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1641,6 +1643,90 @@ sub computeSNR {
...
+            }
+            else {
+                $message = "The DeepQC probability can not be computed for $base. ".
+                    "The imaging modality is not ".
+                    "supported by the DeepQC computation. \n";
+                $this->{LOG}->print($message);
+                $this->spool($message, 'N', $upload_id, $notify_detailed);
+            }
+        } else {
+            print "File doesn't exist.\n"
```
cmadjar: Maybe elaborate this message: 
'File $filefullpath not found in the filesystem' or something like that?

uploadNeuroDB/tarchiveLoader
```diff
@@ -570,6 +570,10 @@ foreach my $minc (@minc_files) {
 ################################################################
 $utility->computeSNR($tarchiveInfo{TarchiveID}, $upload_id, $profile);
 ################################################################
+############### Compute DeepQC on T1W images ###################
+################################################################
+$utility->computeDeepQC($tarchiveInfo{TarchiveID}, $upload_id, $profile);
```
cmadjar: technically, the call to this function should be optional based on whether DeepQC is allowed for the project. Maybe you could add a configuration in the imaging pipeline section "Compute Deep QC" with "yes/no" options. Then in the code, if the config is set to "yes" then call that computeDeepQC function.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
llevitis: @cmadjar - http://127.0.0.1:5000/ would just be the localhost address since the app would be hosted on the same server where all the MNC files are. I can add this more explicitly to the config file though so that it can be customizable.

uploadNeuroDB/tarchiveLoader
llevitis: @cmadjar - good point!! will add that :sparkles:

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
crocodoyle: I agree that the URL should be in configuration somewhere

uploadNeuroDB/tarchiveLoader
crocodoyle: This should be split up into two calls, a POST to send the image to DeepQC, and a GET (not necessarily here) to get the results. If DeepQC goes down/never returns, it should not block LORIS forever.

MounaSafiHarab:
the Loris one got merged to 19.2 which means it made it to 20.0 but the MRI side is not ready yet...

cmadjar:
No one is maintaining this branch, needs rebase and code changes to be applied. Since it is very IBIS specific, will close this PR for now.

PR #307, opened 2018-05-31
Fixed a bug in database_files_update.pl. 

cmadjar:
Gets rid of the final / if it exists in the data_dir path given in the Config module.

Redmine ticket: https://redmine.cbrain.mcgill.ca/issues/14234

### DIFF
--- a/tools/database_files_update.pl
+++ b/tools/database_files_update.pl
@@ -67,9 +67,8 @@ if  (!$profile) {
 my $dbh     =   &NeuroDB::DBI::connect_to_db(@Settings::db);
 
 # these settings are in the database and can be set in the Configuration module of LORIS
-my $data_dir = &NeuroDB::DBI::getConfigSetting(
-                    \$dbh,'dataDirBasepath'
-                    );
+my $data_dir = &NeuroDB::DBI::getConfigSetting(\$dbh,'dataDirBasepath');
+$data_dir =~ s/\/$//;
 
 # Needed for log file
 my  $log_dir    =   "$data_dir/logs";


cmadjar:
> I did not test but I don't see any reason why this wouldn't work.

I tested just in case and it indeed worked :)

MounaSafiHarab:
@cmadjar 
19.2? 
the only comment I have here is what happens to other scripts needing this variable? do they all add the `/` ?

cmadjar:
> 19.2?
@MounaSafiHarab As you wish. Technically, it is a bug fix so it could be nice to release it soon but at the same time, it is a tool script that probably nobody uses... I'm fine either way.

> the only comment I have here is what happens to other scripts needing this variable? do they all add the / ?
actually, I realized this bug while testing the imaging pipeline with data_dir ending with a / and the only scripts that failed were:
- database_files_update.pl that I tested for the documentation I think?
- DTIPrep_pipeline and DTIPrepRegister that I tend to be the only one to test ;) See PR #308

MounaSafiHarab:
> database_files_update.pl that I tested for the documentation I think?
> DTIPrep_pipeline and DTIPrepRegister that I tend to be the only one to test ;) See PR #308

never mind my comment. I just looked at your changes and you are removing it in the specific scripts themselves. I thought it was being changed in the Config which is why I was worried about other scripts using this value when I know they worked (and I do have a traling `/` in this path).

MounaSafiHarab:
works perfectly well!
I manually added the $data_dir to one of the files in my files table and ran the script before your changes (nothing changed at all) ; and after your changes, and indeed the filepath strips out the $data_dir now.

PR #308, opened 2018-05-31
Removing trailing / of data_dir in DTIPrep_pipeline and DTIPrepRegister scripts

cmadjar:
DTIPrepRegister.pl and DTIPrep_pipeline.pl crashes when there is a trailing / in the data_dir variable set in the Config module. This PR removes the trailing / when it exists from the data_dir variable.

Redmine ticket: https://redmine.cbrain.mcgill.ca/issues/14351


MounaSafiHarab:
I did not test this one, but I went through the changes, you seem to be doing exactly like PR #307 (which I tested), plus removing the multi-lines which bother you! so all good; merging.

p.s: still curious why this extra `/` fails the script :) We have it in `minc_insertion.pl` for example, and it is all good.

PR #311, opened 2018-06-01
Select statements in the new database framework now return a reference to an array of hash references.

nicolasbrossard:
When a `SELECT` statement is executed with the new database framework, a reference to an array of hash references is returned. The key/value pairs of this hash are the column names (as specified in the `SELECT` statement) and their values. This consequently modifies the behaviour of all object brokers. 

Basically, in the past you had code like:

```perl
# First array: what you want
# Second array: constraints in the where clause
$r = $objectBroker->get( ['col1', 'col4', 'col12'], { col2 => 'nbro', col7 => 5 });

 # col12 is accessed using its column index in the get call
$col12ForFourthRecord = $r->[3]->[2];   
```    

Whereas now, you would see:

```perl
# First array: what you want
# Second array: constraints in the where clause
$r = $objectBroker->get( ['col1', 'col4', 'col12'], { col2 => 'nbro', col7 => 5 });

 # col12 is accessed directly
$col12ForFourthRecord = $r->[3]->{'col12'};   
```

### DIFF
--- a/dicom-archive/updateMRI_Upload.pl
+++ b/dicom-archive/updateMRI_Upload.pl
@@ -100,8 +100,6 @@ use File::Basename;
 use lib "$FindBin::Bin";
 use DICOM::DICOM;
 
-use NeuroDB::MRI;
-
 use NeuroDB::Database;
 use NeuroDB::DatabaseException;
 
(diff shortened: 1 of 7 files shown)


dicom-archive/updateMRI_Upload.pl
```diff
@@ -235,6 +237,9 @@ =head1 AUTHORS
 );
 $db->connect();
 
+my $xxxxx = &NeuroDB::MRI::scan_type_text_to_id('scout', $db);
+my $yyyyy = &NeuroDB::MRI::scan_type_id_to_text(52, $db);
```
cmadjar: not sure I understand what those two lines are for. Maybe you added them for testing MRI.pm and forgot to remove them? If so, I guess the use NeuroDB::MRI is not necessary as well in line 103.

nicolasbrossard: Arghh!!! The perils of issuing a PR at 10PM...I'll remove this

PR #313, opened 2018-06-01
Perldoc harmonization

cmadjar:
**Takes care of harmonizing the perl documentation:**
- remove empty todo sections and empty bugs sections
- harmonize /data/project/data/ to be /data/$PROJECT/data
- harmonize authorship for the LORIS team
- try to follow this pull request (#226) for the TODO section; i.e. use this: "Fix comments written as #fixme in the code." whenever appropriate in other scripts
- logHeader() function documented consistently (check #237 versus #246)
- ensure that all tables and variables and paths to files are written in a C<> format 
- consistent INPUT vs INPUTS in Methods
- consistency in the (constructor) method 
- INPUTS to functions when many inputs exist are now documented with (...), Nicolas raised this in this PR: #300; so PERHAPS reconsider revisiting all scripts to list all inputs anyway? (done)

Note: the .md files are not created yet. Writing a script to automatically create perldoc becomes I don't feel doing it by hand (the script will be committed in a separate PR and could be run before each release so that if we forget to update a MD file in a PR, we can always mass run the creation of MD files) 

Associated redmine ticket: https://redmine.cbrain.mcgill.ca/issues/13688

PR #314, opened 2018-06-01
Creation of a tool script that creates all .md documentation from the perl scripts

cmadjar:
To run the tool:
`perl tools/mass_perldoc_md_creation.pl -profile prod`
or 
`perl tools/mass_perldoc_md_creation.pl -profile prod -verbose` if wants to run it with verbose option

This script can then be run in preparation of each release to make sure all the .md documentation has been updated. If new files have been added to a given release, then will need to include it in the `@script_list` variable set at the beginning of the script.


tools/mass_perldoc_md_creation.pl
```diff
@@ -0,0 +1,176 @@
+#!/usr/bin/perl -w
```
nicolasbrossard: I would be in favour of simplifying the script. I would get rid of `@script_list`: no need to explicitly list (and maintain) the list of scripts/modules when a Unix `find` command can do this for you. What do you think?

MounaSafiHarab: I would tend to agree with @cmadjar simply because we do not have all the files perl documented properly.
It is a bit like Travis where we add the .js files to run npm compile and phpcs on the php scripts; but those are ran on files manually added... so it gives us flexibility to exclude some if need be.

MounaSafiHarab:
one comment I have:
once PR#313 is merged, cna you please include, in this PR, the script after you ran it and attach all the `.md files`? or were you planning on doing it in a separate PR?
because this step has to be done before release, in case any new modified `.md` files are generated so they can be added to the released branch.

tools/mass_perldoc_md_creation.pl
```diff
@@ -0,0 +1,176 @@
...
+
+=pod
+
+=head1 NAME
+
+mass_perldoc_md_creation.pl -- Script to mass produce the .md files derived
```
MounaSafiHarab: perhaps C<> for `.md`?

tools/mass_perldoc_md_creation.pl
```diff
@@ -0,0 +1,176 @@
...
+Available options are:
+
+-profile: name of the config file in C<../dicom-archive/.loris_mri>
+
+-verbose: be verbose (boolean)
+
+
+=head1 DESCRIPTION
+
+This script will need to be run once per release to make sure the .md files
```
MounaSafiHarab: here too for `.md` if you opted for the C<> option above

tools/mass_perldoc_md_creation.pl
```diff
@@ -0,0 +1,176 @@
...
+derived from the documentation of the perl scripts and libraries are updated.
+
+If any new script have been added to a given release, make sure to include it
+in the variable called @script_list at the beginning of the script.
```
MounaSafiHarab: C<> for `@scripts`

tools/mass_perldoc_md_creation.pl
```diff
@@ -0,0 +1,176 @@
...
+This script will need to be run once per release to make sure the .md files
+derived from the documentation of the perl scripts and libraries are updated.
+If any new script have been added to a given release, make sure to include it
+in the variable called \@script_list at the beginning of the script.
+
+Documentation: perldoc mass_perldoc_md_creation.pl
+
+HELP
+my $Usage = <<USAGE;
+usage: tools/mass_perldoc_md_creation.pl -profile \$profile
```
MounaSafiHarab: perhaps add the word `perl` before `tools/mass...`

tools/mass_perldoc_md_creation.pl
```diff
@@ -0,0 +1,176 @@
...
+my @md_list  = map { $md_path . basename($_, @suffixes) . ".md"} @script_list;
+
+
+
+## loop through script array and create the .md files using pod2markdown
+for my $index (0 .. $#script_list) {
+    my $script  = $script_list[$index];
+    my $md_file = $md_list[$index];
+    my $command =  "pod2markdown $script $md_file";
+    print $command if $verbose;
```
MounaSafiHarab: add a line break as otherwise, your screen will look like this:

screen

MounaSafiHarab:
tested it and it works well!
nice tool.

But perhaps one useful feature, if not too complicated to add, is it possible if the script outputs a list of files that were changed so it is easy for us to simply run `git add` on those?

cmadjar:
> once PR #313 is merged, cna you please include, in this PR, the script after you ran it and attach 

> all the .md files? or were you planning on doing it in a separate PR?
because this step has to be done before release, in case any new modified .md files are generated so they can be added to the released branch.

Haha, that answers the question I just asked on PR #313. Sure I will create the .md file once PR #313 is merged and append it to this PR

tools/mass_perldoc_md_creation.pl
cmadjar: done

tools/mass_perldoc_md_creation.pl
cmadjar: done

tools/mass_perldoc_md_creation.pl
cmadjar: done

tools/mass_perldoc_md_creation.pl
cmadjar: done

tools/mass_perldoc_md_creation.pl
cmadjar: done

cmadjar:
> But perhaps one useful feature, if not too complicated to add, is it possible if the script outputs a list of files that were changed so it is easy for us to simply run git add on those?

oh yeeees, great idea!! I've added that functionality. Thank you for suggesting it!!!

@nicolasbrossard @MounaSafiHarab ready for another review. :)

MounaSafiHarab:
Same comment as the previous PR; just added the tag as it is going to minor and it is good to have it in 19.2... so this tag is for us @cmadjar to decide what is best without forgetting the branch/milestone inconsitency

tools/mass_perldoc_md_creation.pl
```diff
@@ -0,0 +1,182 @@
...
+
+## create the list of md file based on @script_list and change extension to .md
+my $md_path  = $loris_mri_path . '/docs/scripts_md/';
+my @suffixes = (".pl", ".pm");
+my @md_list  = map { $md_path . basename($_, @suffixes) . ".md"} @script_list;
+
+
+
+## loop through script array and create the .md files using pod2markdown
+my $git_add = "git add";
```
MounaSafiHarab: A note for me:
it is fair to assume that `git` is installed since this is how the `readme` tells projects to install Loris-MRI,
so I think we r all good here

MounaSafiHarab:
> oh yeeees, great idea!! I've added that functionality. Thank you for suggesting it!!!

so the `git add` command feature you added will add everything (even if the .md file did not change).
But I guess this is Okay?

tools/mass_perldoc_md_creation.pl
MounaSafiHarab: still not working, but It is because you append with `. '\n'`
tried it with `"\n"` and it works (double instead of single quotation)

cmadjar:
> so the git add command feature you added will add everything (even if the .md file did not change). But I guess this is Okay?

Actually, if you do a `git add` of a file that did not change, it won't do anything (just tried it) so it should be OK to do a massive `git add`. It will only indeed add the ones that changed.

tools/mass_perldoc_md_creation.pl
cmadjar: @MounaSafiHarab Thanks! This should be resolved now.

PR #315, opened 2018-06-14
Adding perldoc and markdown as dependencies at install time: Redmine 14724

MounaSafiHarab:


MounaSafiHarab:
@cmadjar @nicolasbrossard 
I created this PR, but it is debatable if we need this.
The Pod2Markdown is only for us/development I think to generate the updated .md files, there is no reason why a production environment or an outsider installing LORIS  would have those, non?

perldoc might be nice IN CASE they want to type perldoc at the command line...

not sure really how you both feel about adding those to the installer or keeping them out?

MounaSafiHarab:
@kongtiaowang, @um4r12 

Hi guys, do you think one of you can test this where you have an MRI install?

it is straightforward to test (I need someone without that dependency in their MRI install and most the MRI people do by now including @davidblader @cmadjar @nicolasbrossard !)
so try in the mri directory to run
`pod2markdown uploadNeuroDB/tarhiveLoader docs/scripts/tarchuveLoader.md`
and
`perldoc uploadNeuroDB/tarhiveLoader`

BEFORE you run the added dependencies in this pull request
then repeat
AFTER you run the 3 lines modified in this PR

let me know

PR #316, opened 2018-06-15
Fixing minc_insertion.pl call to validateCandidate, had too many parameters

cmadjar:
This fixes the call to the `validateCandidate` function from the `MRIProcessingUtility.pm` file to have only one parameter (as it should have been since tarchive source location is not an argument of the function. 

Tested on a scan and it works as expected.

@MounaSafiHarab was not sure if this should go to 19.2. I think we can easily change the branch if we decide to merge it for 20.0 instead.

### DIFF
--- a/uploadNeuroDB/minc_insertion.pl
+++ b/uploadNeuroDB/minc_insertion.pl
@@ -445,9 +445,7 @@ my $subjectIDsref = $utility->determineSubjectID(
 ################################################################
 
 my $CandMismatchError = undef;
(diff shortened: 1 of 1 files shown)


MounaSafiHarab:
@cmadjar 
added 19.2 milestone as I think it should go to the next release, whatever it is called
(as discussed today with @nicolasbrossard and you, we will conitnue to send 19.2 and consider this our next release, whatever LORIS decides to call it).

PR #317, opened 2018-06-19
getPSC() now checks the centerID from the session table first, if it exists, before it regex matches the patientname with the site Alias or MRI alias: Redmine 14422

MounaSafiHarab:
So the logic of getPSC() now becomes:
check first if the visit label exists in the session table for that PSCID, if a match is found, then use that CenterID, otherwise, use regex matching on the PatientName (this was the default option from before the change in this PR)

In addition, the original script was using the `getSubjectIDs` routine in MRI.pm as opposed to the one in the `prod` file. Projects can not customize those in the script without deviating from the released MRI codebase, so now I make it point to the one in the `prod` file so that 1) the same `getSubjectIDs` routine is used consistently throughout the insertion scripts and 2) if projects customized the one in their `prod` file, their customization is fully propagated (as opposed to partially) throughout when calling this routine.

### DIFF
--- a/docs/scripts_md/MRI.md
+++ b/docs/scripts_md/MRI.md
@@ -257,10 +257,10 @@ RETURNS: (int) CandID
 
 ### getPSC($patientName, $dbhr)
 
-Looks for the site alias using the session table CenterID as a first
-resource, for the cases where it is created using the front-end,
-otherwise, find the site alias in whatever field (usually patient\_name 
-or patient\_id) is provided, and return the MRI alias and CenterID.
+Looks for the site alias using the `session` table `CenterID` as 
+a first resource, for the cases where it is created using the front-end,
+otherwise, find the site alias in whatever field (usually `patient_name` 
+or `patient_id`) is provided, and return the `MRI_alias` and `CenterID`.
 
 INPUTS: patient name, database handle reference
 
--- a/docs/scripts_md/MRI.md
+++ b/docs/scripts_md/MRI.md
@@ -257,8 +257,10 @@ RETURNS: (int) CandID
 
 ### getPSC($patientName, $dbhr)
 
-Looks for the site alias in whatever field (usually patient\_name or
-patient\_id) is provided and return the MRI alias and CenterID.
+Looks for the site alias using the session table CenterID as a first
+resource, for the cases where it is created using the front-end,
+otherwise, find the site alias in whatever field (usually patient\_name 
+or patient\_id) is provided, and return the MRI alias and CenterID.
 
 INPUTS: patient name, database handle reference
 
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -1156,10 +1156,10 @@ sub createNewCandID {
 
 =head3 getPSC($patientName, $dbhr)
 
-Looks for the site alias using the session table CenterID as a first
-resource, for the cases where it is created using the front-end,
-otherwise, find the site alias in whatever field (usually patient_name 
-or patient_id) is provided, and return the MRI alias and CenterID.
+Looks for the site alias using the C<session> table C<CenterID> as 
+a first resource, for the cases where it is created using the front-end,
+otherwise, find the site alias in whatever field (usually C<patient_name> 
+or C<patient_id>) is provided, and return the C<MRI_alias> and C<CenterID>.
 
 INPUTS: patient name, database handle reference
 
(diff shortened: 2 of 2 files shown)


MounaSafiHarab:
@cmadjar @nicolasbrossard 

can you take a look? discussed on April 13th in one of the imaging meeting. It is necessary now for CCNA as otherwise, the existing insertion will not work (our candidates travel to a site other than the candidate's site for MRI visits for scanning, so I can not rely on the PSCID), hence the bugfix tag and the 19.2-dev branch for that.
So here I basically check if the visit is created first from the front end, and if so, use the CenterID from session table, otherwise, for projects like PreventAD for example, where the scripts create the visit, use the original way of checking the PatientName and regex match it to the alias in the psc table.

docs/scripts_md/MRI.md
```diff
@@ -257,8 +257,10 @@ RETURNS: (int) CandID
 
 ### getPSC($patientName, $dbhr)
 
-Looks for the site alias in whatever field (usually patient\_name or
-patient\_id) is provided and return the MRI alias and CenterID.
+Looks for the site alias using the session table CenterID as a first
```
cmadjar: session and CenterID should be in code format

docs/scripts_md/MRI.md
```diff
@@ -257,8 +257,10 @@ RETURNS: (int) CandID
...
+resource, for the cases where it is created using the front-end,
+otherwise, find the site alias in whatever field (usually patient\_name 
+or patient\_id) is provided, and return the MRI alias and CenterID.
```
cmadjar: idem, patient_name, patient_id and CenterID should be in code format

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -1169,35 +1171,42 @@ RETURNS: a two element array:
...
+                            null,
+                            $dbhr
+                        );
+    my $PSCID = $subjectIDsref->{'PSCID'};
+    my $visitLabel = $subjectIDsref->{'visitLabel'};
+
+    ## Get the CenterID from the session table, if the PSCID and visit labels exist 
+    ## and could be extracted  
+    if ($PSCID && $visitLabel) {
+    	my $query = "SELECT s.CenterID,p.MRI_alias FROM session s
```
cmadjar: missing a space between the tow fields you select (very very nitpicky...)

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -1169,35 +1171,42 @@ RETURNS: a two element array:
...
+    my $PSCID = $subjectIDsref->{'PSCID'};
+    my $visitLabel = $subjectIDsref->{'visitLabel'};
+
+    ## Get the CenterID from the session table, if the PSCID and visit labels exist 
+    ## and could be extracted  
+    if ($PSCID && $visitLabel) {
+    	my $query = "SELECT s.CenterID,p.MRI_alias FROM session s 
+                    JOIN psc p on p.CenterID=s.CenterID  
+                    JOIN candidate c on c.CandID=s.CandID  
+                    WHERE c.PSCID = '$PSCID' AND s.Visit_label = '$visitLabel'";
```
cmadjar: values for $PSCID and $visitLabel should be in the execute statement, not directly in the query.

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -1169,35 +1171,42 @@ RETURNS: a two element array:
...
+
+        my $sth = $${dbhr}->prepare($query);
```
cmadjar: fix indentation across the function

docs/scripts_md/MRI.md
MounaSafiHarab: done

docs/scripts_md/MRI.md
MounaSafiHarab: done

uploadNeuroDB/NeuroDB/MRI.pm
MounaSafiHarab: done

uploadNeuroDB/NeuroDB/MRI.pm
MounaSafiHarab: good catch, thanks

uploadNeuroDB/NeuroDB/MRI.pm
MounaSafiHarab: weird, it looked OKay in my editor... I think it is fixed now

MounaSafiHarab:
> @MounaSafiHarab Quickly looked at your PR between two meetings and noticed a few minor things that would need to be fixed. Once fixed, I will test your PR.

your comments should be addressed now.

PR #318, opened 2018-06-20
Exit if comma separated values in mri_protocol checks ValidRange column

MounaSafiHarab:
follow up to PR #283

@llevitis PR in LORIS (https://github.com/aces/Loris/pull/3523) had 2 tables, so the MRI counterpart should ensure projects no longer have comma separated values in BOTH those tables. So add the `mri_protocol_checks` here besides the check in PR#283 for `mri_protocol`.

### DIFF
--- a/docs/scripts_md/MRIProcessingUtility.md
+++ b/docs/scripts_md/MRIProcessingUtility.md
@@ -388,7 +388,8 @@ INPUTS:
 
 ### isValidMRIProtocol()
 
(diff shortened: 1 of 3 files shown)


MounaSafiHarab:
I sent this to minor because the other counter part is merged there;
so we can decide how best to do this
(either merge minor changes onto 19.2 then rebase this one, or merge this onto minor first then merge minor onto 19.2)...

PR #264, opened 2018-01-24
Move localizer exclusion setting into the Config module

cmadjar:
Until now, the exclusion of localizers were hardcoded in the tarchiveLoader in the variable $exclude.

This PR allows the possibility to configure the sequences that a user might want to exclude from the dcm2mnc command (localizers, scouts and other). In the "Imaging Pipeline" section of the Config module, it is now possible to specify the different series descriptions to be excluded from the insertion pipeline. Two default example have been set in LORIS: is `localizer` and `scout` (see screenshot below)

screen shot 2018-02-07 at 5 45 34 pm

The imaging script will then used the values stored in the Config modules for the `excluded_series_description` config setting and create a regular expression out of those values to use afterwards when calling dcm2mnc via the `find` command which will exclude the files matching the created regular expression pattern.

PR on the LORIS side with schema changes: https://github.com/aces/Loris/pull/3449

PR #312, opened 2018-06-01
More overall Documentation tweaking, and updating README to 19.2

MounaSafiHarab:
Mostly Readme version change, but also more documentation changes as I read them all at once...

PR #319, opened 2018-06-21
MRI exclusion scans such as localizers read from the Config module. 

cmadjar:
Basically redoing #264 after catastrophic rebasing. Only took 3 hours of my time today...

Until now, the exclusion of localizers were hardcoded in the tarchiveLoader in the variable $exclude.

This PR allows the possibility to configure the sequences that a user might want to exclude from the dcm2mnc command (localizers, scouts and other). In the "Imaging Pipeline" section of the Config module, it is now possible to specify the different series descriptions to be excluded from the insertion pipeline. Two default example have been set in LORIS: is `localizer` and `scout` (see screenshot below)

screen shot 2018-02-07 at 5 45 34 pm

The imaging script will then used the values stored in the Config modules for the `excluded_series_description` config setting and create a regular expression out of those values to use afterwards when calling dcm2mnc via the `find` command which will exclude the files matching the created regular expression pattern.

PR on the LORIS side with schema changes: https://github.com/aces/Loris/pull/3449

### DIFF
--- a/uploadNeuroDB/NeuroDB/DBI.pm
+++ b/uploadNeuroDB/NeuroDB/DBI.pm
@@ -89,7 +89,16 @@ sub getConfigSetting
     $query = $query . $where;
     my $sth = $$dbh->prepare($query);
     $sth->execute($name);
-    if ( $sth->rows > 0 ) {
+
+    if ( $sth->rows > 1 ){
+        # if more than one row returned, push data into an array that will be
+        #  dereferenced into $value
+        my @values;
+        while (my $row = $sth->fetchrow_array()) {
+            push (@values, $row);
+        }
+        $value = \@values;
+    } elsif ( $sth->rows > 0 ) {
         $value = $sth->fetchrow_array();
     }
     return $value;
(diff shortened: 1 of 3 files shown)


um4r12:
@cmadjar You could add to the [Install readme](https://github.com/MounaSafiHarab/Loris-MRI/blob/827cf8052acfbf5cfa9b13c1944552dc85470cfa/docs/02-Install.md) under the [configuration settings](https://github.com/MounaSafiHarab/Loris-MRI/blob/827cf8052acfbf5cfa9b13c1944552dc85470cfa/docs/02-Install.md#-22-configuration) or the [Post-installation checks](https://github.com/MounaSafiHarab/Loris-MRI/blob/827cf8052acfbf5cfa9b13c1944552dc85470cfa/docs/02-Install.md#-23-post-installation-checks)?

MounaSafiHarab:
@cmadjar 
all good, I am ready to merge this
Do you prefer to update the documentation now or in a separate PR?
(or perhaps you can do it by pushing to PR #312 directly? 
let me know so I know if I should merge this as is now

cmadjar:
@MounaSafiHarab I think the best is to send a PR when we are ready to release to update all the documentation. It is possible that some other documentation change between now and release time. Would that plan work?

MounaSafiHarab:
@cmadjar

I think I was not clear when I added the tag (which by the way is for both the perldoc, i.e. the .md files, and the documentation in the docs/ directory). Here I am referring to the documentation in the docs/ directory since it is a feature and projects should be made aware of it.
This is why I said you can update PR #312 as it can be added there.
I will tag you there with a reference to this PR where I thin you can edit the documentation to reflect this new feature.

MounaSafiHarab:
merging this as @cmadjar  will update the relevant documentation in PR #312!

PR #321, opened 2018-07-09
Modified error message when no MINC was created so that it displays which series descriptions were excluded

cmadjar:
Now that we can configure which series description to exclude from the `dcm2mnc` command, the hardcoded display of `no valid mincs could be converted. Localizers will not be considered` when no MINC file was created is misleading.

With this PR, the following error message would appear when no MINC file was created during the dcm2mnc command:
screen shot 2018-07-09 at 12 35 27 pm

Sequel to #319

PR #324, opened 2018-07-11
Fixed missing \n at the end of a print statement

cmadjar:
This is in reference to a comment made by @MounaSafiHarab on PR #321 where missing a `\n` at the end of a `print` statement in the `tarchiveLoader`


MounaSafiHarab:
all good!

with an exclude:
withsettings

and without:
without

PR #326, opened 2018-07-12
Fixing logic for excluded series from the dcm2mnc conversion

cmadjar:
This fixes a bug in the logic of the excluded series from the dcm2mnc conversion.

Make sure to test those three situations:
1) run the pipeline with 2 or more excluded series in the Config module
2) run the pipeline with only 1 excluded series in the Config module
3) run the pipeline with no excluded series in the Config module

See also: https://redmine.cbrain.mcgill.ca/issues/14916


MounaSafiHarab:
@nicolasbrossard 

can you please code review this? I will manually test it
this is very important as we can not release without this fix

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1092,17 +1092,22 @@ sub dicom_to_minc {
...
+        $excluded_regex = join('|', map { quotemeta($_) } @$exclude);
+    } elsif ($exclude) {
+        $excluded_regex = $exclude;
+    }
     $d2m_cmd = "find $study_dir -type f | $get_dicom_info -studyuid -series".
                " -echo -image -file -series_descr -attvalue 0018 0024".
-               " -stdin | sort -n -k1 -k2 -k6 -k3 -k7 -k4 | grep -iv".
-               " -E \"($excluded_regex)\" | cut -f 5 | ";
+               " -stdin | sort -n -k1 -k2 -k6 -k3 -k7 -k4 ";
+    $d2m_cmd .= " | grep -v -E \"($excluded_regex)\"" if ($excluded_regex);
```
MounaSafiHarab: @cmadjar 
shouldn;t there by an `i` before the `v`? it used to be `grep -iv` now it is only `grep -v` (not even sure what this `i` stands for; case insensitivity?)

cmadjar: weird... Don't know where the `i` went when I made the changes... Changing it back.

FYI - `i` in `grep` is for `ignore-case`

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1092,17 +1092,22 @@ sub dicom_to_minc {
...
+        $excluded_regex = join('|', map { quotemeta($_) } @$exclude);
+    } elsif ($exclude) {
+        $excluded_regex = $exclude;
+    }
     $d2m_cmd = "find $study_dir -type f | $get_dicom_info -studyuid -series".
                " -echo -image -file -series_descr -attvalue 0018 0024".
-               " -stdin | sort -n -k1 -k2 -k6 -k3 -k7 -k4 | grep -iv".
-               " -E \"($excluded_regex)\" | cut -f 5 | ";
+               " -stdin | sort -n -k1 -k2 -k6 -k3 -k7 -k4 ";
+    $d2m_cmd .= " | grep -iv -E \"($excluded_regex)\"" if ($excluded_regex);
```
nicolasbrossard: Just thought of this: the `grep` command will check for an occurence of any of the excluded series _anywhere_ on the string, not just in the series_descr value returned by the call to $get_dicom_info. Could this be a problem?

Also, if your study has two series described as, say 't1' and 't1_weighted', and you specify that you want to exclude serie 't1', then serie 't1_weighted' will also be excluded (as 't1_weighted` matches regex 't1')

MounaSafiHarab:
cool, so I will test tomorrow then now that the code is finalized and approved by @nicolasbrossard

MounaSafiHarab:
~~> cool, so I will test tomorrow then now that the code is finalized and approved by @nicolasbrossard~~

will wait until the last comment is resolved

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
cmadjar: At first, I thought it would not be a bad idea so that anything that has Scout or Localizer in it can be removed from `dcm2mnc` (this was also the behavior before BTW).

However, I see your point about the t1 excluding other series descriptions. Not sure how to fix this though

cmadjar: @nicolasbrossard I updated the PR so that the regex can have a regular expression that can look like something like that: `"AAHScout\s+$|AAHScout_MPR\s+$`. 
Unfortunately, I cannot use the `^` at the beginning of the each value of the regex as the regex is run on the output of `get_dicom_info.pl` which looks like this:
```
1.3.12.2.1107.5.2.32.35442.30000017021714191057800000010	18	1 	1 	./MR.1.3.12.2.1107.5.2.32.35442.2017021715254857655874003	*fl3d1_ns 	AAHScout_MPR	
1.3.12.2.1107.5.2.32.35442.30000017021714191057800000010	18	1 	2 	./MR.1.3.12.2.1107.5.2.32.35442.2017021715254858304274005	*fl3d1_ns 	AAHScout_MPR	
1.3.12.2.1107.5.2.32.35442.30000017021714191057800000010	18	1 	3 	./MR.1.3.12.2.1107.5.2.32.35442.2017021715254858817874007	*fl3d1_ns 	AAHScout_MPR	
1.3.12.2.1107.5.2.32.35442.30000017021714191057800000010	18	1 	4 	./MR.1.3.12.2.1107.5.2.32.35442.2017021715254859335574009	*fl3d1_ns 	AAHScout_MPR
```
See commit https://github.com/aces/Loris-MRI/pull/326/commits/423826946ba06b1c20604c98dea0f525fc0e1561

It was the best I could do. At least, the `$` at the end should help restricting more the regex. Any thoughts?

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1092,18 +1092,23 @@ sub dicom_to_minc {
...
-    my ($d2m_cmd,$d2m_log,$exit_code);
+    my ($d2m_cmd, $d2m_log, $exit_code, $excluded_regex);
     my $message = '';
 
     # create the excluded series description regex necessary to exclude the
     # series description specified in the Config Setting
     # excluded_series_description
-    my $excluded_regex = join('|', map { quotemeta($_) } @$exclude);
+    if ($exclude && ref($exclude) eq 'ARRAY') {
+        $excluded_regex = join('\s+$|', map { quotemeta($_) }@$exclude) . '\s+$';
```
nicolasbrossard: I think what you really want is

```perl
$excluded_regex = join('|', map { "\t\Q$_\E$" } @$exclude)
```

basically I am taking all the elements of `@$exclude` and doing this:

1. Apply `quotemeta` on the element (note that `"\Q$var\E" = quotemeta($var)`)
2. Add a tab at the beginning
3. Add a `$` at the end

Once I've done this to all the elements in `@$exclude`, I build the regex by joining the elements with `|`.

This should check if there's a match in the *last* field in the (tab-separated) list reported by `get_dicom_info.pl`.

Makes sense?

cmadjar: Thanks for the suggestion! I applied it in my last commit. The only thing is that I could not include the $ `{ "\t\Q$_\E$" }` as it got me an error and if I escape it, it is not what I want either. Not the prettiest but at least it works!

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1092,18 +1092,23 @@ sub dicom_to_minc {
...
-    my ($d2m_cmd,$d2m_log,$exit_code);
+    my ($d2m_cmd, $d2m_log, $exit_code, $excluded_regex);
     my $message = '';
 
     # create the excluded series description regex necessary to exclude the
     # series description specified in the Config Setting
     # excluded_series_description
-    my $excluded_regex = join('|', map { quotemeta($_) } @$exclude);
+    if ($exclude && ref($exclude) eq 'ARRAY') {
+        $excluded_regex = join('\s+$|', map { "\t\Q$_\E" } @$exclude) . '\s+$';
```
nicolasbrossard: Try this instead

```perl
$excluded_regex = join('|', map { quotemeta($_) } @$exclude);
.
.
.
$d2m_cmd .= " | grep -iv -E '\t($excluded_regex)$'" if $excluded_regex;
```

cmadjar:
@nicolasbrossard I finally got it! I had to use the `-P` option of `grep` so that the `\t` could be understood in the regex. I also add `\s*` before the `$` at the end of the regex as it is possible there are extra space after the series description.

MounaSafiHarab:
@cmadjar @nicolasbrossard 
I tested this under all 3 cases (none excluded, 1 excluded, and 2 excluded), all works as expected from what I can see.
Also it seems to indeed exclude now based on an exact match
(I have a series of `t2_star` in the DICOM and I excluded based on `t2` and it did not exclude it, as required)

so all good!

PR #329, opened 2018-07-24
Fixed a statement in tarchiveLoader that would cause a PERL warning in specific circumstances.

nicolasbrossard:
A warning would be printed by `tarchiveLoader` if it tries to update a record in table `mri_upload` for which `number_of_mincInserted` is `NULL`. Variable `$oldCount` is undefined in this case. This PR sets `$oldCount` to `0` in this situation so it is always defined.

### DIFF
--- a/uploadNeuroDB/tarchiveLoader
+++ b/uploadNeuroDB/tarchiveLoader
@@ -604,8 +604,8 @@ if ($valid_study) {
 
     my $sth = $dbh->prepare($query);
     $sth->execute($tarchiveInfo{TarchiveID});
-    my $oldCount = $sth->rows == 0 ? 0 : $sth->fetchrow_array;
-    my $newCount = $minc_inserted + $oldCount;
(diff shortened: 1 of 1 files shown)


uploadNeuroDB/tarchiveLoader
```diff
@@ -604,7 +604,7 @@ if ($valid_study) {
 
     my $sth = $dbh->prepare($query);
     $sth->execute($tarchiveInfo{TarchiveID});
-    my $oldCount = $sth->fetchrow_array;
+    my $oldCount = $sth->rows == 0 ? 0 : $sth->fetchrow_array;
```
cmadjar: for me `$sth->rows` = 1 (it still returns one row whether the value is NULL or a number).

Should probably be something like that instead:
`my $oldCount = $sth->fetchrow_array ? $sth->fetchrow_array : 0`

nicolasbrossard:
Corrected. Ready for more tests! Thanks!

MounaSafiHarab:
all good, no more warning for $oldCount not initialized.
merging

PR #334, opened 2018-08-02
Removing creation of site incoming directory from the install scripts

cmadjar:
Removal of the creation of `sites` directory in the `incoming` folder since they are not used anymore by the pipeline.

See also: Redmine https://redmine.cbrain.mcgill.ca/issues/14967

### DIFF
--- a/imaging_install.sh
+++ b/imaging_install.sh
@@ -92,14 +92,9 @@ echo "Creating the data directories"
 echo
 
 #####################################################################################
-###############incoming directory using sites########################################
+###############incoming directory ###################################################
 #####################################################################################
 sudo -S su $USER -c "mkdir -m 2770 -p /data/incoming/"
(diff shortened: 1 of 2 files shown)


cmadjar:
@um4r12 Maybe you could review this since you identified that issue :). Thanks!!

nicolasbrossard:
@cmadjar  For old projects that did use the "site" subdirectory, were there any plans to provide a script that would move the files that were written there to another location?

cmadjar:
> @cmadjar For old projects that did use the "site" subdirectory, were there any plans to provide a script that would move the files that were written there to another location?

@nicolasbrossard technically, those "site" directory are "incoming" data and should be emptied out as soon as the scans are uploaded (I believe we only keep the data that is in the `tarchive` directory to avoid duplication).

PR #338, opened 2018-09-12
Adding regex check for phantom validation

cmadjar:
Currently, in the `IsCandidateInfoValid()` function of the `ImagingUpload.pm` library, we only checked the patient name in the DICOM file for `PSCID_CandID_VisitLabel`. For the phantom, no checks were performed. Since the convention of phantom naming always contains the name "phantom" in all LORIS instances, this PR adds a check for phantom scans to make sure the patient name in the DICOM file contains the string "phantom".

## How to test this PR

- try inserting a candidate scan with the code from that PR
- try inserting a phantom scan with the code from that PR

### DIFF
--- a/uploadNeuroDB/NeuroDB/ImagingUpload.pm
+++ b/uploadNeuroDB/NeuroDB/ImagingUpload.pm
@@ -266,8 +266,12 @@ sub IsCandidateInfoValid {
             ############## and the file is of type DICOM###########
             #######################################################
                     if ($row[4] eq 'N') {
-                        if ( !$this->PatientNameMatch($_) ) {
-                                $files_with_unmatched_patient_name++;
+                        if ( !$this->PatientNameMatch($_, $this->{'pname'}) ) {
+                            $files_with_unmatched_patient_name++;
+                        }
+                    } elsif ($row[4] eq 'Y') {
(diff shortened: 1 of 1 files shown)


cmadjar:
@davidblader I talked with Zia about that issue and this fixes partially an issue setting phantom data to autolaunch (do some validation on the patient name to make sure it contains the string "phantom"). Since you guys need that for CCNA, can you do thorough testing of this and make sure it works? (I did not have a chance to fully test it unfortunately, running after time)
Thanks!

cmadjar:
@davidblader Very good point! Can't believe I forgot that ;). Should be fixed in the next commit. So:
- for candidates: check that the patient name starts with the patientname present in mri_upload
- for phantoms: check that the patient name contains the string "phantom"

cmadjar:
@davidblader Ready for review :)

davidblader:
woops! sorry, i put PassedManualTest too quickly on this one. the code itself is good and it works, the only issue i came across is that it seems to conflict with the current default Config value for phantoms

running the insertion scripts for an example phantom named `lego_phantom_L2_CAL_20180905`
`The PatientID read from the DICOM header does not start with (/phantom/i)|(/phantom/i) from the mri_upload table`

i'll issue a followup PR to address this

PR #351, opened 2018-11-27
Cleanup multiple ways of getting header info from MINC

cmadjar:
This PR cleans up the multiple functions to grep MINC header information and merges them into a single function in `MRI.pm`.

See also: https://redmine.cbrain.mcgill.ca/issues/15106

To test:
- [ ] test tarchiveLoader
- [x] test DTIPrep pipeline (to be noted, the DTIPrep_pipeline breaks when there are NIfTI files and MINC files for the t1, see issue #352, a separate PR will be issued to minor to fix this)
- [x] test register_processed_data.pl (to be noted, register_processed_data.pl is broken on major due to the call to the object broker MriScannerOB with not the proper database connection, see issue #353, a separate PR will be issued to major to fix this)

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -1540,6 +1540,42 @@ sub my_trim {
 	return $str;
 }
 
+=pod
+
+=head3 fetch_minc_header_info($minc, $field, $awk, $keep_semicolon)
+
+Function that fetches header information in MINC file.
+
+INPUTS:
+  - $minc : MINC file
+  - $field: field name to look for in MINC header (or 'all' to grep all headers)
+  - $awk  : awk info to check if argument inserted in MINC header
+  - $keep_semicolon: if set, keep ";" at the end of extracted value
+
+RETURNS: value of the field found in the MINC header
+
+=cut
+
+sub fetch_header_info {
+    my ($minc, $field, $awk, $keep_semicolon) = @_;
+    my $value;
+    if ($field eq 'all') {
+        # run mincheader and return all the content of the command
+        $value = `mincheader -data "$minc"`;
+    }
+    else {
+        # fetch a particular header value, remove extra spaces and optionally
+        # the semicolon
+        my $val = `mincheader $minc | grep $field | awk '{print $awk}' | tr '\n' ' '`;
+        $value = $val if $val !~ /^\s*"*\s*"*\s*$/;
+        return undef unless ($value); # return undef if no value found
+        $value =~ s/"//g;     # remove "
+        $value =~ s/^\s+//;   # remove leading spaces
+        $value =~ s/\s+$//;   # remove trailing spaces
+        $value =~ s/;// unless ($keep_semicolon);  # remove ";" unless $keep_semicolon is defined
+    }
+    return $value;
+}
 
 1;
 
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -1542,14 +1542,13 @@ sub my_trim {
 
 =pod
 
-=head3 fetch_minc_header_info($minc, $field, $awk, $keep_semicolon)
+=head3 fetch_minc_header_info($minc, $field, $keep_semicolon)
 
 Function that fetches header information in MINC file.
 
 INPUTS:
   - $minc : MINC file
   - $field: field name to look for in MINC header (or 'all' to grep all headers)
-  - $awk  : awk info to check if argument inserted in MINC header
   - $keep_semicolon: if set, keep ";" at the end of extracted value
 
 RETURNS: value of the field found in the MINC header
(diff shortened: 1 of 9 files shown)


uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -1540,6 +1540,42 @@ sub my_trim {
...
+INPUTS:
+  - $minc : MINC file
+  - $field: field name to look for in MINC header (or 'all' to grep all headers)
+  - $awk  : awk info to check if argument inserted in MINC header
+  - $keep_semicolon: if set, keep ";" at the end of extracted value
+
+RETURNS: value of the field found in the MINC header
+
+=cut
+
```
nicolasbrossard: I think using `awk` to extract the header values makes the function calls cumbersome since it has to contain an awk expression. I think it'd be better to use `cut` (see below).

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -1540,6 +1540,42 @@ sub my_trim {
...
+    my ($minc, $field, $awk, $keep_semicolon) = @_;
+    my $value;
+    if ($field eq 'all') {
+        # run mincheader and return all the content of the command
+        $value = `mincheader -data "$minc"`;
+    }
+    else {
+        # fetch a particular header value, remove extra spaces and optionally
+        # the semicolon
+        my $val = `mincheader $minc | grep $field | awk '{print $awk}' | tr '\n' ' '`;
```
nicolasbrossard: Since we know that MINC headers (and correct me if I'm wrong) are of the form

```perl
        xxxxxx:yyyyyy = value ;
```

where value can be surrounded by quotes, we can use this command instead

```
my $val = `mincheader $minc | grep $field | cut -d= -f2 | tr '\n' ' '`;
```

This will enable us to get rid of the `$awk` parameter.

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -1540,6 +1540,42 @@ sub my_trim {
...
+        $value = $val if $val !~ /^\s*"*\s*"*\s*$/;
+        return undef unless ($value); # return undef if no value found
+        $value =~ s/"//g;     # remove "
+        $value =~ s/^\s+//;   # remove leading spaces
```
nicolasbrossard: The removal of leading and trailing spaces should be done with a call to `chomp`.

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -1540,6 +1540,42 @@ sub my_trim {
...
+sub fetch_header_info {
+    my ($minc, $field, $awk, $keep_semicolon) = @_;
+    my $value;
+    if ($field eq 'all') {
+        # run mincheader and return all the content of the command
+        $value = `mincheader -data "$minc"`;
+    }
+    else {
+        # fetch a particular header value, remove extra spaces and optionally
+        # the semicolon
```
nicolasbrossard: Why is the `-data` switch used in the call to `mincheader` above but not in the one below? I am a bit surprised also that there is not an option in `mincheader` to retrieve a specific header (there is one in `dcmdump`). Oh well...

uploadNeuroDB/NeuroDB/MRI.pm
cmadjar: however, I just noticed I could have called the function in MRI.pm my_trim which removes the leading and trailing spaces.

uploadNeuroDB/NeuroDB/MRI.pm
cmadjar: From the mincheader manual:
```If the input file  is compressed, then its header is decompressed and only the  header  is  printed.  If  the -data option is given, then the whole file is decompressed and all data (except the image) is dumped.```

For fun I tried running mincheader on a MINC file with the option `-data` and without and compared the outputs. They are exactly the same. I think there is a difference in using the `-data` option in old `mnc.gz` files?

The reason it is in the first `if` statement is that it comes from the File.pm library which is old and probably needed to use that `-data` option on `mnc.gz` files. The `mincheader` in the second `if` statement was created way later and I did not know that a `-data` option existed ;).

Just to be on the safe side, I will add the option `-data` to the second statement.

cmadjar: done

uploadNeuroDB/NeuroDB/MRI.pm
cmadjar: done

uploadNeuroDB/NeuroDB/MRI.pm
cmadjar: used my_trim function in MRI.pm to do that

uploadNeuroDB/NeuroDB/MRI.pm
cmadjar: done

cmadjar:
@nicolasbrossard Ready for another round of review. Note that the suppression of $awk affected other functions that I needed to fix. Regardless, it is better to use `cut` than `awk`. Thanks for the suggestion!!
FYI: tested the `DTIPrep_pipeline.pl` and `register_processed_data.pl` after my last commit and it all seems to work as expected.

nicolasbrossard:
The changes look good but calls to function `my_trim` of `MRI.pm` should eventually be eliminated and replaced by calls to `chomp`, which does the same thing. Approving.

Issue #352, opened 2018-11-28
DTIPrep pipeline breaks when NIfTI files are available for the native t1


cmadjar:
If a project has the NIfTI file creation checked, the DTIPrep pipeline will break as it will try running some processes on the NIfTI files instead of the MINC file.

Change function `DTI::getAnatFile()` to make sure to return a MINC file and not a NIfTI file.

cmadjar:
PR merged

PR #355, opened 2018-11-28
Fix DTIPrep pipeline when create NIfTI is set to yes

cmadjar:
### Description

When a project uses NIfTI files creation, the DTIPrep pipeline breaks when searching for a t1 in the native subdirectory as it returns the NIfTI file instead of the MINC file. This fixes the issue and returns only the MINC file.

### This fixes...

-[x] issue #352

### DIFF
--- a/DTIPrep/DTI/DTI.pm
+++ b/DTIPrep/DTI/DTI.pm
@@ -162,14 +162,14 @@ RETURNS:
 sub getAnatFile {
     my ($nativedir, $t1_scan_type)  = @_;
 
-    # Fetch files in native directory that matched t1_scan_type
-    my $anat_list   = DTI::getFilesList($nativedir, $t1_scan_type);
+    # Fetch files in native directory that matched t1_scan_type and MINC file type
+    my $anat_list = DTI::getFilesList($nativedir, "\_$t1_scan_type\_.*mnc\$");
 
     # Return undef if no anat found, first anat otherwise
     if (@$anat_list == 0) { 
         return undef; 
     } else { 
-        my $anat    = @$anat_list[0];
+        my $anat = @$anat_list[0];
         return $anat;
     }
 }
--- a/DTIPrep/DTI/DTI.pm
+++ b/DTIPrep/DTI/DTI.pm
@@ -166,12 +166,7 @@ sub getAnatFile {
     my $anat_list = DTI::getFilesList($nativedir, "\_$t1_scan_type\_.*mnc\$");
 
     # Return undef if no anat found, first anat otherwise
-    if (@$anat_list == 0) { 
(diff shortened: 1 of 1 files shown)


DTIPrep/DTI/DTI.pm
```diff
@@ -162,14 +162,14 @@ RETURNS:
 sub getAnatFile {
     my ($nativedir, $t1_scan_type)  = @_;
 
-    # Fetch files in native directory that matched t1_scan_type
-    my $anat_list   = DTI::getFilesList($nativedir, $t1_scan_type);
+    # Fetch files in native directory that matched t1_scan_type and MINC file type
+    my $anat_list = DTI::getFilesList($nativedir, "\_$t1_scan_type\_.*mnc\$");
```
nicolasbrossard: Shouldn't the pattern passed as argument be

`"\_$t1_scan_type\_*.mnc\$"`

(watch the position of the `*`)?

cmadjar: this is a regex and if I understand correctly `.*` means that it matches any characters 0 or more times between the `$t1_scan_type` and `mnc` file extension.

nicolasbrossard: Actually, I think we want to check that the file name ends with this pattern:

1. An underscore, followed by
2. The scan type followed by
3. An underscore, followed by
4. Any non-zero length sequence of characters (except underscore, as these are used as delimiters in a MINC file name) and finally
5. The string `.mnc`

This is written as `"\_$t1_scan_type\_[^_]+\.mnc\$"`.

Makes sense?

DTIPrep/DTI/DTI.pm
```diff
@@ -162,14 +162,14 @@ RETURNS:
...
 
     # Return undef if no anat found, first anat otherwise
     if (@$anat_list == 0) { 
         return undef; 
     } else { 
-        my $anat    = @$anat_list[0];
+        my $anat = @$anat_list[0];
```
nicolasbrossard: Could be written as 

```perl
return @$anat_list ? @$anat_list[0] : undef;
```

cmadjar: Very good point. Should have thought of that...

nicolasbrossard:
I noticed that you dismissed my suggestion for the regex change. Is it that you don't agree with it or just that you forgot to implement the change?

cmadjar:
> I noticed that you dismissed my suggestion for the regex change. Is it that you don't agree with it or just that you forgot to implement the change?

@nicolasbrossard so weird. I thought I implemented it... This is done now :).

nicolasbrossard:
LGTM . 👍

PR #358, opened 2018-11-28
Move getSNRModalities() out of the profile file (prod file)

cmadjar:
### Description

Moves the function `getSNRModalities()` out of the profile file. The list of modalities to compute SNR on will now be stored in the Config module of LORIS under the Imaging Pipeline Section as shown in the screenshot below:
screen shot 2018-11-28 at 3 50 12 pm

 The function `computeSNR()` of `MRIProcessingUtility.pm` has been updated to grep that information from the Config module now instead of the profile file.

SQL for this changes are in https://github.com/aces/Loris/pull/4159

### This resolves...

- [x] Redmine: https://redmine.cbrain.mcgill.ca/issues/15721

### Caveat for existing projects:

Make sure you have run the patch from this LORIS PR (https://github.com/aces/Loris/pull/4159; link to be updated to final URL of the patch at the time of release documentation) and that you have updated the Config module with the modalities listed in the function `getSNRModalities` of your profile file.

From now on, the insertion scripts will rely on the values stored in the Config module instead of the profile file and the function `getSNRModalities` can be removed from your profile file.

### To test this

- [ ] Source the patch from the LORIS PR https://github.com/aces/Loris/pull/4159
- [ ] Run the tarchiveLoader and make sure it computes the SNR only for the modalities mentioned in the Config module
- [ ] Run the tool `BackPopulateSNRAndAcquisitionOrder.pl` on one `TarchiveID` for which you have previously removed the SNR values from `parameter_file` and make sure it computes the SNR as expected

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
+++ b/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
@@ -38,7 +38,7 @@ utilities
                             $subjectIDsref
                           );
 
-  $utility->computeSNR($TarchiveID, $ArchLoc, $profile);
+  $utility->computeSNR($TarchiveID, $ArchLoc);
   $utility->orderModalitiesByAcq($TarchiveID, $ArchLoc);
 
 =head1 DESCRIPTION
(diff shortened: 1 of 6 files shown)


cmadjar:
I added the blocked tag for now until we agree on which milestone this PR and the LORIS one should go to. To follow up with @ridz1208

cmadjar:
This should go to 20.2 so removing the Blocked label

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1742,68 +1742,65 @@ sub validateCandidate {
...
-                    }
-                    $file->setParameter('SNR', $SNR);
+    while (my $row = $minc_file_arr->fetchrow_hashref()) {
+        my $filename     = $row->{'File'};
+        my $fileID       = $row->{'FileID'};
+        my $fileScanType = $row->{'Scan_type'};
+        my $base         = basename($filename);
+        my $fullpath     = "$data_dir/$filename";
+        my $message;
+        if ( grep(/$fileScanType/, @{$modalities}) ) {
```
nicolasbrossard: Why not check for equality instead of a pattern match, as in 

```perl
if( grep($_ eq $fileScanType, @$modalities) ) {
```

If you have modalities like `T1` and `T1weighted` and you are checking for the modalities that match pattern `/T1/` then both will match and that's probably not what you want...

PR #359, opened 2018-11-28
Additional check to verify avoid duplication of imaging files (using SeriesUID/EchoTime)

cmadjar:
### Description

Example case: A file named `project_123456_V01_t1_001.mnc` with SeriesUID='1.2.3' and EchoTime='30' has been inserted into the files table. The study site relabel that scan session to V02 instead of V01 and reupload the DICOMs. When running the insertion scripts on those new DICOM, the file would be named `project_123456_V02_t1_001.mnc` and would have a different MD5sum than `project_123456_V01_t1_001.mnc` as the PatientName field has been modified prior to upload. Therefore, the current insertion script will insert a duplicate of a scan that was already in the database since the SeriesUID and EchoTime of this new file remains '1.2.3' and '30'.

This PR fixes the issue that a file could be reinserted in the `files` table if the MD5sum changed. It will now check that no file can be found in the `files` table with the combination SeriesUID/EchoTime (which is unique for each imaging file) in addition to the MD5sum check and exit `minc_insertion.pl` if there was a match found.

Note: in the case of Fieldmap or multi echo acquisitions, a given SeriesUID could be found in multiple files after conversion to MINC. This is why we are using the combination SeriesUID/EchoTime to verify that an imaging file is unique. To date, no exception was found to this rule.

### This resolves...

- [x] Redmine https://redmine.cbrain.mcgill.ca/issues/5498

### To test this

- [ ] with the code from aces/minor, reupload a scan that was already uploaded in the past but with a different PatientName and notice that all acquisitions are reinserted into the `files` table creating duplicates
- [ ] check out this branch and reupload the scan with a different PatientName and notice that no duplicates will be created this time and the LOG will say:
```
ERROR: there is already a file registered in the files table with SeriesUID='1.2.3' and EchoTime='30'.
The already registered file is 'assembly/123456/V01/mri/native/project_123456_V01_t1_001.mnc'
```

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
+++ b/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
@@ -193,7 +193,7 @@ sub lookupNextVisitLabel {
 
 =pod
 
-=head3 getFileNamesfromSeriesUID($seriesuid, @alltarfiles)
+=head3 getDICOMFileNamesfromSeriesUID($seriesuid, @alltarfiles)
 
 Will extract from the C<tarchive_files> table a list of DICOM files
 matching a given C<SeriesUID>.
@@ -206,7 +206,7 @@ RETURNS: list of DICOM files corresponding to the C<SeriesUID>
 
 =cut
 
-sub getFileNamesfromSeriesUID {
+sub getDICOMFileNamesfromSeriesUID {
 
     # longest common prefix
     sub LCP {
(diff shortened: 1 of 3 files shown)


uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1997,6 +1997,60 @@ sub isValidMRIProtocol  {
...
+    my $this = shift;
+    my ($file, $upload_id) = @_;
+
+    my $seriesUID = $file->getParameter( 'series_instance_uid' );
+    my $echo_time = $file->getParameter( 'echo_time'           );
+
+    # check that no files are already in the files table with the same SeriesUID
+    # and EchoTime
+    my $query     = "SELECT File FROM files WHERE SeriesUID=? AND EchoTime=?";
+    my $sth       = ${$this->{'dbhr'}}->prepare( $query );
```
nicolasbrossard: Not really related to your PR but it is very annoying to have to dereference `$this->{'dbhr'}` every time we use it (as in `${ $this->{'dbhr'} }`). The constructor for the MRIProcessingUtility class should receive the `$dbhr` argument and not a reference to it and the usage of this variable would be simplified greatly. I opened [Redmine ticket #15738](https://redmine.cbrain.mcgill.ca/issues/15738) for that.

PR #361, opened 2018-11-30
Non DICOM files are now ignored when running the pipeline.

nicolasbrossard:
When trying to insert an archive that contained non-DICOM files, the pipeline would detect this and would not insert anything. Insertion of the DICOM files now takes place even if non-DICOM files are found in the archive. Fixes [Redmine ticket #15451](https://redmine.cbrain.mcgill.ca/issues/15451).

### DIFF
--- a/uploadNeuroDB/NeuroDB/ImagingUpload.pm
+++ b/uploadNeuroDB/NeuroDB/ImagingUpload.pm
@@ -247,55 +247,50 @@ sub IsCandidateInfoValid {
         ########################################################
         #1) Exlcude files starting with . (and ._ as a result)##
         #including the .DS_Store file###########################
-        #2) Check to see if the file is of type DICOM###########
+        #2) Exclude files that are not of type DICOM ###########
         #3) Check to see if the header matches the patient-name#
         ########################################################
-        if ( (basename($_) =~ /^\./)) {
+        if ( (basename($_) =~ /^\./) || !$this->isDicom($_)) {
             $cmd = "rm " . ($_);
             print "\n $cmd \n";
             system($cmd);
+            # increment counter only for non-DICOM files that do not
+            # start with '.'
+            $files_not_dicom++ if basename($_) !~ /^\./;
         }
         else {
-            if ( ( $_ ne '.' ) && ( $_ ne '..' )) {
-                if ( !$this->isDicom($_) ) {
-                    $files_not_dicom++;
-                }
-    	        else {
             #######################################################
             #Validate the Patient-Name, only if it's not a phantom#
             ############## and the file is of type DICOM###########
             #######################################################
-                    if ($row[4] eq 'N') {
-                        # make sure the regex used for PatientNameMatch starts
-                        # with $this->{'pname'}
-                        if (!$this->PatientNameMatch($_, "^$this->{'pname'}")) {
-                            $files_with_unmatched_patient_name++;
-                        }
-                    } elsif ($row[4] eq 'Y') {
-                        # make sure the regex used for PatientNameMatch
-                        # includes "phantom" string
-                        my $lego_phantom_regex = NeuroDB::DBI::getConfigSetting(
-                            $this->{dbhr}, 'LegoPhantomRegex'
-                        );
-                        my $living_phantom_regex = NeuroDB::DBI::getConfigSetting(
-                            $this->{dbhr}, 'LivingPhantomRegex'
-                        );
-                        my $phantom_regex =
-                            "($lego_phantom_regex)|($living_phantom_regex)";
-                        if (!$this->PatientNameMatch($_, $phantom_regex)) {
-                            $files_with_unmatched_patient_name++;
-                        }
-                    }
+            if ($row[4] eq 'N') {
+                # make sure the regex used for PatientNameMatch starts
+                # with $this->{'pname'}
+                if (!$this->PatientNameMatch($_, "^$this->{'pname'}")) {
+                   $files_with_unmatched_patient_name++;
+                }
+            } elsif ($row[4] eq 'Y') {
+                # make sure the regex used for PatientNameMatch
+                # includes "phantom" string
+                my $lego_phantom_regex = NeuroDB::DBI::getConfigSetting(
+                    $this->{dbhr}, 'LegoPhantomRegex'
+                );
+                my $living_phantom_regex = NeuroDB::DBI::getConfigSetting(
+                    $this->{dbhr}, 'LivingPhantomRegex'
+                );
+                my $phantom_regex =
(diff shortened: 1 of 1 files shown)


uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -247,55 +247,50 @@ sub IsCandidateInfoValid {
...
         }
     }
 
     if ( $files_not_dicom > 0 ) {
-        $message = "\nERROR: There are $files_not_dicom file(s) which"
-          . " are not of type DICOM \n";
-        $this->spool($message, 'Y', $notify_notsummary);
-        return 0;
+        $message = "\nWARNING: There are $files_not_dicom file(s) which"
+          . " are not of type DICOM: these will be ignored\n";
```
cmadjar: Or maybe in the warning message, we could show the list of files that were ignored or removed from the archive so at least you can keep a trace of what was removed?

cmadjar: I am worried that you mind end up erasing DICOM files that are not of type DICOM or files that people might still want to keep in the archive for some reason. 

I would have a tendency to only execute the command for files that starts with `.`. What do you think?

I don't have a strong opinion about this though. Just thought I would raise this since it crossed my mind.

uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -247,55 +247,50 @@ sub IsCandidateInfoValid {
...
         #including the .DS_Store file###########################
-        #2) Check to see if the file is of type DICOM###########
+        #2) Exclude files that are not of type DICOM ###########
         #3) Check to see if the header matches the patient-name#
         ########################################################
-        if ( (basename($_) =~ /^\./)) {
+        if ( (basename($_) =~ /^\./) || !$this->isDicom($_)) {
             $cmd = "rm " . ($_);
             print "\n $cmd \n";
             system($cmd);
```
cmadjar: I am worried that you mind end up erasing DICOM files that are not of type DICOM or files that people might still want to keep in the archive for some reason.

I would have a tendency to only execute the command for files that starts with .. What do you think?

I don't have a strong opinion about this though. Just thought I would raise this since it crossed my mind.

cmadjar: (this comment was supposed to be part of my previous review...)

davidblader:
i tried inserting an upload that contained an embedded zip in it, which normally crashes the pipeline at the dicomTar step

it managed to get to the tarchiveLoader step, but i got a bunch of `Error opening file UNDEF!` notices and 
`Number of MINC files that will be considered for inserting into the database: 0
No data could be converted into valid MINC files.
localizer, scout will not be considered! `

davidblader:
it works!

uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -247,55 +247,50 @@ sub IsCandidateInfoValid {
...
-                        my $phantom_regex =
-                            "($lego_phantom_regex)|($living_phantom_regex)";
-                        if (!$this->PatientNameMatch($_, $phantom_regex)) {
-                            $files_with_unmatched_patient_name++;
-                        }
-                    }
+            if ($row[4] eq 'N') {
+                # make sure the regex used for PatientNameMatch starts
+                # with $this->{'pname'}
+                if (!$this->PatientNameMatch($_, "^$this->{'pname'}")) {
```
davidblader: i think putting the `^` within the string causes Perl to ignore it as a regex operator,
trying to insert a scan with this change causes:
`The PatientID read from the DICOM header does not start with ^VIA1847_809669_Initial_MRI from the mri_upload table`

uploadNeuroDB/NeuroDB/ImagingUpload.pm
paiva: I agree with Cecile here. I would be very useful to have the list of files that were ignore or removed, it can help identify files that should have not been uploaded

uploadNeuroDB/NeuroDB/ImagingUpload.pm
paiva: This is a very good point. It happened in the Quad project with Rick. There were some non-dicom files that contained some metadata or something. I don't think that these non-dicom files should be removed.

nicolasbrossard: If you have come across projects that needed to keep files in the archive even though they were not DICOM files, then yeah we should leave the archive as is and not delete anything (maybe only the files that start with ".") 

@cmadjar What do you mean when you say "...end up erasing DICOM files that are not of type DICOM"?

cmadjar: @nicolasbrossard I think someone (can't remember which project) at some point that came up with a DICOM dataset but the files were not of type DICOM for some reason when typing in the terminal `file dicom.dcm`.

uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -247,55 +247,50 @@ sub IsCandidateInfoValid {
...
-                            $this->{dbhr}, 'LivingPhantomRegex'
-                        );
-                        my $phantom_regex =
-                            "($lego_phantom_regex)|($living_phantom_regex)";
-                        if (!$this->PatientNameMatch($_, $phantom_regex)) {
-                            $files_with_unmatched_patient_name++;
-                        }
-                    }
+            if ($row[4] eq 'N') {
+                # make sure the regex used for PatientNameMatch starts
```
PapillonMcGill:
```suggestion
                # ensure the regex used for PatientNameMatch starts
```
more formal than `make sure`

uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -247,55 +247,50 @@ sub IsCandidateInfoValid {
...
+                # with $this->{'pname'}
+                if (!$this->PatientNameMatch($_, "^$this->{'pname'}")) {
+                   $files_with_unmatched_patient_name++;
+                }
+            } elsif ($row[4] eq 'Y') {
+                # make sure the regex used for PatientNameMatch
```
PapillonMcGill:
```suggestion
                # ensure the regex used for PatientNameMatch
```

nicolasbrossard:
Closing this PR since it is now rendered obsolete with the merging of [PR #371](https://github.com/aces/Loris-MRI/pull/371).

nicolasbrossard:
Closing this PR since it is now rendered obsolete with the merging of [PR #371](https://github.com/aces/Loris-MRI/pull/371).

Issue #364, opened 2018-12-12
Remove DICOM header (0020,0032) from check to decide whether to include/reject a DICOM file


cmadjar:
We should not let the absence/presence of DICOM header (0020,0032) be a decisive factor in deciding whether to include/reject a DICOM file in the pipeline’s archiving process.
=> Get rid of that check when not using option `-slicepos` of `get_dicom_info.pl`.

_Discussed at the imaging meeting of December 7th, 2018 where everyone agreed on the removal of that check_

PR #365, opened 2018-12-12
Header (0020,0032) now only checked when -slicepos is passed on the command line

nicolasbrossard:
This PR modifies the behaviour of script `get_dicom_info.pl`: DICOM files passed as arguments to this script are now reported as not being DICOM only if they __don't__ have header `(0020,0032)` __and__ `-slicepos` was passed on the command line.

`dcmodify -e` can be useful during testing if one wants to remove header `(0020,0032)` from an existing DICOM file.

This fixes [issue #364](https://github.com/aces/Loris-MRI/issues/364)

### DIFF
--- a/dicom-archive/get_dicom_info.pl
+++ b/dicom-archive/get_dicom_info.pl
@@ -136,7 +136,8 @@ foreach my $filename (@input_list) {
     my(@position) = 
 	# ImagePositionPatient (0x0020, 0x0032)
         &convert_coordinates(&split_dicom_list(&trim($dicom->value('0020', '0032'))));
-    if (scalar(@position) != 3) {
+    my $computeSlicePos = grep($_->[1] eq 'slicepos', @Variables);
+    if (scalar(@position) != 3 && $computeSlicePos) {
        warn "Warning: The file: $filename is not DICOM!\n";
        push my @croft, $filename;
        next;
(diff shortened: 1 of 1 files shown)


dicom-archive/get_dicom_info.pl
```diff
@@ -136,7 +136,8 @@ =head2 Methods
     my(@position) = 
 	# ImagePositionPatient (0x0020, 0x0032)
         &convert_coordinates(&split_dicom_list(&trim($dicom->value('0020', '0032'))));
-    if (scalar(@position) != 3) {
+    my $computeSlicePos = grep($_->[1] eq 'slicepos', @Variables);
+    if (scalar(@position) != 3 && $computeSlicePos) {
        warn "Warning: The file: $filename is not DICOM!\n";
```
cmadjar: @nicolasbrossard I wonder if we should not update the warning error to something clearer. Like "Warning: the file $filename is either not DICOM or does not contain any value for DICOM header (0020,0032)"

nicolasbrossard: @cmadjar Since header (0020,0032) is not mandatory, I vote for changing the warning to

 "DICOM header (0020,0032) not found in $filename: slice position cannot be computed. Skipping file.\n"

What do you think?

cmadjar: Perfect!!!

PR #369, opened 2019-01-16
Add support for multiple MRI protocols.

nicolasbrossard:
### Description

This PR adds the possibility to have multiple MRI protocols for a given study and multiple groups of MRI protocol checks. You can now group scan type specifications listed in `mri_protocol` and have specific groups taken into account __or__ ignored based on the scan properties (visit at which the scan was taken, etc...). Similarly, you can now group lines of the `mri_protocol_check` and specify under which conditions specific groups are taken into account __or__ ignored for a given scan. In a nutshell, this grouping feature specifies which lines of table `mri_protocol` and `mri_protocol_checks` are active or inactive for a given scan. 

### How to define new MRI protocol groups and MRI protocol checks groups

First you have to run the SQL patch that updates the database schema for this new feature. Once you have done this, the behaviour of the pipeline should not have changed: all the lines of the `mri_protocol` and `mri_protocol_checks` will be active for all scans . To create a new MRI protocol group:

**Step 1:** Insert a record in `mri_protocol_group` with the new group name and ID.  

**Step 2:** Associate each line in table `mri_protocol` to a protocol group in `mri_protocol_group` by filling out column`MriProtocolGroupID`.

**Step 3:** Determine the conditions under which the new group will be used to identify the type of a scan. Conditions include the project ID, sub-project ID or visit label of the session at which the scan was done. There are columns in table `mri_protocol_group_target` for each of these types of conditions. When more than one column is filled in this table, the conditions are **AND**ed together.

The steps to create a new MRI protocol checks group are analog to those described above but use tables `mri_protocol_checks_group` and `mri_protocol_checks_group_target` instead.

Each line of the `mri_protocol` table belongs to either one group or no group at all, and this will be enforced in the MySQL schema. Lines of `mri_protocol_checks` can belong to 0, 1 or many protocol checks group though.

In addition to creating the new tables described above, this PR also modifies table `mri_violated_scans` and adds a new column that contains the MRI protocol group that was used when trying to identify the scan type. When the MRI cannot determine the type of a scan because the scan does not match any of the conditions listed in `mri_protocol_group_target`, the value in this new column will be set to `NULL`. 

### How to test

Make sure you have sourced the SQL patch that extends the database in order to support multiple MRI protocols. Create new MRI protocol groups and new groups of MRI protocol check and make sure that the correct groups are used by the MRI pipeline according to the scan properties.

### See also

Associated Redmine ticket: https://redmine.cbrain.mcgill.ca/issues/15452
Associated PR on the LORIS side: [https://github.com/aces/Loris/pull/4321](https://github.com/aces/Loris/pull/4321)


uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -548,18 +551,45 @@ sub identify_scan_db {
...
 
+    $query .= defined $projectID
+        ? ' AND (mpgt.ProjectID IS NULL OR mpgt.ProjectID = ?)'
+        : ' AND mpgt.ProjectID IS NULL';
+    $query .= defined $subprojectID
+        ? ' AND (mpgt.SubprojectID IS NULL OR mpgt.SubprojectID = ?)'
+        : ' AND mpgt.SubprojectID IS NULL';
+    $query .= defined $visitLabel
+        ? ' AND (mpgt.Visit_label IS NULL OR mpgt.Visit_label = ?)'
+        : ' AND mpgt.Visit_label IS NULL';
```
cmadjar: Let's say you have 2 t1 protocols (one with the default protocol and one specific to a project):
- t1 default protocol: TR=2000-2400
- t1 project X protocol: TR=2100
What happens if for project X you try to insert a t1 with TR=2200. Theoretically, you don't want it to be inserted but wouldn't it get inserted anyway because of the t1 default protocol that would allow for its insertion?

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -374,15 +376,54 @@ sub determineSubjectID {
...
+    #==============================================#
+    # Retrieve the candidate's project ID (if any) #
+    # and session's subproject ID                  #
+    #==============================================#
+    (my $query = <<QUERY) =~ s/\n/ /gm;
+    SELECT
+      c.ProjectID, s.SubprojectID
+    FROM
+      session s 
+    JOIN candidate c USING(CandID)
```
cmadjar: missing space between USING and (

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -548,18 +551,45 @@ sub identify_scan_db {
...
+    #===========================================================#
+    ($query = <<QUERY) =~ s/\n/ /gm;
+    SELECT mp.Scan_type, mp.ScannerID, mp.Center_name, 
+           mp.TR_range, mp.TE_range, mp.TI_range, mp.slice_thickness_range,
+           mp.xspace_range, mp.yspace_range, mp.zspace_range,
+           mp.xstep_range, mp.ystep_range, mp.zstep_range, 
+           mp.time_range, mp.series_description_regex
+    FROM mri_protocol mp
+    JOIN mri_protocol_group_rel mpgr ON (mp.ID=mpgr.MriProtocolID)
+    JOIN mri_protocol_group_target mpgt USING(MriProtocolGroupID)
```
cmadjar: missing space between USING and (

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -374,15 +376,54 @@ sub determineSubjectID {
...
+QUERY
+
+    my $sth = $dbh->prepare($query);
+    $sth->execute($subjectIDsref->{'visitLabel'}, $subjectIDsref->{'CandID'});
+    
+    #=====================================================#
+    # Number of results has to be 1 or exit with failure  #
+    #=====================================================#
+    if($sth->rows == 0) {
+		$subjectIDsref->{'ProjectID'}    = undef;
```
cmadjar: indent is  weird for that if else statement.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -734,14 +789,19 @@ sub extra_file_checks() {
     # check if the scan is valid. If the scan is not valid, then, return the
     # severity of the failure.
     foreach my $severity (qw/exclude warning/) {
-        $sth->execute($scan_type, $severity);
+		my @bindValues = ($scan_type, $severity);
```
cmadjar: weird indent in the foreach loop

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -762,25 +822,39 @@ a hash with all the checks that need to be applied on that specific scan type
...
 
-    # query to fetch list of valid protocols in mri_protocol_checks table
-    my $query = "SELECT * FROM mri_protocol_checks "
-                . "WHERE Scan_type=? AND Severity=? AND Header=?";
+    # query to fetch list of valid protocol checks in mri_protocol_checks table
+    my $query = "SELECT * FROM mri_protocol_checks mpc "
+        . "JOIN mri_protocol_checks_group_rel mpcgr ON (mpc.ID=mpcgr.MriProtocolChecksGroupID) "
+        . "JOIN mri_protocol_checks_group_target mpcgt USING(MriProtocolChecksGroupID) "
+        . "WHERE mpc.Scan_type=? AND mpc.Severity=? AND mpc.Header=?";
+    $query .= defined $projectID
```
cmadjar: same question as for the mri_protocol table. What happens if a project has a stricter t1 protocol than the generic t1 protocol?

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
nicolasbrossard: I think this is valid SQL syntax. I assume the space is for clarity?

cmadjar: It is valid SQL. It's just that my OCD misses the space ;)

uploadNeuroDB/NeuroDB/MRI.pm
cmadjar: haha. OK. Will wait for the next commit then. Thanks!

christinerogers:
could you add link to the associated Loris PR (for SQL) in this PR header? thanks! @nicolasbrossard

cmadjar:
@nicolasbrossard Another thing I just thought about. Would be good to update the MRI documentation with the new tables and the new structure :).

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -374,15 +376,54 @@ sub determineSubjectID {
...
+    (my $query = <<QUERY) =~ s/\n/ /gm;
+    SELECT
+      c.ProjectID, s.SubprojectID
+    FROM
+      session s 
+    JOIN candidate c USING(CandID)
+    WHERE
+      s.Visit_label = ?
+    AND
+      s.CandID = ?
```
PapillonMcGill: Should not you use where c.Active = 'Y' and s.Active = 'Y'?   c.active can probably be omitted as the candidate should be active to add MRI.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -762,25 +822,39 @@ a hash with all the checks that need to be applied on that specific scan type
...
-    my ($this, $scan_type, $severity, $headers, $file) = @_;
+    my ($this, $scan_type, $severity, $headers, $file, $projectID, $subprojectID, $visitLabel) = @_;
 
     my %valid_fields; # will store all information about what fails
 
-    # query to fetch list of valid protocols in mri_protocol_checks table
-    my $query = "SELECT * FROM mri_protocol_checks "
-                . "WHERE Scan_type=? AND Severity=? AND Header=?";
+    # query to fetch list of valid protocol checks in mri_protocol_checks table
+    my $query = "SELECT * FROM mri_protocol_checks mpc "
```
PapillonMcGill: Do you nedd 'SELECT * ...'? 
Could you use the attribute's list in the SELECT?

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
nicolasbrossard: Ok

uploadNeuroDB/NeuroDB/MRI.pm
nicolasbrossard: Ok

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
nicolasbrossard: Corrected

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
nicolasbrossard: Good point. And yes, we should also add `c.active='Y'` since a candidate can be scanned and be inactivated afterwards.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
nicolasbrossard: It is probably clearer to explicitly state what you are `SELECT`ing. Will change.

nicolasbrossard:
Made the corrections suggested by cmadjar and PapillonMcGill. Ready for another round of review! Note that I corrected a lot of incorrect indentation in `MRI.pm` and `MRIProcessingUtility.pm` so even though you see a lot more modifications that in the first commit, there are very few "real" changes to the code.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
PapillonMcGill: Yes it make sense. I wasn't really sure myself if active flag where important here.

cmadjar:
@nicolasbrossard Sorry, the PR needs rebasing.

paiva:
Hi @nicolasbrossard, please send me some sample data to test this PR when you have some time. I will set this up while we wait for the rebasing. Thanks!

paiva:
Hi @nicolasbrossard, just touching base again on this PR. Please let me know when it is ready for testing. Thanks so much

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -376,21 +378,49 @@ sub identify_scan_db {
...
-              OR ((Center_name='ZZZZ' OR Center_name='AAAA') AND ScannerID='0')
-              ORDER BY Center_name ASC, ScannerID DESC";
+    #===========================================================#
+    # Get the list of lines in the mri_protocol table that      #
+    # apply to the given scan based on the center name, project # 
+    # ID, subproject ID and visit label                         #
+    #===========================================================#
+    ($query = <<QUERY) =~ s/\n/ /gm;
+    SELECT mp.Scan_type, mp.ScannerID, mp.Center_name, 
+           mp.TR_range, mp.TE_range, mp.TI_range, mp.slice_thickness_range,
```
cmadjar: I think something went wrong with the rebase cause there are no more `*_range` fields. They all have been replaced by `*_min` `*_max`. Can you please make sure to have the latest code for that function?

nicolasbrossard:
@cmadjar Bizarre error you got...It's not even in the code I modified. Is it possible you have an entry in table `tarchive` with `ArchiveLocation` set to NULL? That's what the error message seems to suggest

nicolasbrossard:
@cmadjar Good catch for the missing field in table `mri_protocol_checks`. Corrected. As for the missing losanges, I believe they are used to identify fields that are defined as `KEY`s in MySQL. Since the new tables I created do not have any `KEY`s defined, they do not have any losanges.  The position of the foreign key arrow's start and end seem OK to me also. Do you have a specific example?

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -730,42 +732,44 @@ sub getAcquisitionProtocol {
...
-                        $acquisitionProtocolID, 
-                        $file, 
-                        $subjectIDsref->{'CandID'}, 
-                        $subjectIDsref->{'visitLabel'},
-                        $tarchiveInfoRef->{'PatientName'}
-                    );
-          $message = "\nextra_file_checks from table mri_protocol_check " .
+                $acquisitionProtocolID, 
+                $file, 
+                $subjectIDsref->{'CandID'},
```
cmadjar: c'est icitttttt

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -313,21 +315,23 @@ INPUTS:
...
-    ) = @_;
+    my  ($psc, $subjectref, $tarchiveInfoRef, $fileref, $dbhr, $db, $minc_location, $uploadID) = @_;
 
-    my $candid = ${subjectref}->{'CandID'};
-    my $pscid = ${subjectref}->{'PSCID'};
-    my $visit = ${subjectref}->{'visitLabel'};
+    my $candid       = ${subjectref}->{'CandID'};
+    my $pscid        = ${subjectref}->{'PSCID'};
+    my $visitLabel   = ${subjectref}->{'visitLabel'};
+    my $projectID    = ${subjectref}->{'ProjectID'};
```
cmadjar: `ProjectID` and `SubprojectID` keys are not present in the `$subjectref` hash passed here so the only returned group is the default with value NULL for ProjectID, SubprojectID and visit label.

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -119,71 +121,77 @@ INPUTS:
...
-    my ($subjectIDref, $studyDate, $dbhr, $objective, $db) = @_;
+    my ($subjectIDref, $studyDate, $dbhr, $db) = @_;
     my ($sessionID, $studyDateJD);
     my ($query, $sth);
     my $dbh = $$dbhr;
 
-# find a matching timepoint
-    $query = "SELECT ID, Date_visit, Visit FROM session WHERE CandID=$subjectIDref->{'CandID'} AND LOWER(Visit_label)=LOWER(".$dbh->quote($subjectIDref->{'visitLabel'}).") AND Active='Y'";
+    # find a matching timepoint
+    $query = "SELECT ID, Date_visit, Visit "
```
cmadjar: shouldn't you select the projet and subproject of the visit here?

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -119,71 +121,77 @@ INPUTS:
...
+                    push @files, $filehr->{'FileID'};
+                }
+                $sth->finish();
+            }
+        }
+    #####  if there is no existing session, which always happens if you create candidates based on incoming data
     } else {
-
-	# determine the visit number and centerID for the next session
+		# Since we'll be creating a visit, ensure that the subprojectID has been supplied
```
cmadjar: indent is off for that block of code

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -208,41 +216,49 @@ sub getSessionID {
...
+        #### insert the new session setting Current_stage to 'Not started' because that column is important to the behavioural data entry gui.
+        $query = "INSERT INTO session "
+               . "SET CandID=".$dbh->quote($subjectIDref->{'CandID'}).", "
+               . "    Visit_label=".$dbh->quote($subjectIDref->{'visitLabel'}).", "
+               . "    CenterID=$centerID, "
+               . "    VisitNo=$newVisitNo, "
+               . "    Current_stage='Not Started', "
+               . "    Scan_done='Y', "
+               . "    Submitted='N', "
+               . "    SubprojectID=".$dbh->quote($subjectIDref->{'SubprojectID'});
```
cmadjar: shouldn't this contain the ProjectID too?

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -119,71 +121,77 @@ INPUTS:
...
+                $sth->finish();
+            }
+        }
+    #####  if there is no existing session, which always happens if you create candidates based on incoming data
     } else {
-
-	# determine the visit number and centerID for the next session
+		# Since we'll be creating a visit, ensure that the subprojectID has been supplied 
+		# in the profile file
+		return undef if !defined $subjectIDref->{'SubprojectID'};
```
cmadjar: what about the ProjectID?

dicom-archive/profileTemplate.pl
```diff
@@ -49,6 +49,12 @@ sub getSubjectIDs {
         $subjectID{'visitLabel'} = NeuroDB::MRI::my_trim($patientName);
         $subjectID{'isPhantom'} = 1;
 
+        # If config setting 'useProject' and 'createCandidates' are true
+        # $subjectID{'ProjectID'} must be set.
+        $subjectID{'ProjectID'} = 1;
+
```
cmadjar: what about the SubprojectID?

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -259,75 +275,65 @@ INPUTS:
...
-
-    my $query = "SELECT SubprojectID FROM session WHERE CandID='$subjectIDs{'CandID'}' AND Visit_label='$subjectIDs{'visitLabel'}' AND Active='Y' ORDER BY ID DESC LIMIT 1";
-    my $sth = $${dbhr}->prepare($query) or die "Can't prepare $query: ".$${dbhr}->errstr."\n";
 
-    $sth->execute();
-
-    if($sth->rows > 0) {
-        @results = $sth->fetchrow_array();
-    }
+    my $query = "SELECT SubprojectID "
```
cmadjar: should also select the ProjectID of the session since protocol can be restricted per Project or Subproject

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -259,75 +275,65 @@ INPUTS:
...
-        @results = $sth->fetchrow_array();
-
-        $objective = $results[0] if $results[0];
-    }
-    return $objective;
+    $sth->execute($subjectIDs{'CandID'}, $subjectIDs{'visitLabel'});
 
+    return undef if $sth->rows == 0;
+       
+    @results = $sth->fetchrow_array();
```
cmadjar: since will return both SubprojectID and ProjectID, will probably be easier to grep a hash.

Also, this function is missing the return statement ;)

uploadNeuroDB/minc_insertion.pl
```diff
@@ -538,11 +538,26 @@ =head2 Methods
...
         $studyInfo{'DateAcquired'},
         \$dbh,
-        $subjectIDsref->{'subprojectID'},
         $db
 
-   );
-
+);
+ 
+$subjectIDsref->{'SubprojectID'} = NeuroDB::MRI::getObjective($subjectIDsref, $dbh);
```
cmadjar: shouldn't there be a check for ProjectID here too?

dicom-archive/profileTemplate.pl
nicolasbrossard: The more I think about it, the more I feel like not providing a default for ProjectID and SubprojectID: I'd rather indicate with an appropriate comment that these must be set when `createCandidate`/`createVisitLabel` are turned on. There is not even a default project defined in `0000-00-00-schema.sql` so `$subjectID{'ProjectID'} = 1` does not even make much sense when I think about it...Let me know what you think.

uploadNeuroDB/NeuroDB/MRI.pm
nicolasbrossard: Since the function is called `getSessionID`, only the session ID is returned. The sub-project ID is computed in method `getObjective`.

Also, I checked on branch `master` and there are no project IDs associated to sessions, only sub-project IDs. AM I missing something. When determining which protocol to use we only consider the **candidate** project ID anyway (i.e. the project to which a candidate is originally registered).

uploadNeuroDB/NeuroDB/MRI.pm
nicolasbrossard: Ok, I'll correct.

uploadNeuroDB/NeuroDB/MRI.pm
nicolasbrossard: See comment above.

uploadNeuroDB/NeuroDB/MRI.pm
nicolasbrossard: See previous comments.

uploadNeuroDB/NeuroDB/MRI.pm
nicolasbrossard: Arghhh.....!!!!! Corrected.

uploadNeuroDB/NeuroDB/MRI.pm
nicolasbrossard: Since previous comments.

uploadNeuroDB/minc_insertion.pl
nicolasbrossard: There is a call the `determineSubjectID` a couple of lines further up and this method will validate that a ProjectID was supplied, so no need to do this twice.

uploadNeuroDB/NeuroDB/MRI.pm
cmadjar: OK. Makes sense to select the subproject in `getObjective`. Probably should select the `ProjectID` there too or create a new function for it?

Regarding your last comment about no ProjectID in `session`, I just checked `master` on the LORIS repo and there is a `ProjectID` field in the `session` table (see https://github.com/aces/Loris/blob/12a235e511b810d35c39475db1b48f04fa8e85c9/SQL/0000-00-00-schema.sql#L182) so it might be a good idea to select the `ProjectID` of the session in the MRI scripts. What do you think?

uploadNeuroDB/NeuroDB/MRI.pm
cmadjar: I am guessing you referred to your comments saying there is no ProjectID in session table? ProjectID is in the session table and should also be filled here. See previous comment on the topic

uploadNeuroDB/NeuroDB/MRI.pm
cmadjar: same comment. either create a new function to get the ProjectID of the session or add it here.

uploadNeuroDB/NeuroDB/MRI.pm
cmadjar: hihihi ;)

dicom-archive/profileTemplate.pl
cmadjar: I looked at the default schema and it looks like this [line](https://github.com/aces/Loris/blob/12a235e511b810d35c39475db1b48f04fa8e85c9/SQL/0000-00-00-schema.sql#L13) would create a ProjectID = 1 so I guess there would be a default ProjectID = 1 anyway?

uploadNeuroDB/minc_insertion.pl
cmadjar: I don't see where in `determineSubjectID` the `ProjectID` is validated. Could you point it to me? Thanks!

uploadNeuroDB/minc_insertion.pl
```diff
@@ -534,15 +534,28 @@ =head2 Methods
...
-
-   );
-
+my($sessionRef, $errMsg) = NeuroDB::MRI::getSession(
+    $subjectIDsref, 
+    $studyInfo{'DateAcquired'},
+    $dbh,
+    $db
+);
+$subjectIDsref->{'SessionID'} = $sessionRef->{'ID'};
```
cmadjar: should also add the `ProjectID` and `SubprojectID` key to the `$subjectIDsref` otherwise, `identify_scan_db` does not have that information (the keys are only present in `$sessionRef ` after the call to `getSession`

uploadNeuroDB/imaging_non_minc_insertion.pl
```diff
@@ -461,23 +461,25 @@ =head2 Methods
...
+        'imaging non minc file insertion', $errMsg,   0,
+        'imaging_non_minc_insertion.pl' , $upload_id, 'Y',
         'N'
     );
-    exit $NeuroDB::ExitCodes::GET_SESSION_ID_FAILURE;
+    exit ($subjectIDsref->{'createVisitLabel'} == 1
+              ? $NeuroDB::ExitCodes::CREATE_SESSION_FAILURE
+              : $NeuroDB::ExitCodes::GET_SESSION_ID_FAILURE);
 }
+my $session_id = $sessionRef->{'ID'};
```
cmadjar: should also add the `ProjectID` and `SubprojectID` key to the `$subjectIDsref` otherwise, `identify_scan_db` does not have that information (the keys are only present in `$sessionRef ` after the call to `getSession`

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -381,43 +354,93 @@ sub identify_scan_db {
...
+    # all the cases. Warn.
+    if(@rows == 0) {
+		my $msg = "Warning! No protocol group can be used to determine the scan type "
+		        . "of $minc_location.\nIncorrect/incomplete setup of table mri_protocol_group_target: "
+		        . " setting scan type to 'unknown'"; 
+		print "$msg\n";
+        my $notify = NeuroDB::Notify->new( $dbhr );
+        $notify->spool('mri upload processing class', $msg, 0, 'MRI.pm', $uploadID, 'N', 'Y');
+
+        return 'unknown' unless $sth->rows>0;
```
cmadjar: alignment is off

cmadjar: not sure I understand why there is an `unless $sth->rows>0` since we already established here that `@rows == 0`

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -381,43 +354,93 @@ sub identify_scan_db {
...
+    # then table mri_protocol_group_target was not setup properly. Warn.
+    my %mriProtocolGroupIDs = map { $_->{'MriProtocolGroupID'} => 1 } @rows;
+    if(keys %mriProtocolGroupIDs > 1) {
+		my $msg = "Warning! More than one protocol group can be used to identify the scan type of $minc_location\n"
+		        . "Ambiguous setup of table mri_protocol_group_target: setting scan type to 'unknown'"; 
+		print "$msg\n";
+        my $notify = NeuroDB::Notify->new( $dbhr );
+        $notify->spool('mri upload processing class', $msg, 0, 'MRI.pm', $uploadID, 'N', 'Y');
+
+        return 'unknown' unless $sth->rows>0;
```
cmadjar: not sure I understand the `unless` part of the return here. Note: when more than one row matches, we get a message saying the scan type will be set to 'unknown' however, when it is being set to the actual scan type at the return of that function. So it never really returns 'unknown'

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -381,43 +354,93 @@ sub identify_scan_db {
...
+	}
+    # If more than one line in the mri_protocol_group_target matches the ProjectID/SubprojectID/VisitLabel
+    # then table mri_protocol_group_target was not setup properly. Warn.
+    my %mriProtocolGroupIDs = map { $_->{'MriProtocolGroupID'} => 1 } @rows;
+    if(keys %mriProtocolGroupIDs > 1) {
+		my $msg = "Warning! More than one protocol group can be used to identify the scan type of $minc_location\n"
+		        . "Ambiguous setup of table mri_protocol_group_target: setting scan type to 'unknown'"; 
+		print "$msg\n";
```
cmadjar: alignment is off

uploadNeuroDB/NeuroDB/MRI.pm
cmadjar: note if that helps, the two protocol matching that scan type are:
- Default protocol where everything is set to null
- Protocol restricted on the ProjectID that is the same as the image session

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -112,198 +112,163 @@ sub getScannerCandID {
 
 =pod
 
-=head3 getSessionID($subjectIDref, $studyDate, $dbhr, $objective, $db)
+=head3 getSession($subjectIDref, $studyDate, $dbh, $db)
```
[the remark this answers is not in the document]
nicolasbrossard: Done

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -112,198 +112,163 @@ sub getScannerCandID {
...
-    my $dbh = $$dbhr;
 
     # find a matching timepoint
-    $query = "SELECT ID, Date_visit, Visit "
+    $query = "SELECT ID, ProjectID, SubprojectID, CandID, Visit_label "
            . "FROM session "
-           . "WHERE CandID=$subjectIDref->{'CandID'} "
-           . "AND LOWER(Visit_label)=LOWER(".$dbh->quote($subjectIDref->{'visitLabel'}).") "
+           . "WHERE CandID=? "
+           . "AND LOWER(Visit_label)=LOWER(?) "
```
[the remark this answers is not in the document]
nicolasbrossard: Thx!

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -112,198 +112,163 @@ sub getScannerCandID {
...
+		    $subjectIDref->{'CandID'}
+		);
+		return (undef, $msg);
+	}
+	
+    # Since we'll be creating a visit, ensure that the subprojectID and ProjectID
+    # have been defined in the profile file
+    if (!defined $subjectIDref->{'SubprojectID'}) {
+        return (undef, "Cannot create visit: profile file does not define the visit's SubprojectID");
+	}
```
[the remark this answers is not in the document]
nicolasbrossard: Corrected

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -112,198 +112,163 @@ sub getScannerCandID {
...
+    # according to project_subproject_rel. This also validates the existence
+    # of ProjectID and SubprojectID in tables Project and subproject respectively.
+    $query = "SELECT ProjectID, SubprojectID "
+           . "FROM project_subproject_rel "
+           . "WHERE ProjectID=? "
+           . "AND SubprojectID=?";
+    $sth = $dbh->prepare($query);
+    $sth->execute($subjectIDref->{'ProjectID'}, $subjectIDref->{'SubprojectID'});
+
+    # If thre's no entry in project_subproject_rel for (ProjectID, SubprojectID)
```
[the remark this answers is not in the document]
nicolasbrossard: OK

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -112,198 +112,163 @@ sub getScannerCandID {
...
-        # in the profile file
-        return undef if !defined $subjectIDref->{'SubprojectID'};
-		
-        # determine the visit number and centerID for the next session
-        my $newVisitNo = 0;
-        my $centerID = 0;
+    # Ensure relationship between ProjectID and SubprojectID is legit
+    # according to project_subproject_rel. This also validates the existence
+    # of ProjectID and SubprojectID in tables Project and subproject respectively.
+    $query = "SELECT ProjectID, SubprojectID "
```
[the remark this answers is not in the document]
nicolasbrossard: ok

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -112,198 +112,163 @@ sub getScannerCandID {
...
-        }
-        # fixme ask Jon ... is this still useful?
-        # determine the centerID and new visit number (which is now deprecated) if getPSC() failed.
-        if($centerID == 0) {
-            $query = "SELECT IFNULL(MAX(VisitNo), 0)+1 AS newVisitNo, CenterID FROM session WHERE CandID=".$dbh->quote($subjectIDref->{'CandID'})." GROUP BY CandID, CenterID";
+    # determine the visit number and centerID for the next session
+    my $newVisitNo = 0;
+    my $centerID = 0;
+    
+    if($subjectIDref->{'visitLabel'} =~ /PHA/i or $subjectIDref->{'visitLabel'} =~ /TEST/i) {
```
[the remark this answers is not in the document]
nicolasbrossard: Much better! Done.

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -112,198 +112,163 @@ sub getScannerCandID {
...
+        # calibration data (PHANTOM_site_date | LIVING_PHANTOM_site_date | *test*)
+        my @pscInfo = getPSC($subjectIDref->{'visitLabel'}, $dbhr, $db);
+        $centerID = $pscInfo[1];
+    }
+    # fixme ask Jon ... is this still useful?
```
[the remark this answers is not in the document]
nicolasbrossard: I did.

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -112,198 +112,163 @@ sub getScannerCandID {
...
-
-    return $results[0];
+    
+    $newVisitNo = 1 unless $newVisitNo;
+    $centerID = 0 unless $centerID;
+
+    # Insert the new session setting Current_stage to 'Not started' because that column is important 
+    # to the behavioural data entry gui.
+    $query = "INSERT INTO session "
+           . "SET CandID=".$dbh->quote($subjectIDref->{'CandID'}).", "
```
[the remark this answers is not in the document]
nicolasbrossard: Fixed.

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -112,198 +112,163 @@ sub getScannerCandID {
...
+           . "    Visit_label=".$dbh->quote($subjectIDref->{'visitLabel'}).", "
```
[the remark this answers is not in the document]
nicolasbrossard: Fixed.

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -112,198 +112,163 @@ sub getScannerCandID {
...
+           . "    CenterID=$centerID, "
```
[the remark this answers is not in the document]
nicolasbrossard: Fixed.

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -112,198 +112,163 @@ sub getScannerCandID {
...
+           . "    VisitNo=$newVisitNo, "
```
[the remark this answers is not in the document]
nicolasbrossard: Fixed.

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -112,198 +112,163 @@ sub getScannerCandID {
...
+           . "    Current_stage='Not Started', "
+           . "    Scan_done='Y', "
+           . "    Submitted='N', "
+           . "    SubprojectID=".$dbh->quote($subjectIDref->{'SubprojectID'}) . ", "
```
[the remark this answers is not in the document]
nicolasbrossard: Fixed.

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -112,198 +112,163 @@ sub getScannerCandID {
...
+           . "    ProjectID=".$dbh->quote($subjectIDref->{'ProjectID'});
```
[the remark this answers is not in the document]
nicolasbrossard: Fixed.

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -112,198 +112,163 @@ sub getScannerCandID {
...
+    $dbh->do($query); # execute query
```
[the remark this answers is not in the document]
nicolasbrossard: Done.

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -112,198 +112,163 @@ sub getScannerCandID {
...
+    $sessionID = $dbh->{'mysql_insertid'}; # retain id of inserted row
+    $subjectIDref->{'visitNo'} = $newVisitNo; # add visit number to subjectIDref
```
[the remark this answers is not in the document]
nicolasbrossard: Not used, so I removed it.

uploadNeuroDB/minc_insertion.pl
nicolasbrossard: Done.

uploadNeuroDB/imaging_non_minc_insertion.pl
nicolasbrossard: The script does not use `ProjectID` and `SubprojectID` so no need to copy them back into `$subjectIDref`.

uploadNeuroDB/NeuroDB/MRI.pm
nicolasbrossard: Fixed alignment. The `unless` is useless: removed.

uploadNeuroDB/NeuroDB/MRI.pm
nicolasbrossard: The `unless` is responsible for the erroneous behaviour you observed. I removed it.

uploadNeuroDB/NeuroDB/MRI.pm
nicolasbrossard: Fixed.

uploadNeuroDB/minc_insertion.pl
```diff
@@ -534,15 +534,32 @@ =head2 Methods
...
+    $subjectIDsref, 
+    $studyInfo{'DateAcquired'},
+    $dbh,
+    $db
+);
 
+# Copy the session info into the %$subjectIDsref hash array
+$subjectIDsref->{'SessionID'}    = $sessionRef->{'ID'};
+$subjectIDsref->{'ProjectID'}    = $sessionRef->{'ProjectID'};
+$subjectIDsref->{'SubprojectID'} = $sessionRef->{'ProjectID'};
```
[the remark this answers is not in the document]
nicolasbrossard: Aww geez.....Can't believe I did this...Fixed.

PR #370, opened 2019-01-16
[Bug fix] Insert into parameter_type_category_rel table when creating new parameter_type

cmadjar:
### Description

This fixes an issue raised in https://github.com/aces/Loris/pull/3695#discussion_r242642123 where entries into the `parameter_type` table created by the imaging script were not linked to a parameter type category via `parameter_type_category_rel`. The code in this PR fixes this issue.

The imaging part if corrected in the `File.pm` file and the electrophysiology part is corrected in `physiological.py` library. 

### Resolves

- [x] issue raised in @davidblader publication module PR (https://github.com/aces/Loris/pull/3695#discussion_r242642123)

### Caveat for Existing Projects

Make sure to source the SQL patch in the following LORIS PR https://github.com/aces/Loris/pull/4263 to fix existing projects' table.

### DIFF
--- a/python/lib/physiological.py
+++ b/python/lib/physiological.py
@@ -229,8 +229,7 @@ class Physiological:
     def get_parameter_type_category_id(self):
         """
         Greps ParameterTypeCategoryID from parameter_type_category table.
-        If no ParameterTypeCategoryID were found, the script will create it in
-        parameter_type_category.
+        If no ParameterTypeCategoryID was found, it will return None.
 
         :return: ParameterTypeCategoryID
          :rtype: int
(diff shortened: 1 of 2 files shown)


python/lib/physiological.py
```diff
@@ -215,8 +215,46 @@ def get_parameter_type_id(self, parameter_name):
...
+        category_result = self.db.pselect(
+            query='SELECT ParameterTypeCategoryID '
+                  'FROM parameter_type_category '
+                  'WHERE Name = %s ',
+            args=('Electrophysiology Variables',)
+        )
+        if category_result:
+            category_id = category_result[0]['ParameterTypeCategoryID']
+        else:
+            # if no results, create an entry in parameter_type_category
```
PapillonMcGill: I believe that it would be better to insert that value in a one time script executed during the update of LORIS than here. Even if you have to create a related PR on LORIS.

This way, no check would be necessary except maybe to raise and exception/error.

cmadjar: @PapillonMcGill I removed the code creating that category here. I already sent the SQL patch to create the category on the LORIS side so all should be good.

Thanks!

Issue #380, opened 2019-01-30
[major] fix BIDS insertion script and replace Gender by Sex


cmadjar:
Once 20.2 is merged into major, need to fix the BIDS import script and rename `candidate` table field `Gender` to `Sex` as this field was renamed in major. The code changes on the MRI side was all done to account for the change of field name in the database

cmadjar:
PR was merged so closing the issue

PR #336, opened 2018-08-27
Replace the "Gender" field with "Sex" in the candidate and tarchive tables 

cmadjar:
### Description
This pull request goes with the LORIS changes to the renaming of the following fields:
- `tarchive` table: `PatientGender` renamed to `PatientSex`
- `candidate` table: `Gender` renamed to `Sex`

### How to test this PR
- make sure to source the patch from the LORIS PR that changes the field names
- try inserting a scan that will create a candidate and insert into the tarchive table

### See also
PR on the LORIS side: https://github.com/aces/Loris/pull/3897/files

PR #382, opened 2019-02-04
Replace Gender by Sex in bids insertion scripts

cmadjar:
### Description

This PR fixes the BIDS import script and rename the `candidate` table field `Gender` to `Sex` as this field was renamed in major and the BIDS import scripts were part of the latest minor release (20.2.0). 

Note that the code changes on the MRI side was already done to account for the change of field name in the database (see #336).

### This fixes

[x] issue #380


cmadjar:
@PapillonMcGill I figured you could probably review this PR when you get a chance.

PR #395, opened 2019-02-07
Fixed the pipeline for when there is no date in the DICOM files

cmadjar:
### Description

In the context of open science, no dates are available with the dataset therefore, the following needed to be fixed:
- the DICOM archives cannot be moved to a year subfolder when running `tarchiveLoader`. If no acquisition date is available, the DICOM archive will remain in the `tarchive` directory and will not be moved to a subdirectory. If the acquisition date is present in the DICOM, then the behaviour remains the same.
- the call to `Mincinfo_wrapper` in the function `get_mincs` of `MRIProcessingUtility.pm` includes the `-date` option and then sorts based on the date. However, sorting the files based on the date does not really sort the files since they are all acquired on the same date for most of them. Therefore, if we wish to sort the files based on their order of acquisition, we should instead call `Mincinfo_wrapper` with the option `-attvalue acquisition:acquisition_id` which returns the series number which is more informative on the order of acquisition. 
- If no dates are available in the DICOM dataset (a.k.a. for open science), the DICOM archives are called `DCM__Imaging...` and the `tarchive_validation.pl` script fails as it expects the format to be `DCM_date_Imaging...`. This PR fixes the regular expression so that the date is optional when parsing the file.
- several warnings when attempting to concatenate the date with a string also needed to be fixed for when there is no dates in the dataset

### This fixes

all issues found when testing the pipeline on a fully deidentified dataset (including the removal of the acquisition dates). 

Closing #392, #393 and #394 since they are all solved in this PR and one could not test the pipeline when removing the acquisition date from the DICOM dataset fully otherwise.

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
+++ b/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
@@ -380,8 +380,7 @@ sub determineSubjectID {
             $subjectIDsref->{'CandID'},
             $subjectIDsref->{'PSCID'},
             $subjectIDsref->{'visitLabel'},
-            (defined $tarchiveInfo->{'DateAcquired'}
-                ? $tarchiveInfo->{'DateAcquired'} : 'UNKNOWN')
+            $tarchiveInfo->{'DateAcquired'} // 'UNKNOWN'
         );
 	$this->{LOG}->print($message);
         $this->spool($message, 'N', $upload_id, $notify_detailed);
--- a/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
+++ b/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
@@ -324,7 +324,7 @@ sub extractAndParseTarchive {
         $this->extract_tarchive($tarchive, $upload_id, $seriesuid);
     my $ExtractSuffix  = basename($tarchive, ".tar");
     # get rid of the tarchive Prefix 
-    $ExtractSuffix =~ s/DCM_(\d){4}-(\d){2}-(\d){2}_//;
+    $ExtractSuffix =~ s/DCM_(\d{4}-\d{2}-\d{2})?_//;
     my $info       = "head -n 12 $this->{TmpDir}/${ExtractSuffix}.meta";
     my $header     = `$info`;
     my $message = "\n$header\n";
@@ -375,11 +375,14 @@ sub determineSubjectID {
                             $this->{dbhr}
                         );
     if ($to_log) {
-        my $message = "\n==> Data found for candidate   : ".
-                            "CandID: ". $subjectIDsref->{'CandID'} .
-                            "- PSCID: ". $subjectIDsref->{'PSCID'} . "- Visit: ".
-                            $subjectIDsref->{'visitLabel'} . "- Acquired : ".
-                            $tarchiveInfo->{'DateAcquired'} . "\n";
+        my $message = sprintf(
+            "\n==> Data found for candidate CandID: %s, PSCID %s, Visit %s, Acquisition Date %s\n ",
+            $subjectIDsref->{'CandID'},
+            $subjectIDsref->{'PSCID'},
+            $subjectIDsref->{'visitLabel'},
+            (defined $tarchiveInfo->{'DateAcquired'}
+                ? $tarchiveInfo->{'DateAcquired'} : 'UNKNOWN')
+        );
 	$this->{LOG}->print($message);
         $this->spool($message, 'N', $upload_id, $notify_detailed);
     }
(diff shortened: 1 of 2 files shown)


cmadjar:
Was tested on datasets with and without dates.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -375,11 +375,14 @@ sub determineSubjectID {
...
-        my $message = "\n==> Data found for candidate   : ".
-                            "CandID: ". $subjectIDsref->{'CandID'} .
-                            "- PSCID: ". $subjectIDsref->{'PSCID'} . "- Visit: ".
-                            $subjectIDsref->{'visitLabel'} . "- Acquired : ".
-                            $tarchiveInfo->{'DateAcquired'} . "\n";
+        my $message = sprintf(
+            "\n==> Data found for candidate CandID: %s, PSCID %s, Visit %s, Acquisition Date %s\n ",
+            $subjectIDsref->{'CandID'},
+            $subjectIDsref->{'PSCID'},
+            $subjectIDsref->{'visitLabel'},
```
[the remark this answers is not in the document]
cmadjar: done

uploadNeuroDB/tarchiveLoader
```diff
@@ -633,12 +633,14 @@ if ($valid_study) {
...
-	    "\n" . $subjectIDsref->{'CandID'} . 
-            " " . $subjectIDsref->{'PSCID'} .
-            " " . $subjectIDsref->{'visitLabel'} .
-            "\tacquired ". $tarchiveInfo{'DateAcquired'} .
-	    "\n";
+    $message = sprintf(
+        "\n CandID: %s, PSCID: %s, Visit: %s, Acquisition Date: %s\n",
+        $subjectIDsref->{'CandID'},
+        $subjectIDsref->{'PSCID'},
+        $subjectIDsref->{'visitLabel'},
```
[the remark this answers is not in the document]
cmadjar: done

uploadNeuroDB/tarchiveLoader
```diff
@@ -665,11 +667,13 @@ if ($valid_study) {
...
     ## instead of using patientName ############################
     ############################################################
-    $message = 
-	    "\n" . $tarchive. " acquired ". 
-            $tarchiveInfo{'DateAcquired'} .
-            " was deemed invalid\n\n". $study_dir .
-	    "\n";
+    $message = sprintf(
+        "\n %s acquired %s was deemed invalid\n\n%s\n",
+        $tarchive,
```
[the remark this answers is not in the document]
cmadjar: done

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1277,7 +1284,7 @@ sub get_mincs {
     my @files_list;
     foreach my $file (@files) {
         next unless $file =~ /\.mnc(\.gz)?$/;
-        my $cmd= "Mincinfo_wrapper -quiet -tab -file -date $this->{TmpDir}/$file";
+        my $cmd= "Mincinfo_wrapper -quiet -tab -file -attvalue acquisition:acquisition_id $this->{TmpDir}/$file";
```
nicolasbrossard: As discussed, we should use `quotemeta` on all arguments that might contain shell special characters to avoid bad surprises.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -909,26 +911,34 @@ sub update_mri_acquisition_dates {
...
-                " WHERE s.ID='$sessionID' AND s.Active='Y' AND".
-                " (m.AcquisitionDate > '$acq_date'". 
-                " OR m.AcquisitionDate IS NULL) AND '$acq_date'>0";
-    
+    my $query = "SELECT s.ID, m.AcquisitionDate "
+                . " FROM session AS s "
+                . " LEFT OUTER JOIN mri_acquisition_dates AS m ON (s.ID=m.SessionID)"
+                . " WHERE s.ID=? AND s.Active='Y'";
+    my @values = ($sessionID);
+    if ($acq_date) {
```
nicolasbrossard: I think this part of the query should be revisited/simplified to take into account all cases (`$acq_date` defined, `$acq_date` defined but set to `''` or `'0000-00-00'`, `$acq_date` not defined)

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
cmadjar: done in the next commit

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
cmadjar: so I modified the code so that if the date is `'0000-00-00'`, `''` or `0`, `$acq_date` is set to `undef` because if the date is set to one of these values, we don't want to update the table with those values.

nicolasbrossard:
LGTM

Issue #383, opened 2019-02-04
Use UploadID as a key to update mri_upload when running tarchiveLoader


cmadjar:
Review all the queries that updates the `mri_upload` table and make sure that the fields in that table are updated using the `UploadID` in the `WHERE` clause instead of the `TarchiveID` as more than one upload could be associated to a `TarchiveID` and we only want to update the upload with the current `UploadID`. 

See this presentation for details on what problems using `TarchiveID` to update the `mri_upload` table could do: https://docs.google.com/presentation/d/1BpAvGcduTESUcDNrSyZOiz8sszbweJ7H_slKZfM-6so/edit#slide=id.g4e9a10fecf_0_0

cmadjar:
PR was merged so closing the issue.

PR #397, opened 2019-02-08
Use UploadID to update mri_upload instead of TarchiveID

cmadjar:
### Description

Modifies the update and select queries on `mri_upload` to use the `UploadID` instead of the `TarchiveID` to avoid updating multiple entries in the `mri_upload` table when more than one `UploadID` refers to the same `TarchiveID`.

### This fixes

- [x] #383

### DIFF
--- a/uploadNeuroDB/tarchive_validation.pl
+++ b/uploadNeuroDB/tarchive_validation.pl
@@ -15,6 +15,8 @@ Available options are:
 
 -profile     : name of the config file in C<../dicom-archive/.loris-mri>
 
+-uploadID    : UploadID associated to the DICOM archive to validate
+
 -reckless    : upload data to the database even if the study protocol
                is not defined or if it is violated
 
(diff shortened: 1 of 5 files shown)


uploadNeuroDB/tarchive_validation.pl
```diff
@@ -15,6 +15,8 @@ =head1 SYNOPSIS
 
 -profile     : name of the config file in C<../dicom-archive/.loris-mri>
 
+-uploadID    : UploadID associated to the DICOM archive to validate
```
PapillonMcGill: Is there a reason to have 3 different texts for the uploadID  param in theses 3 files?

Apart that, it look good to me. Note that I'am no expert with LORIS-MRI

cmadjar: @PapillonMcGill It is slightly different because each script using that option is at a different step in the pipeline (generic upload, DICOM archive validation or MINC insertion all derived from the same upload). Hope this makes more sense now.

PapillonMcGill: @cmadjar Yes it does. 
Thanks!

Issue #377, opened 2019-01-28
"SliceTiming" values are in milliseconds instead of seconds


PierreEMorin:
I got this mail from Hanad Sharmarke (Pierre Bellec lab).  Briefly, Hanad put the previous BIDS release through a BIDS validator and it failed because the "SliceTiming" values are in milliseconds instead of seconds.

> Hello,
> 
> Regarding the bids release that was done in the summer, I put it through a bids validator using this website. I found that the slice timing correction is not in seconds.
> 
> https://bids-standard.github.io/bids-validator/
> 
> Hanad
> 
> image.png
![image](https://user-images.githubusercontent.com/6494031/51861357-f85c7f00-2309-11e9-871e-a033344552dd.png)

cmadjar:
PR merged.

PR #398, opened 2019-02-08
Convert slice timing to seconds when creating the BIDS MRI structure

cmadjar:
### Description

The MRI BIDS dataset created using the tool `MakeNiiFilesBIDSCompliant.pl ` fails the BIDS validator due to the SliceTiming field being export in milliseconds instead of seconds. This PR converts the BIDS SliceTiming field into seconds so that it can pass the BIDS validator.

### This fixes

- [x] #377


cmadjar:
@PierreEMorin I assigned this PR to you so you can review and test it since you brought up that bug.

PR #409, opened 2019-03-08
Circumvented bug that made tar fail on NFS file systems in get_dicom_script.pl and added option -id

nicolasbrossard:
This PR provides a fix for the following use case:

Let's say a user wants to extract a very large set of C<DICOM> files. Since the extraction algorithm extracts all the wanted files in a temporary directory and then archives the entire set in a C<.tar.gz> file, the temporary directory has to be able to (temporarily) store a **huge** amount of a data. Now suppose that on the user's system, the temporary directory can only be on a NFS file system because of space issues. The script will fail since C<tar> has been known to incorrectly report that some files changed during the archiving process if the files are located on an NFS file system, stopping the script dead in its tracks.

The solution is to extract small portions of the data set at a time in the temporary directory, archive them and then immediately delete them from the temporary directory. This reduces the free space requirement imposed on this directory, which in turn provides a fix for the problem described above, assuming that the user has access to a non-NFS temporary directory with sufficient free space.


tools/get_dicom_files.pl
```diff
@@ -271,36 +279,41 @@ =head2 Methods
         # --absolute-path: since we are extracting in $outDir and since
         #                  $outDir is an absolute path, we need this option otherwise
         #                  tar will refuse to extract
-        my $cmd = "tar zxf $tmpExtractDir/$innerTar"
-                . " --files-from=$fileList "   
-                . " --absolute-names "
-                . " --transform='s#^.*/#$outDir/#'";
+        $cmd = "tar zxf $tmpExtractDir/$innerTar"
```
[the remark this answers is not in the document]
nicolasbrossard: Done.

docs/scripts_md/get_dicom_files.md
```diff
@@ -22,16 +22,20 @@ Available options are:
...
+           `/data/tmp/get_dicom_files.pl.n1d4`. By default, dir\_argument is set to the value of
+           the environment variable `TMPDIR`. Since the UNIX program `tar` has known limitations 
+           with NFS file systems (incorrect reports of files that changed while they are archived), the
+           argument to `-d` should not be a directory that resides on an NFS mounted file system.
+           Failure to do so might result in `get_dicom_files.pl` failing.
 
 \-o       : basename of the final `tar.gz` file to produce, in the current directory (defaults to 
            `dicoms.tar.gz`).
 
+\-p       : when storing a DICOM file in the archive, use the PSCID of the candidate as one of the
```
cmadjar: Random question, wouldn't be easier to include both PSCID and DCCID by default? Or do you think that usually users would only want one or the other? I don't have a strong opinion on this. Was just wondering

tools/get_dicom_files.pl
```diff
@@ -90,7 +95,9 @@ =head2 Methods
     ["-d", "string", 1,   \$tmpExtractBaseDir, 
         "base path of the temporary directory where files are extracted"],   
     ["-o", "string", 1,   \$outTarBasename, 
-        "base path of the final .tar.gz file to produce (defaults to 'dicoms')"],   
+        "base path of the final .tar.gz file to produce (defaults to 'dicoms')"], 
+    ["-p", "const" , 1,   \$usePSCID,
```
cmadjar: should it be 'const' or 'bool' here? $usePSCID seems like a boolean to me but I am not sure of the difference between 'const' and 'bool' here.

tools/get_dicom_files.pl
```diff
@@ -90,7 +95,9 @@ =head2 Methods
     ["-d", "string", 1,   \$tmpExtractBaseDir, 
         "base path of the temporary directory where files are extracted"],   
     ["-o", "string", 1,   \$outTarBasename, 
-        "base path of the final .tar.gz file to produce (defaults to 'dicoms')"],   
+        "base path of the final .tar.gz file to produce (defaults to 'dicoms')"], 
+    ["-p", "const" , 1,   \$usePSCID, 
+        "include the PSCID in the components of the archived DICOM file paths"]
```
cmadjar: components? Maybe replace by filenames instead?

cmadjar: not entirely sure what you mean by components here

docs/scripts_md/get_dicom_files.md
```diff
@@ -22,16 +22,20 @@ Available options are:
            the scan type name is not used to determine which DICOM files to extract. This option
            must be used if no patient name patterns were specified via `-n` (see above).
 
-\-d       : extract the files in directory `<dir_argument>/get_dicom_files.pl.<UNIX_process_number>`
-           For example with `-d /data/tmp`, the DICOM files will be extracted in 
-           `/data/tmp/get_dicom_files.pl.67888` (assuming 67888 is the process number). 
-           By default, dir\_argument is set to the value of the environment variable `TMPDIR`. Note 
-           that the directory argument has to exist and be writeable, otherwise the script will issue
-           an error and abort.
+\-d       : extract the files in directory `<dir_argument>/get_dicom_files.pl.<random_string>`
```
cmadjar: is it possible to have long options instead of just letters like all the other scripts in LORIS-MRI? (a.k.a. -profile, -pscid -directory....)

nicolasbrossard: For compatibility, sure. I'll correct.

docs/scripts_md/get_dicom_files.md
nicolasbrossard: I now provide more than one way of naming the subdirectory (see option `-id`).

tools/get_dicom_files.pl
nicolasbrossard: Option `-id` now takes a string argument.

tools/get_dicom_files.pl
nicolasbrossard: Updated the description.

docs/scripts_md/get_dicom_files.md
```diff
@@ -4,54 +4,53 @@ get\_dicom\_files.pl - extracts DICOM files for specific patient names/scan type
...
            Failure to do so might result in `get_dicom_files.pl` failing.
 
-\-o       : basename of the final `tar.gz` file to produce, in the current directory (defaults to 
+\-outfile : basename of the final `tar.gz` file to produce, in the current directory (defaults to 
            `dicoms.tar.gz`).
 
-\-p       : when storing a DICOM file in the archive, use the PSCID of the candidate as one of the 
-           components of the file path (default is to use the DCCID).
+\-id      : how to name the subdirectory identifying the candidate to which the DICOM files belong:
+           pscid, candid, pscid\_candid or candid\_pscid (defaults to candid)
```
cmadjar: perfecto!!

docs/scripts_md/get_dicom_files.md
cmadjar: Thanks!

PR #417, opened 2019-04-12
fix batch_uploads_imageuploader that inserted duplicated uploads

cmadjar:
### Description

`batch_uploads_imageuploader` creates a duplicate entry in the `mri_upload` table when loading one upload:
```
+----------+-------------+---------------------+---------------------------------------------------------------------------------------+--------------------------------------------------------------------------------------+-------------------+------------------------+-----------------------+------------+-----------+--------------------------+---------------------+-----------+------------------------------------------+-----------+
| UploadID | UploadedBy  | UploadDate          | UploadLocation                                                                        | DecompressedLocation                                                                 | InsertionComplete | number_of_mincInserted | number_of_mincCreated | TarchiveID | SessionID | IsCandidateInfoValidated | IsTarchiveValidated | Inserting | PatientName                              | IsPhantom |
+----------+-------------+---------------------+---------------------------------------------------------------------------------------+--------------------------------------------------------------------------------------+-------------------+------------------------+-----------------------+------------+-----------+--------------------------+---------------------+-----------+------------------------------------------+-----------+
|      166 | cecile      | 2019-04-12 08:52:59 | /data/incoming/preventAD/MTL0078_857385_NAPEN00_20121204_104302249.tgz                | /data/not_backed_up/ImagingUpload-9-5-iq6JYs                                         |                 1 |                     13 |                    15 |    4002683 |      2895 |                        1 |                   1 |         0 | MTL0078_857385_NAPEN00                   | N         |
|      167 | cecile      | 2019-04-12 09:05:28 | /data/incoming/preventAD/MTL0078_857385_NAPEN00_20121204_104302249.tgz                |                                                                                      |                 0 |                   NULL |                  NULL |       NULL |      NULL |                     NULL |                   0 |      NULL | MTL0078_857385_NAPEN00                   | N         |
```

This PR fixes that bug in `batch_uploads_imageuploader`

### DIFF
--- a/batch_uploads_imageuploader
+++ b/batch_uploads_imageuploader
@@ -169,7 +169,6 @@ while(my $line = <STDIN>)
 {
     chomp $line;
     my @linearray = split(" " , $line);
-    push (@resultsarray, @linearray);
     @fullpath = $linearray[0];
     @phantom = $linearray[1];
     @patientname = $linearray[2];
@@ -180,7 +179,7 @@ while(my $line = <STDIN>)
 close STDIN;
 
 ## foreach series, batch magic
-foreach my $input (@resultsarray)
+foreach my $input (@fullpath)
 {
     $counter++;
     $stdout = $stdoutbase.$counter;
--- a/batch_uploads_imageuploader
+++ b/batch_uploads_imageuploader
@@ -57,9 +57,6 @@ use NeuroDB::ExitCodes;
 
 my $profile   = '';
 my $upload_id = undef; 
-my $fullpath = undef; 
-my $phantom = undef; 
-my $patientname = undef; 
 my ($debug, $verbose) = (0,1);
 my $stdout = '';
 my $stderr = '';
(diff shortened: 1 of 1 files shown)


batch_uploads_imageuploader
```diff
@@ -169,7 +169,6 @@ while(my $line = <STDIN>)
 {
     chomp $line;
     my @linearray = split(" " , $line);
-    push (@resultsarray, @linearray);
     @fullpath = $linearray[0];
```
cmadjar: Here, if only one line is given in STDIN, `@resultsarray` becomes:
0 'full path'
1 'is phantom'
2 'patient name'

hence, the `@resultsarray` length is 3 instead of 1

nicolasbrossard: Not sure I understand....Since `@fullPath` is initialized every time the `while` loop is executed, when the loop terminates, `@fullpath` will have the value it had in the last iteration of the loop, which means all other iterations will have no impact whatsoever on its value...This is going to work __only__ if one line is fed to `STDIN`.

cmadjar: oops, I should have used @fullpatharray for the looping below, not @fullpath...

PR #424, opened 2019-05-23
Remove deprecated function isValidMRIProtocol now that the mri_protocol contains MIN MAX values

cmadjar:
This PR removes the function that was added in `MRIProcessingUtility.pm` when the `mri_protocol` table was modified to remove the `,` in the range columns. This function is not needed anymore now that the range columns have been split into `Min` `Max` columns.


cmadjar:
@nicolasbrossard I regenerated the MD file. Please re-review. Thanks!!!

mohadesz:
@cmadjar  Let me know if you need this tested by inserting a scan or is it all good?

PR #426, opened 2019-05-23
Fix in_range logic

cmadjar:
The `in_range` logic was not properly set and led to wrong identification of the MRI protocol and mislabelled scans... 

This PR fixes this bug found while testing 21.0.


uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -711,10 +711,6 @@ Determines whether numerical value falls within the range described by range
 string. Range string is a single range unit which follows the syntax
```
nicolasbrossard: Has the `.md` been regenerated?

cmadjar: @nicolasbrossard oops. I just regenerated it. Thanks!

PR #439, opened 2019-06-20
[install script] Add the missing DBI.pm library to the CPAN install

cmadjar:
The `DBI.pm` library is missing in the CPAN install instructions run by the install script. This PR adds it to the install script to correct this mistake.


cmadjar:
Just tested the install of those CPAN libraries on a VM with no CPAN libraries. All good!

PR #440, opened 2019-06-20
[mass_perdoc_md_creation.pl] updated the list of files with Moose files and .pl extensions

cmadjar:
Updated the `mass_perldoc_md_creation.pl` script to include the Moose files and also integrate the `.pl` extensions that were added to some scripts in release 21.0.


nicolasbrossard:
Here's what I noticed:

- Every `*.pm` file under directory `MNI` is missing. These modules are documented. Not sure why they are not in the mass doc generation script.
- `tools/MakeArchiveLocationRelative.pl` is listed twice.
- `tools/addSeriesAndFileRecords.pl` is missing but it's not documented anyway
- `tools/dcmconvTarchive.pl`  is missing but it's not documented anyway
- `uploadNeuroDB/bin/bestlinreg_claude.pl`, `uploadNeuroDB/bin/deface_minipipe.pl` and `uploadNeuroDB/bin/pipeline_qc_face.pl` are missing but they are not really **our** scripts anyway
- `tools/example_scripts/deletemincsqlwrapper.pl` is missing but this is an example script

Bottom line: it's probably OK, but please confirm.

cmadjar:
@nicolasbrossard I checked the MNI ones and they are not documented using perldoc (they are old libraries from the MINC tools I believe) so they probably should not be in the script to create the MD files (same as the other ones you mentioned in your comment)

I removed the duplicated entry for `MakeArchiveLocationRelative.pl` from the script.

Thanks!

PR #447, opened 2019-06-26
[BIDS EEG] Get age at scan from sub-XXX_scans.tsv file of the BIDS structure

cmadjar:
If the age at the time of acquisition can be parsed in the BIDS `sub-XXX_scans.tsv` file, then the `age_at_scan` will be inserted into the `physiological_parameter_file` table.

In addition, added the copy of the `sub-XXX_scans.tsv` file into the LORIS BIDS import directory.

Finally, refactored the way the `sub-XXX_scans.tsv` file is read so it can be extended to other modalities following the BIDS structure. A new library called `lib/scanstsv.py` was created to deal with that file.

This fixes the age import for the Cuban EEG dataset which will be using 21.0.

### DIFF
--- a/python/lib/eeg.py
+++ b/python/lib/eeg.py
@@ -5,7 +5,6 @@ import json
 import getpass
 import string
 from pyblake2 import blake2b
-from dateutil.parser import parse
 
 import lib.exitcode
 import lib.utilities as utilities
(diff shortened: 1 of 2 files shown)


cmadjar:
tested on several cuban datasets and on the face13 dataset. Works well for all EEG dataset we have so far.

PR #461, opened 2019-07-11
Simplification of the candidate validation

cmadjar:
### Description

There were a lot of redundancies in functions that were called to validate the candidate information and at the same time some differences. These sometimes lead to misleading messages about what went wrong with the insertion.

This PR harmonizes the way the candidate information is validated throughout all insertion scripts which also simplified the code.

#### Summary of the changes made:
- removed unused or duplicated functions related to validation of candidate information
- harmonized how to get the candidate information and how to perform validation on the candidate information into `MRIProcessingUtility::determineSubjectID`) 
- simplified and reorganized the logic of some functions involved in candidate creation or validation.
- error message more robust and clearer

#### Note: typical flow on how to determine and validate subject IDs (from refactored `MRIProcessingUtility::determineSubjectID` function):
1) get the subject IDs from the `PatientName` or `PatientID` into `$subjectIDsref`
2) create the candidate if `createCandidate` config setting is set to true and the candidate does not already exist (**New feature**: check that both the `PSCID` and `CandID` do not exists already before creating the candidate to make sure we do not create candidates with same `PSCID` (current bug))
3) validate the candidate information stored in `$subjectIDsref`
> - checks that the `CandID` exists in the database
> - checks that the `PSCID` exists in the database
> - checks that the `PSCID` and `CandID` refers to the same candidate in the database
> - checks that the `visitLabel` is valid (a.k.a. present in the `Visit_Windows` table) 

If the candidate information stored in `$subjectIDsref` fails the validation step, then `$subjectIDsref->{'CandMismatchError'}` will be set to the correct mismatch error.

### Caveat for existing projects

Remove this part from the `getSubjectIDs` function of the `prod` file since this validation is now automatically performed by the insertion pipeline:
```
# Note, this function will ensure that the PSCID/CandID information refers to a valid candidate
# and it will check that the Visit label exists in the Visit_Windows table
if(!NeuroDB::MRI::subjectIDIsValid($subjectID{'CandID'}, $subjectID{'PSCID'}, $subjectID{'visitLabel'}, $dbhr)) {
        return undef;
}
```


dicom-archive/profileTemplate.pl
```diff
@@ -45,29 +45,24 @@ sub getSubjectIDs {
...
-        $subjectID{'CandID'}        = NeuroDB::MRI::my_trim($2);
-        $subjectID{'visitLabel'}    = NeuroDB::MRI::my_trim($3);
-        $subjectID{'createVisitLabel'} = 0; 
+        $subjectID{'PSCID'}      = NeuroDB::MRI::my_trim($1);
+        $subjectID{'CandID'}     = NeuroDB::MRI::my_trim($2);
+        $subjectID{'visitLabel'} = NeuroDB::MRI::my_trim($3);
 
-        # Note, this function will ensure that the PSCID/CandID information refers to a valid candidate
-        # and it will check that the Visit label exists in the Visit_Windows table
-        if(!NeuroDB::MRI::subjectIDIsValid($subjectID{'CandID'}, $subjectID{'PSCID'}, $subjectID{'visitLabel'}, $dbhr)) {
```
cmadjar: removed this as removed the function subjectIDIsValid from `MRI.pm` as it has been refactored to use the functions in `MRIProcessingUtilities.pm` only. There was duplication of validation (one done in `MRI.pm` and the other done in `MRIProcessingUtilities`)

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -62,86 +62,6 @@ $FLOAT_EQUALS_NB_DECIMALS = 4;
...
-  - $scannerID  : scanner ID
-  - $dbhr       : database handle reference
-  - $db         : database object
-
-RETURNS: a reference to a hash containing elements including C<CandID>,
-C<visitLabel> and C<visitNo>, or, in the case of failure, C<undef>
-
-=cut
-
-sub getSubjectIDs {
```
cmadjar: removed this function as we are only using the ones from the prod file, hence, we only have to maintain one instead of two

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -62,86 +62,6 @@ $FLOAT_EQUALS_NB_DECIMALS = 4;
...
-  - $pscid             : candidate's C<PSCID>
-  - $visit_label       : visit label
-  - $dbhr              : the database handle reference
-  - $create_visit_label: boolean, if true, will create the visit label
-
-RETURNS: 1 if the ID pair matches, 0 otherwise
-
-=cut
-
-sub subjectIDIsValid {
```
cmadjar: this was also done in `candidateValidation` of `MRIProcessingUtilities.pm` which was more complete than this validation. Note that `candidateValidation` is now called automatically by `determineSubjectID` to validate the candidate information right after having guessed the PSCID, DCCID, Visit label with `Settings::getSubjectIDs`

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -1491,32 +1411,6 @@ sub DICOMDateToUnixTimestamp {
     }
 }
 
-=pod
-
-=head3 lookupCandIDFromPSCID($pscid, $dbhr)
```
cmadjar: This is not used anywhere so removed the function for clean up.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1770,72 +1785,84 @@ or a phantom
...
-    my ($subjectIDsref)= @_;
+    my ($subjectIDsref) = @_;
+
     my $CandMismatchError = undef;
+
+
+    ############################################################
+    ################ No Checking if the subject is Phantom #####
+    ############################################################
+    if ($subjectIDsref->{'isPhantom'}) {
```
cmadjar: moved this up in the function as there is not need to do anything else if it is a phantom.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1770,72 +1785,84 @@ or a phantom
...
+
     if (($sth->rows == 0) && (!$subjectIDsref->{'createVisitLabel'})) {
         print LOG "\n=> No Visit label\n";
-        $CandMismatchError= 'Visit label does not exist';
+        $CandMismatchError = "Visit label $visit_label does not exist in Visit_Windows";
+
         return $CandMismatchError;
     } elsif (($sth->rows == 0) && ($subjectIDsref->{'createVisitLabel'})) {
-        print LOG "\n=> Will create visit label $subjectIDsref->{'visitLabel'}\n";
+        print LOG "\n=> Will create visit label $visit_label in Visit_Windows\n";
```
cmadjar: this is a more appropriate log error as it always get confusing as to why the visit label is deemed invalid. It is invalid only if it is not part of the Visit_Windows table.

uploadNeuroDB/imaging_non_minc_insertion.pl
```diff
@@ -408,9 +417,7 @@ =head2 Methods
 }
 
 # check whether there is a candidate IDs mismatch error
-my $CandMismatchError;
-$CandMismatchError = $utility->validateCandidate($subjectIDsref);
```
cmadjar: removed this as it is now part of the `determineSubjectIDs` function and is returned into `$subjectIDsref->{'CandMismatchError'}`

uploadNeuroDB/minc_insertion.pl
```diff
@@ -476,17 +477,17 @@ =head2 Methods
...
-    $scannerID, \%studyInfo, 0, $upload_id
+    $scannerID, \%studyInfo, 0, $upload_id, $User, $centerID
 );
 
 
 
 
 
 ## Validate that the candidate exists and that PSCID matches CandID
-my $CandMismatchError = $utility->validateCandidate($subjectIDsref);
```
cmadjar: removed this as it is now part of the determineSubjectIDs function and is returned into $subjectIDsref->{'CandMismatchError'}

uploadNeuroDB/tarchive_validation.pl
```diff
@@ -304,20 +310,6 @@ =head2 Methods
...
 
-################################################################
-################################################################
-## Check the CandID/PSCID Match It's possible that the CandID ## 
-## exists, but doesn't match the PSCID. This will fail further #
-## down silently, so we explicitly check that the data is ######
-## correct here. ###############################################
-################################################################
-################################################################
-my $CandMismatchError= $utility->validateCandidate($subjectIDsref);
```
cmadjar: removed the call to `validateCandidate()` as it is now part of the determineSubjectIDs function and is returned into $subjectIDsref->{'CandMismatchError'}

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1547,69 +1559,75 @@ sub CreateMRICandidates {
...
+    my ($subjectIDsref, $tarchiveInfo, $User, $centerID, $upload_id) = @_;
+
+    my ($sex, $query, $message);
+    my $dbh    = $this->{dbhr};
+    my $pscID  = $subjectIDsref->{'PSCID'};
+    my $candID = $subjectIDsref->{'CandID'};
+
+
+    # return from the function if createCandidate config setting is not set
+    return if (!NeuroDB::DBI::getConfigSetting($dbh, 'createCandidates'));
```
cmadjar: changed a bit the logic here. Checked first if `createCandidates` config setting is set to `true`. If set to `false`, then return from this function candidates should not be created as per the config setting.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1547,69 +1559,75 @@ sub CreateMRICandidates {
...
+
+
+    # Check that no candidates with the same PSCID is already registered
+    if ($pscID ne 'scanner' && NeuroDB::MRI::subjectIDExists('PSCID', $pscID, $dbh)) {
 
+        $message = "ERROR: Cannot create candidate ($pscID, $candID) as "
```
cmadjar: then check if the candidate's `PSCID` and `CandID` values stored in `$subjectIDsref` do not already refer to another candidate (there should not be two candidates with the same `PSCID` or `CandID`). If found a matching candidate, exit with proper error message.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1547,69 +1559,75 @@ sub CreateMRICandidates {
...
-                       $tarchiveInfo->{'PatientID'}. "\n ".
-                       "The dicom header PatientName is: ". 
-                       $tarchiveInfo->{'PatientName'}. "\n\n";
-            $this->writeErrorLog(
-                $message, $NeuroDB::ExitCodes::INSERT_FAILURE
-            );
-            $this->spool($message, 'Y', $upload_id, $notify_notsummary);
-            exit $NeuroDB::ExitCodes::INSERT_FAILURE;
-     }
+    chomp($User);
```
cmadjar: if we end up there, then all is good to create a new candidate with PSCID, CandID.

uploadNeuroDB/tarchiveLoader.pl
```diff
@@ -411,26 +411,10 @@ =head2 Methods
...
-###### Get the SessionID #######################################
-################################################################
-if (!defined($subjectIDsref->{'visitLabel'})) { 
-    $subjectIDsref->{'visitLabel'} = 
-    $utility->lookupNextVisitLabel(
-        $subjectIDsref->{'CandID'}, 
-        \$dbh
-    ); 
-}
-my ($sessionID) =
```
cmadjar: moved the `getSessionID` function in line 585 where it is needed. Otherwise, queries in that function would fail in case of invalid `CandID`. Since the `SessionID` in this script is only used to updated the `tarchive` and `mri_upload` tables, the call to that function was moved in the same section.

uploadNeuroDB/tarchive_validation.pl
```diff
@@ -291,38 +292,13 @@ =head2 Methods
...
+if (defined $subjectIDsref->{'CandMismatchError'}) {
+    print "$subjectIDsref->{'CandMismatchError'} \n";
     ##Note that the script will not exit, so that further down
     ##it can be inserted per minc into the MRICandidateErrors
 }
-################################################################
-############ Get the SessionID #################################
-################################################################
-my ($sessionID) =
-    $utility->setMRISession($subjectIDsref, \%tarchiveInfo, $upload_id);
```
cmadjar: remove this as `$sessionID` is not used here and `setMRISession` is calling `MRI::getSessionID` which would be failing when `CandID` is not valid.

cmadjar:
@nicolasbrossard if you could review and test this PR, that would be awesome! Thank you!

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1770,72 +1804,96 @@ or a phantom
...
     
     #################################################################
     ## Check if CandID exists
     #################################################################
     my $query = "SELECT CandID, PSCID FROM candidate WHERE CandID=?";
-    my $sth = ${$this->{'dbhr'}}->prepare($query);
-    $sth->execute($subjectIDsref->{'CandID'});
-    print "candidate id " . $subjectIDsref->{'CandID'} . "\n" 
-	if ($this->{verbose});
+    my $sth   = $dbh->prepare($query);
```
nicolasbrossard: I think the entire check block could be simplified. I would only query the DB once, like this:

```sql
SELECT c1.CandID, c2.PSCID
FROM candidate c1 
LEFT JOIN candidate c2 ON (c1.CandID=c2.CandID AND c2.PSCID=?)
WHERE c1.CandID=?
```

If you get no results, then the CandID is invalid. If you get a result but the PSCID column is NULL, then the CandID and PSCID do not match. There is no need to check whether the provided PSCID exists: validating that it matches is enough IMHO

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1770,72 +1804,96 @@ or a phantom
...
-    ################ Check if visitLabel exists ################
+    ################ Check if visit label is valid##############
     ############################################################
-
     $query = "SELECT Visit_label FROM Visit_Windows WHERE BINARY Visit_label=?";
-    $sth =  ${$this->{'dbhr'}}->prepare($query);
-    $sth->execute($subjectIDsref->{'visitLabel'});
+    $sth   = ${$this->{'dbhr'}}->prepare($query);
+    $sth->execute($visit_label);
+
```
nicolasbrossard: Instead of creating a statement and then executing it, you can do this all at once:

```perl
my @rows = $this->{'dbhr'}->selectall_array($query, {}, $visit_label);
```

and then check if `@rows == 0` afterwards. I think there are other occurence of this pattern elsewhere in the code also.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1770,72 +1804,96 @@ or a phantom
...
         return $CandMismatchError;
     } elsif (($sth->rows == 0) && ($subjectIDsref->{'createVisitLabel'})) {
-        print LOG "\n=> Will create visit label $subjectIDsref->{'visitLabel'}\n";
+        $message = "\n=> Will create visit label $visit_label in Visit_Windows\n";
+        $this->spool($message, 'Y', $upload_id, $notify_notsummary);
     } 
 
-   return $CandMismatchError;
+    return $CandMismatchError;
```
nicolasbrossard: So if all goes well, `$CandMismatchError` remains uninitialized and the method will return `undef`, right? If that's the case, I would state so explicitly with `return undef`.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
nicolasbrossard: Good. Since there is a `ConfigOB` that exists now, we should use it to fetch the `createCandidates` setting. This could be done at a later stage though and would in fact constitue a good starting point for someone who wants to start "Moosing" around. :)

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1547,69 +1573,75 @@ sub CreateMRICandidates {
...
-                       $tarchiveInfo->{'PatientName'}. "\n\n";
-            $this->writeErrorLog(
-                $message, $NeuroDB::ExitCodes::INSERT_FAILURE
-            );
-            $this->spool($message, 'Y', $upload_id, $notify_notsummary);
-            exit $NeuroDB::ExitCodes::INSERT_FAILURE;
-     }
+    chomp($User);
+    $candID = NeuroDB::MRI::createNewCandID($dbh) unless $candID;
+    $query  = "INSERT INTO candidate ".
```
nicolasbrossard: I just realized something: we do not check if the new PSCID is valid for the given centerID. You could create a new candidate in DB RaisinBread that has PSCID='OTT001' and center ID = <MTL's center ID>. I know that LORIS checks this when a new candidate is created. We should probably do the same. Bummer...We can postpone this too.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
cmadjar: done

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
cmadjar: perfect! Done!

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1547,69 +1573,75 @@ sub CreateMRICandidates {
...
     ############################################################
-    my $this = shift;
-    my $query = '';
-    my ($subjectIDsref,$sex,$tarchiveInfo,$User,$centerID, $upload_id) = @_;
-    my ($message);
+    my $this  = shift;
+    my ($subjectIDsref, $tarchiveInfo, $User, $centerID, $upload_id) = @_;
+
+    my ($sex, $query, $message);
+    my $dbh    = $this->{dbhr};
```
nicolasbrossard: Just to avoid confusion, I would rename that variable into `$dbhr`

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
nicolasbrossard: OK

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
cmadjar: ok. Done in the next commit.

docs/scripts_md/MRIProcessingUtility.md
```diff
@@ -117,18 +117,25 @@ RETURNS:
...
   - $header       : study meta data header
 
-### determineSubjectID($scannerID, $tarchiveInfo, $to\_log, $upload\_id)
+### determineSubjectID($scannerID, $tarchiveInfo, $to\_log, $upload\_id, $User, $centerID)
 
-Determines subject's ID based on scanner ID and DICOM archive information.
+This function does:
+1) Determines subject's ID based on scanner ID and DICOM archive information.
+2) Call the `CreateMRICandidate` function (will create the candidate if it does 
+not exists and `createCandidates` config option is set to yes
```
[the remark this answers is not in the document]
nicolasbrossard: Huh? I don't see it...The `.md` file is probably not up to date.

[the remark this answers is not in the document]
cmadjar: oops... Forgot to run pod2markdown apparently. Done in the next commit!!

cmadjar:
@nicolasbrossard ready for another review!

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
nicolasbrossard: I think it'd be better if the query was prepared (with argument placeholders) and then executed with the appropriate arguments instead of building a "one-shot" query in which all the arguments are quoted.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -368,25 +368,33 @@ sub extractAndParseTarchive {
 
 =pod
 
-=head3 determineSubjectID($scannerID, $tarchiveInfo, $to_log, $upload_id)
+=head3 determineSubjectID($scannerID, $tarchiveInfo, $to_log, $upload_id, $User, $centerID)
 
-Determines subject's ID based on scanner ID and DICOM archive information.
+This function does:
+1) Determines subject's ID based on scanner ID and DICOM archive information.
```
nicolasbrossard: **Determine** subject's....

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1574,69 +1600,83 @@ sub CreateMRICandidates {
...
+    RegistrationCenterID, Date_active, Date_registered,      UserID,
+    Entity_type
+  ) VALUES (
+    ?,                    ?,           ?,                    ?,
+    ?,                    NOW(),       NOW(),                ?,
+    'Human'
+  )
+QUERY
+
+    print $query . "\n" if ($this->{debug});
```
nicolasbrossard: Why not just `print "$query\n" if ($this->{debug});` ?

PR #464, opened 2019-07-17
[dicomTar.pl] Exit when insertion or update fails

cmadjar:
### Description

`dicomTar.pl` continues executing insertion into the tarchive tables even if one insertion failed. In this PR, we catch the insertion failure and return the error so `dicomTar.pl` can be stopped and further insertions prevented.

Needed to modified the database function of `DCMSUM.pm` to return the error when an SQL insertion failed during the execution of `dicomTar.pl`.

Note: ideally, the returned error message would be inserted in the `notification_spool` table which should be a separate PR.

### This resolves

https://redmine.cbrain.mcgill.ca/issues/13841

### DIFF
--- a/dicom-archive/DICOM/DCMSUM.pm
+++ b/dicom-archive/DICOM/DCMSUM.pm
@@ -128,7 +128,7 @@ sub database {
     }
 
     # whether the query worked fine
-    my $success = undef;
+    my ($success, $error);
     # check if this StudyUID is already in your database
     my ($unique_study, $message) = $self->is_study_unique($dbh, $update, $Archivemd5);
     if (!$unique_study && !$update) {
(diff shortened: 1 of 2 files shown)


cmadjar:
Hi Corey,

Thank you for reviewing and testing. 

Maybe you could try modifying the `tarchive` table in your sandbox (drop or rename a few columns) so that the insertion in the `tarchive` table fails and see if the pipeline stops and prints the correct error (in this case, that the query to insert into the `tarchive` table failed).

Then you could restore the `tarchive` table and modify the `tarchive_series` table and rerun `dicomTar.pl`. This time, you should see that the query to insert into `tarchive_series` failed.

Let me know if you have any question.

PR #466, opened 2019-07-17
[Moose] ConfigOB - more config options available via ConfigOB

cmadjar:
### Description

In this PR, all Config options that returns a string will be queried using the ConfigOB instead of the function in DBI.pm.

PR #468, opened 2019-08-07
Gzip NIfTI files to save disk space

cmadjar:
### Description

NIfTI files created by `mnc2nii` are uncompressed files taking a lot of disk space. However, most softwares using NIfTI files are supporting the compressed version of NIfTI (`.nii.gz`). 

In this PR, the created NIfTI files are automatically gzipped during the creation of NIfTI files and a tool script has been created to be able to gzip all NIfTI files already present in the database if one wants to gzip every single NIfTI files.

### This resolves

No issue opened but needed to save disk space on a LORIS project (before gzipping 2.3TB of disk space was used, after gzipping 714GB of space was used).

### How to test this

- [ ]  Run the insertion pipeline after having made sure that the `create_nii` config setting is set to 'Yes'
- [ ] If you already had NIfTI files on your test VM, run the tool `gzip_nifti_files.pl` with the `-profile prod` option and check that NIfTI files are being gzipped and updated in the database

Note: if a NIfTI file is not present on the filesystem for whichever reason, the tool script will issue a warning and go to the next NIfTI file.

### DIFF
--- a/tools/gzip_nifti_files.pl
+++ b/tools/gzip_nifti_files.pl
@@ -118,10 +118,12 @@ $db->connect();
 
 
 # ===========================================
-## Establish database connection
+## Get config setting using ConfigOB
 # ===========================================
 
-my $data_dir   = &NeuroDB::DBI::getConfigSetting(\$dbh, 'dataDirBasepath');
+my $configOB = NeuroDB::objectBroker::ConfigOB->new(db => $db);
+
+my $data_dir   = $configOB->getDataDirPath();
 $data_dir      =~ s#/$##;
 
 
--- a/tools/gzip_nifti_files.pl
+++ b/tools/gzip_nifti_files.pl
@@ -1,0 +1,219 @@
+#! /usr/bin/perl
+
+=pod
+
+=head1 NAME
+
+gzip_nifti_files.pl -- Gzip all NIfTI files found in the LORIS database system
+
+=head1 SYNOPSIS
+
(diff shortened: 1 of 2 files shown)


cmadjar:
Added the block label as once #466 is merged, will need to use `ConfigOB.pm` library to grep the `dataDirBasePath` config setting.

cmadjar:
@coreyAbout Thank you for the review!! Answers to your comment are below with appropriate question number :).

1. You're right. I changed the permissions in the next commit. Thanks!

2. Those warnings are there to let you know that there are some NIfTI files that were found in the database table but were not found on the filesystem. I guess the discrepancy comes from some manual removal you may have done before?

3. Is it possible that the `environment` file changed when you checked out my branch? If so, then all the Path settings from that file would lead to incorrect paths and could lead to the `non-existent MRI pipeline setup` error?

coreyAbout:
1. Fixed.
2. Still there. I cleaned everything up then uploaded a totally clean scan, the same thing appears again. It seems that there are those files created in some tables during uploading process, but for some reason they do not appear on the file system?  This might be another issue, though, not necessarily of this PR.
3. Fixed.

cmadjar:
@coreyAbout Looks like for 2. some of the NIfTI files never get created but an entry for them is still registered in the database. In the last commit I added a check to make sure the NIfTI file exists on the filesystem before inserting an entry in the parameter file. So I guess it should be fixed for 2. now.
Thanks again for reviewing!

PR #469, opened 2019-08-16
[mri_upload] Using LORIS API instead of database insert

xlecours:
draft 1

This is meant to replace batch_uploads_imageuploader.pl

Note: patient name is required for every lines (even phantoms). This is to capture the candid and visit_label to build the url of the enpoint. (/candidate/$candid/$visit_label/dicoms) 

help : `batch_uploads_imageuploader.py -h`

### DIFF
--- a/batch_uploads_imageuploader.py
+++ b/batch_uploads_imageuploader.py
@@ -1,0 +1,143 @@
+import argparse
+import textwrap
+import sys, getopt  # For script options
+import os           # Operating System library to create directories and files
+import errno        # For python 2.7 compatibility
+import getpass      # For input prompt not to show what is entered
+import json         # Provide convenient functions to handle JSON objects 
+import requests     # To handle HTTP requests
+
+import warnings
+warnings.simplefilter('ignore') # Because I am using unverified ssl certificates 
+
+parser = argparse.ArgumentParser(
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    description=textwrap.dedent("""\
+This script runs the Loris-MRI insertion pipeline on multiple scans. The list of
+scans are provided through a text file (e.g. C<list_of_scans.txt>) with one scan
+details per line.
+
+The scan details includes the path to the scan, identification as to whether the
+scan is for a phantom (Y) or not (N), and the candidate name. As opposred to the
+perl script, this one always requires a patient_name, even for non-phantom entries.
+
+Like the LORIS Imaging Uploader interface, this script also validates the
+candidate's name against the (start of the) filename and creates an entry in the
+C<mri_upload> table.
+
+An example of what C<list_of_scans.txt> might contain for 3 uploads to be
+inserted:
+
+ /data/incoming/PSC0001_123457_V1.tar.gz N PSC0000_123456_V1
+ /data/incoming/lego_Phantom_MNI_20140101.zip Y
+ /data/incoming/PSC0001_123457_V1_RES.tar.gz N PSC0000_123456_V1
+"""
+))
+
+parser.add_argument('-b', '--baseurl', required=True)
+parser.add_argument('-i', '--inputfile', required=True)
+parser.add_argument('-a', '--autostart', action='store_const', const=True)
+
+args = parser.parse_args()
+
+def prettyPrint(string):
+    print(json.dumps(string, indent=2, sort_keys=True))
+
+def dicomUpload(filepath, isphantom, patientname):
+    pscid, candid, visit_label = patientname
+    filename = os.path.basename(filepath)
+    files = {'mriFile': (filename, open(filepath, "rb"), 'application/x-tar')}
+    payload = {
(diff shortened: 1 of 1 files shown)


batch_uploads_imageuploader.py
```diff
@@ -0,0 +1,143 @@
...
+
+parser = argparse.ArgumentParser(
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    description=textwrap.dedent("""\
+This script runs the Loris-MRI insertion pipeline on multiple scans. The list of
+scans are provided through a text file (e.g. C<list_of_scans.txt>) with one scan
+details per line.
+
+The scan details includes the path to the scan, identification as to whether the
+scan is for a phantom (Y) or not (N), and the candidate name. As opposred to the
```
mohadesz: Typo: Should be opposed

xlecours: Fixed. Thanks

batch_uploads_imageuploader.py
```diff
@@ -0,0 +1,147 @@
...
+    return json.loads(response.content.decode('ascii'))
+
+if __name__ == '__main__':
+    
+    credentials = {
+        'username': input('LORIS username: '), 
+        'password': getpass.getpass('password: ')
+    }
+    
+    jwtoken = json.loads(requests.post(
```
xlecours: A check should be added to see if login was successful.

batch_uploads_imageuploader.py
```diff
@@ -0,0 +1,147 @@
...
+
+To start the mri_upload process on the LORIS server for each file, use the 
+autostart (-a) option when calling this script.
+
+"""
+))
+
+parser.add_argument('-b', '--baseurl', required=True, help='LORIS API url with version.')
+parser.add_argument('-i', '--inputfile', required=True, help='The list of scans.')
+parser.add_argument('-a', '--autostart', action='store_const', const=True, help='If present, will attemp to start the mri_upload process for each scans.')
```
mohadesz: Typo: Should be "attempt"

batch_uploads_imageuploader.py
```diff
@@ -0,0 +1,147 @@
...
+    payload = {
+        'IsPhantom': 'true' if isphantom == 'Y' else 'false',
+        'CandID': candid,
+        'PSCID': pscid,
+        'Visit': visit_label
+    }
+    response = requests.post(
+        url = args.baseurl + '/candidates/' + candid + '/' + visit_label + '/dicoms/',
+        headers = {'Authorization': 'Bearer %s' % jwtoken, 'LORIS-Overwrite': 'overwrite'},
+        verify = False,
```
xlecours: `verify` needs to be removed before merging.

batch_uploads_imageuploader.py
```diff
@@ -0,0 +1,147 @@
...
+            print('Upload failed')
+            print(err.args[0])
+            continue
+            
+        print('Upload succesfull')
+        print('mri_upload_id: ' + str(mri_upload_id))
+        
+        if (args.autostart):
+            try:
+                responsebody = startProcess(os.path.basename(filepath), patientname.split('_',2), mri_upload_id)
```
cmadjar: what happens here if the PatientName is `PSCID_CandID_VisitLabel_some_other_string`?

If I understand correctly, it would split like:
PSCID = PSCID
CandID = CandID
VisitLabel = `VisitLabel_some_other_string` instead of `VisitLabel` which would fail when entering the `VisitLabel` in the URL of the API. We should support both scenarios (`PSCID_CandID_VisitLabel` and `PSCID_CandID_VisitLabel_some_other_string`)

batch_uploads_imageuploader.py
```diff
@@ -0,0 +1,147 @@
...
+    responsebody = response.content.decode('ascii')
+    if (response.status_code != 200):
+        raise ValueError(responsebody)
+    
+    
+    data = json.loads(responsebody)
+    return data['mri_uploads'].pop()['mri_upload_id']
+
+def startProcess(filename, patientname, mri_upload_id):
+    pscid, candid, visit_label = patientname
```
cmadjar: shouldn't the split on patientName be here instead of when calling the function?

xlecours:
@cmadjar Thanks for the review! I will work on it.

nicolasbrossard:
Curious as to why was Python was chosen to implement this while the rest of the LORIS-MRI codebase is in Perl?

batch_uploads_imageuploader.py
```diff
@@ -0,0 +1,147 @@
...
+    )
+    if (response.status_code != 202):
+        raise ValueError(responsebody)
+        
+    return json.loads(response.content.decode('ascii'))
+
+if __name__ == '__main__':
+    
+    credentials = {
+        'username': input('LORIS username: '),
```
cmadjar: is there a way to not have to type in username and password? Have it set somewhere in a config?

xlecours: you could ask for a token?

Issue #465, opened 2019-07-17
Grep the configuration by the ConfigOB


cmadjar:
Since there is a ConfigOB that exists now, we should use it to fetch the `createCandidates` setting (or other settings). This could be done at a later stage though and would in fact constitue a good starting point for someone who wants to start "Moosing" around. :)

All config settings that are strings are now queried via the ConfigOB.

Left to do:
- [x] Config settings that come as a single text or drop down entry (#466)
- [x] Config settings that returns boolean values
- [x] Config settings that return arrays #770

PR #500, opened 2019-10-08
[Moose] ConfigOB - grepping the boolean config settings using ConfigOB

cmadjar:
### Description

In this PR, all boolean Config options are queried using the `ConfigOB` instead of the function in `DBI.pm`.

### This fixes partly

- [x] #465

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
+++ b/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
@@ -498,7 +498,7 @@ sub determinePSC {
     # ----------------------------------------------------------------
     ## Get config settings using ConfigOB
     # ----------------------------------------------------------------
-    my $configOB              = $this->{configOB};
+    my $configOB              = $this->{'configOB'};
     my $lookupCenterNameUsing = $configOB->getLookupCenterNameUsing();
 
 
@@ -1175,7 +1175,7 @@ sub registerScanIntoDB {
     # ----------------------------------------------------------------
     ## Get config settings using ConfigOB
     # ----------------------------------------------------------------
-    my $configOB = $this->{configOB};
+    my $configOB = $this->{'configOB'};
     my $data_dir = $configOB->getDataDirPath();
     my $prefix   = $configOB->getPrefix();
 
@@ -1506,7 +1506,7 @@ sub moveAndUpdateTarchive {
     # ----------------------------------------------------------------
     ## Get config settings using ConfigOB
     # ----------------------------------------------------------------
-    my $configOB = $this->{configOB};
+    my $configOB = $this->{'configOB'};
     my $tarchivePath = $configOB->getTarchiveLibraryDir();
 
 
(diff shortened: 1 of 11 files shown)


uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1589,11 +1595,9 @@ sub CreateMRICandidates {
     ## Create non-existent candidate if the profile allows for #####
     ## Candidate creation ##########################################
     ################################################################
-    if (!NeuroDB::MRI::subjectIDExists(
-            $subjectIDsref->{'CandID'},
-            $this->{dbhr}
-        ) 
-        && (NeuroDB::DBI::getConfigSetting($this->{dbhr},'createCandidates'))
+    my $configOB = $this->{'configOB'};
+    if (!NeuroDB::MRI::subjectIDExists($subjectIDsref->{'CandID'}, $this->{dbhr})
```
nicolasbrossard: No need to instantiate `$configOB` again: it was already defined on line 1585. Also, the definition of `$createCandidates` should go here instead of further up.

uploadNeuroDB/NeuroDB/objectBroker/ConfigOB.pm
```diff
@@ -348,6 +352,66 @@ sub getLivingPhantomRegex {
...
 }
 
+=head3 getCreateNii()
+
+Get the create_nii Config setting.
+
+RETURN: (boolean) 1 if create_nii is set to Yes in the Config module, 0 otherwise
+
+=cut
+sub getCreateNii {
```
nicolasbrossard: I think it would be a good idea to have a private function `getBoolean` that the other methods would call instead of "hardcoding" the logic inside each method. What do you think?

cmadjar: I agree!

cmadjar:
@nicolasbrossard Thank you for the review! I integrated all your comments in the last commit. Please re-review :)

uploadNeuroDB/NeuroDB/objectBroker/ConfigOB.pm
```diff
@@ -143,6 +143,23 @@ my $getConfigSettingRef = sub {
     return $result->[0]->{'value'};
 };
 
+
+=head3 &$getBoolean($value)
```
nicolasbrossard: Nitpicking here: I would rename getBoolean into getBooleanRef to emphasize the fact that this variable is a _reference_ to a method.

uploadNeuroDB/NeuroDB/objectBroker/ConfigOB.pm
```diff
@@ -143,6 +143,23 @@ my $getConfigSettingRef = sub {
...
+
+Private method. This methods returns 1 if a boolean value is set to either 'true'
```
nicolasbrossard: ...This **method** returns...

cmadjar:
@nicolasbrossard Thank you! Your suggestions have been integrated in the last commit. Please re-review.

PR #503, opened 2019-10-09
[Install] Removed the site question from the install script

cmadjar:
### Description

PR #334 removed the creation of subdirectories in the `/data/incoming` directory but the question about the list of sites to provide at the time of install was not removed.

This PR updates the install instruction from the install scripts (Linux and Mac OS) and the `README.md` to remove that question when running the install script since the variable `site` is not used anywhere in the script.

### This resolves

- [x] part 2 of #334 that was missed...

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,6 @@ For the defacing scripts, you will also need to download the pre-compiled `bic-m
  * What is the project name? $projectname
  * What is your email address? 
  * What prod file name would you like to use? default: prod  [leave blank]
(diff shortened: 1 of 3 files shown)


cmadjar:
@Jkat Any chance you could review this PR? Figured you would be a good candidate since you took care of the install of LORIS for a while and are starting to dive into MRI code. 

Should be pretty quick to review as the changes are very straightforward ;).

nicolasbrossard:
LGTM

PR #506, opened 2019-11-07
remove unnecessary extra extraction from tarchive_validation.pl

cmadjar:
### Description

In `tarchive_validation.pl`, the extraction of the tarchive is done to call a function that does not exist and have not existed for years. This removes the unnecessary additional extraction that contributes to slowing the insertion pipeline.

Notes for imaging meeting:
- even if `dicomFilter` was filtering the DICOM, it would not update the `tarchive` so it would just be doing it with no follow up
- no studies are using such a function in their `prod` file (CCNA, IBIS, PREVENT-AD, QPN, myelineurogene...)

Should be safe to remove but will discuss at the next imaging meeting just in case.

### DIFF
--- a/uploadNeuroDB/tarchive_validation.pl
+++ b/uploadNeuroDB/tarchive_validation.pl
@@ -314,20 +314,6 @@ if (defined $subjectIDsref->{'CandMismatchError'}) {
 }
 
 ################################################################
-### Extract the tarchive and feed the dicom data dir to ######## 
-### The uploader ###############################################
-################################################################
-my ($ExtractSuffix,$study_dir,$header) = 
-    $utility->extractAndParseTarchive($tarchive, $upload_id);
-
-################################################################
-# Optionally do extra filtering on the dicom data, if needed ###
-################################################################
-if ( defined( &Settings::dicomFilter )) {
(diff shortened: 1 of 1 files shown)


uploadNeuroDB/tarchive_validation.pl
```diff
@@ -313,20 +313,6 @@ =head2 Methods
     ##it can be inserted per minc into the MRICandidateErrors
 }
 
-################################################################
-### Extract the tarchive and feed the dicom data dir to ######## 
-### The uploader ###############################################
-################################################################
-my ($ExtractSuffix,$study_dir,$header) =
```
coreyAbout: Is wondering what the variable `$ExtractSuffix` and `$header` was doing here before? The `$study_dir` was used in below, but not these two.

cmadjar: It's there because the function used to extract and parse the tarchive returns 3 values after completion. Those two variables are not useful in this case (as the extraction, lol).

coreyAbout: I suppose so as well, but wanted to double check. It looks great to me.

PR #507, opened 2019-11-13
Fixes the OB to grep the config setting createCandidate

cmadjar:
The function `getCreateCandidates` of the config object broker was not looking for the right config setting.


nicolasbrossard:
I also checked other methods for bugs similar to this one but could not find any. :+1:

cmadjar:
Awesome! Thank you!

PR #508, opened 2019-11-14
Allow deletion of a file with a specific base name in delete_imaging_upload.pl

nicolasbrossard:
Addition of a new option to `delete_imaging_upload.pl`: `-file basename`, which allows the deletion of a file with a specific base name. The file must meet one of these criteria:

1. Listed in table `file` and is directly produced by the MRI pipeline (as in `TarchiveSource != NULL`).
2. Listed in 'files` but has been produced/computed using another file directly produced by the pipeline (i.e referenced in `files_intermediary`).
3. Listed in `parameter_file` and is associated to parameter `check_pic_filename`, `check_nii_filename`, `check_bval_filename` or `check_bvec_filename`.

Note that if a file is deleted, all the files that were produced using it are also deleted.

### DIFF
--- a/tools/delete_imaging_upload.pl
+++ b/tools/delete_imaging_upload.pl
@@ -10,59 +10,59 @@ delete_mri_upload.pl -- Delete everything that was produced (or part of what was
 =head1 SYNOPSIS
 
 perl delete_mri_upload.pl [-profile file] [-ignore] [-backup_path basename] [-protocol] [-form] [-uploadID list_of_uploadIDs]
-            [-type list_of_scan_types] [-defaced] [-file fileBaseName] [-nosqlbk] [-nofilesbk]
+            [-type list_of_scan_types] [-defaced] [-basename fileBaseName] [-nosqlbk] [-nofilesbk]
 
 Available options are:
 
--profile            : name of the config file in C<../dicom-archive/.loris_mri> (defaults to C<prod>).
+-profile               : name of the config file in C<../dicom-archive/.loris_mri> (defaults to C<prod>).
 
--ignore             : ignore files whose paths exist in the database but do not exist on the file system.
-                      Default is to abort if such a file is found, irrespective of whether a backup file will
-                      be created by the script (see C<-nofilesbk> and C<-nosqlbk>). If this option is used, a 
-                      warning is issued and program execution continues.
+-ignore                : ignore files whose paths exist in the database but do not exist on the file system.
+                         Default is to abort if such a file is found, irrespective of whether a backup file will
+                         be created by the script (see C<-nofilesbk> and C<-nosqlbk>). If this option is used, a 
+                         warning is issued and program execution continues.
                
--nofilesbk          : when creating the backup file for the deleted upload(s), do not backup the files produced by
-                      the imaging pipeline (default is to backup these files).
+-nofilesbk             : when creating the backup file for the deleted upload(s), do not backup the files produced by
+                         the imaging pipeline (default is to backup these files).
                
--backup_path <path> : specify the path of the backup file, which by default contains a copy of everything that the
-                      script will delete, both on the file system and in the database (but see C<-nofilesbk> and
-                      C<-nosqlbk>). The extension C<.tar.gz> will be added to this base name to build the name of the final
-                      backup file. If a file with this resulting name already exists, an error message is shown and the script
-                      will abort. Note that C<path> can be an absolute path or a path relative to the current directory. A 
-                      backup file is always created unless options C<-nofilesbk> and C<-nosqlbk> are both used. By default, the
-                      backup file name is C<imaging_upload_backup.tar.gz> and is written in the current directory. Option 
-                      C<-backup_path> cannot be used if C<-nofilesbk> and C<-nosqlbk> are also used.
-
--file fileBaseName  : basename of the file to delete. The file is assumed to either exist in table C<files> or table 
-                      C<parameter_files>. This option should be used when targeting a specific (unique) file for deletion.
-                      Note that the file will be deleted from both the database and the filesystem. This optin cannot be 
-                      used with options C<-defaced> and C<-form>.
+-backup_path <path>    : specify the path of the backup file, which by default contains a copy of everything that the
+                         script will delete, both on the file system and in the database (but see C<-nofilesbk> and
+                         C<-nosqlbk>). The extension C<.tar.gz> will be added to this base name to build the name of the final
+                         backup file. If a file with this resulting name already exists, an error message is shown and the script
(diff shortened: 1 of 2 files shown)


christinerogers:
This is fantastic @nicolasbrossard.  
* Would you be open to better specifying the _argument name_ to something like `-filebasename` or  `-fileroot` to add clarity ?
I think `-file` is too standard and suggests it will only delete the file specified, instead of how it actually deletes all downstream data generated from it.

* One question - does it delete any file starting with that string? e.g. partial matches for all kinds of endings?

nicolasbrossard:
@christinerogers @cmadjar Thanks C! I can change the option name to `-filebasename` to make it clearer. And no, it does not perform partial matches: only exact matches, case-insensitive though. I am starting to think I should make this case sensitive. Any thoughts?

cmadjar:
@nicolasbrossard maybe making the match case-sensitive would be a good idea in this case. This way you are sure to really just delete that file.

christinerogers:
Thanks, Nic. And I'm ok with adding case sensitivity. 
What I'm really thinking about is `Note that if a file is deleted, all the files that were produced using it are also deleted.` 
Is there a prompt/warning generated for the user saying roughly `All the following files that were generated from the filebasename you specified are about the be deleted -- OK?`

nicolasbrossard:
@christinerogers There is no issued warning per se but the script will indicate which files are being deleted. Also, all files that are deleted are backed up to allow undoing an erroneous command.

christinerogers:
remind me, is there a `-confirm` mode? 
because otherwise -- because it deletes *all* downstream data -- we need a more specific warning.

nicolasbrossard:
Good idea. I'll issue another PR for that new feature though. Stay tuned!

tools/delete_imaging_upload.pl
```diff
@@ -32,6 +32,11 @@ =head1 SYNOPSIS
                       backup file is always created unless options C<-nofilesbk> and C<-nosqlbk> are both used. By default, the
                       backup file name is C<imaging_upload_backup.tar.gz> and is written in the current directory. Option 
                       C<-backup_path> cannot be used if C<-nofilesbk> and C<-nosqlbk> are also used.
+
+-file fileBaseName  : basename of the file to delete. The file is assumed to either exist in table C<files> or table 
+                      C<parameter_files>. This option should be used when targeting a specific (unique) file for deletion.
+                      Note that the file will be deleted from both the database and the filesystem. This optin cannot be
```
cmadjar:
```suggestion
                      Note that the file will be deleted from both the database and the filesystem. This option cannot be 
```

tools/delete_imaging_upload.pl
```diff
@@ -129,6 +134,11 @@ =head1 DESCRIPTION
    an error message and abort, leaving the database and file system untouched, if:
        a) A MINC file should have a corresponding defaced file but does not.
        b) A MINC file that has been defaced has an associated file that is not a defaced file.
+
+4. Delete a specific file that exists in table C<files> or table C<parameter_file> (in which case it will be associated to parameter
+   C<check_pic_filename>, C<check_nii_filename>, C<check_bval_filename> or C<check_bvec_filename>). Note that once a file is
+   deleted, all the files that have been derived/built using this deleted file will also be deleted. Use option C<-file fileBaseName>, 
+   where C<fileBaseName> is the basename of the file to delete, along with option C<-uploadID> to use the script this way.
```
cmadjar: what happens if the file has a link to the `mri_violations_log` when a caveat is set to `warning`?

tools/delete_imaging_upload.pl
```diff
@@ -871,43 +919,71 @@ sub getIntermediaryFilesRef {
...
+               . '    SELECT f2.FileID '
+               . '    FROM files f2 '
+               .      $mriScanTypeJoin
+               . '    WHERE f2.TarchiveSource = ? '
+               .      $mriScanTypeAnd
+               . ') '
+               . 'AND SUBSTRING_INDEX(f.File, "/", -1) = ?';
+        @queryArgs = ($tarchiveID, @$scanTypesToDeleteRef, $fileBaseName); 
+    # Either -file was not used or it was used but nothing matched in table files.
+    #
```
cmadjar:
```suggestion

```

tools/delete_imaging_upload.pl
```diff
@@ -917,43 +993,75 @@ =head3 getParameterFilesRef($dbh, $tarchiveID, $dataDirBasePath, $scanTypesToDel
...
-           . "WHERE pt.Name IN ('check_pic_filename', 'check_nii_filename', 'check_bval_filename', 'check_bvec_filename') "
-           . 'AND ( '
-           . '         f.TarchiveSource = ? '
-           . '      OR f.SourceFileID IN (SELECT FileID FROM files WHERE TarchiveSource = ?) '
-           . ') '
-           . $mriScanTypeAnd;
 
-    my $filesRef = $dbh->selectall_arrayref(
-        $query, { Slice => {} }, $tarchiveID, $tarchiveID, @$scanTypesToDeleteRef
+    # If -file was used but nothing matched in tables files and fiels_intermediary
```
cmadjar:
```suggestion
    # If -file was used but nothing matched in tables files and files_intermediary
```

tools/delete_imaging_upload.pl
```diff
@@ -917,43 +993,75 @@ =head3 getParameterFilesRef($dbh, $tarchiveID, $dataDirBasePath, $scanTypesToDel
...
+    # Try finding a file in parameter_file thjat matches the -file argument
```
cmadjar:
```suggestion
    # Try finding a file in parameter_file that matches the -file argument
```

tools/delete_imaging_upload.pl
cmadjar: hehe, I think I answered my question by reading the code. Looks like this is handled. Just let me know if I got it right.

nicolasbrossard: Yeah, files in `mri_violations_log` are deleted too.

tools/delete_imaging_upload.pl
nicolasbrossard: Done

tools/delete_imaging_upload.pl
nicolasbrossard: Done

tools/delete_imaging_upload.pl
nicolasbrossard: Done

tools/delete_imaging_upload.pl
nicolasbrossard: Done

nicolasbrossard:
Ready for another review

PR #512, opened 2019-12-10
[BIDS EEG import bug fix] This fixes a bug happening during the file type determination

cmadjar:
### Description

This fixes a bug during the file type determination in the python scripts. Instead of looking for the file type string in the whole file path, it now uses a regex to properly match the extension as it should have been done from the beginning...

### To test this

Use the code from this branch on the cuban dataset that Iris has.

### DIFF
--- a/python/lib/imaging.py
+++ b/python/lib/imaging.py
@@ -12,6 +12,8 @@ import nibabel as nib
 from nilearn import plotting
 from nilearn import image
 
+import lib.exitcode
+
 
 __license__ = "GPLv3"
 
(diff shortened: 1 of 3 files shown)


cmadjar:
This has been kindly tested by Iris from the Cuban Brain Mapping project and that resolved the bug she was experiencing so merging it.

cmadjar:
@christinerogers actually the regular expression would be able to determine if a file is `.zip` or `.tgz` if those files make it to the BIDS standard one day. I added the `(\.gz)?` in the regex so that compressed NIfTI files (.nii.gz) can still be recognized as `nii` file type.

Hope my explanation makes sense...

christinerogers:
perfectly, thanks.  i think i'm just forgetting what the BIDS standard includes in the alphabet soup...

PR #515, opened 2020-01-22
[python] Addition of a mass NIfTI pic creation script

cmadjar:
### Description

This adds a `mass_nifti_pic.py` script to allow mass creation of NIfTI files pic rendering stored in the `files` table.

### How to test

Insert a `.nii.gz` file into the `files` table and run `mass_nifti_pic.py` on the `FileID` associated to that file. 
Note: make sure you have all python dependencies install (check the install script for the list of `pip` packages that are required by LORIS-MRI)
```
python mass_nifti_pic.py -p database_config.py -s 10000 -l 10000 -v
```

### DIFF
--- a/python/mass_nifti_pic.py
+++ b/python/mass_nifti_pic.py
@@ -1,13 +1,13 @@
 #!/usr/bin/env python
 
-"""Script to mass chunk electrophysiology datasets."""
+"""Script to mass create the pic images of inserted NIfTI files."""
 
 import os
 import sys
 import getopt
 import lib.exitcode
-from lib.database      import Database
-from lib.physiological import Physiological
+from lib.database import Database
+from lib.imaging  import Imaging
 
 
 __license__ = "GPLv3"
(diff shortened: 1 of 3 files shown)


cmadjar:
@xlecours Before the holidays, we talked about a script that would create the pic rendering of NIfTI files. The script in that PR will do exactly that. If I remember correctly, you need this for one project so I assigned the review to you. Let me know if you have any question. Thanks!

xlecours:
That is a huge step toward uploading BIDS datasets. Thank you :)

python/mass_nifti_pic.py
```diff
@@ -0,0 +1,202 @@
+#!/usr/bin/env python
+
+"""Script to mass create the pic images of inserted NIfTI files."""
+
+import os
+import sys
+import getopt
```
cmadjar: import re

jesscall:
I cannot attest to the code because I am not familiar with python, but manual test is good :+1:

python/mass_nifti_pic.py
```diff
@@ -0,0 +1,203 @@
...
+        return
+
+    # grep the time length from the NIfTI file header
+    is_4d_dataset = False
+    length_parameters = imaging.get_nifti_image_length_parameters(data_dir + nii_file_path)
+    if len(length_parameters) == 4:
+        file_parameters['time'] = length_parameters[3]
+        is_4d_dataset = True
+
+    # grep the SessionID of the file
```
xlecours: CandID, not SessionID

cmadjar: @xlecours corrected the comment in the code. Thanks!

PR #516, opened 2020-01-24
[config update] profileTemplate cleanup

cmadjar:
### Description

Both CCNA and IBIS set an `isPhantom` key in the `getSubjectIDs` function of the config file to specify if the scan is a phantom based on the PatientName. [The code refers to that key](https://github.com/aces/Loris-MRI/blob/b38e7c0af96a541632e3bdad14d9f1083daee43e/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm#L1789) when checking if a scan is a phantom or an actual candidate session. The `profileTemplate.pl` was not updated with these changes unfortunately so this PR takes care of it.

Also removed the `isPhantom` function of the config file as it is not used anywhere in the scripts.

### Caveat for existing project with phantom datasets

Make sure to update your config file with those changes if it was not done already.


mohadesz:
LGTM

PR #519, opened 2020-01-24
Removed leftover occurrences of requiresStaging 

cmadjar:
### Description

requiresStaging was removed from the backend pipeline but somehow, two instances remained in the code. This PR cleans that up and completely removes it from the codebase.

### DIFF
--- a/uploadNeuroDB/imaging_non_minc_insertion.pl
+++ b/uploadNeuroDB/imaging_non_minc_insertion.pl
@@ -461,7 +461,7 @@ if ($subjectIDsref->{'CandMismatchError'}){
 }
 
 # determine the session ID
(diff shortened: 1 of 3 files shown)


cmadjar:
@nicolasbrossard Thank you! This is updated!

uploadNeuroDB/imaging_non_minc_insertion.pl
```diff
@@ -461,7 +461,7 @@ =head2 Methods
 }
 
 # determine the session ID
-my ($session_id, $requiresStaging) = NeuroDB::MRI::getSessionID(
+my ($session_id) = NeuroDB::MRI::getSessionID(
```
nicolasbrossard: I think `NeuroDB::MRI::getSessionID` requires 5 arguments. You only supply 4...Might happen elsewhere in the code too.

PR #521, opened 2020-01-29
[python] fixed the wait the value was encoded in the events file for electrophysiology datasets

cmadjar:
### Description

The BIDS events.tsv header names for physiological files have changed from `events_value` to `value` and `event_sample` to `sample` according to the latest BIDS release. This PRs implements the new header names and supports both conventions.


cmadjar:
Tested on the newest Cuban dataset and the fix works well so merging it

Issue #526, opened 2020-03-12
remove -new_scanner script option and do this by default


cmadjar:
Discussed at the imaging meeting of January 31st, 2020.

Will remove the `-new_scanner` option from the script calls. Now, every time a script encounters data from a new scanner, a new scanner will automatically be created.

PR #528, opened 2020-03-12
Remove new_scanner option from the pipeline

cmadjar:
### Description

These changes removes the option `-new_scanner` from the pipeline scripts since by default all projects register new scanners when they are not already present in the `mri_scanner` table.

### Link to issues or PR

#526

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -935,7 +935,7 @@ sub mapDicomParameters {
 }
 =pod
 
-=head3 findScannerID($manufacturer, $model, $serialNumber, $softwareVersion, $centerID, $dbhr, $register_new, $db)
+=head3 findScannerID($manufacturer, $model, $serialNumber, $softwareVersion, $centerID, $dbhr, $db)
 
 Finds the scanner ID for the scanner as defined by C<$manufacturer>, C<$model>,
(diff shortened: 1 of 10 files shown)


uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -969,9 +968,6 @@ sub findScannerID {
     
     # Scanner exists
     return $resultsRef->[0]->{'ID'} if @$resultsRef;
-
-    # Scanner does not exist and we don't want to register a new one: ID defaults to 0
-    return 0 if !$register_new;
     
     # only register new scanners when told to do so !!!
```
nicolasbrossard: Maybe change comment here to reflect the fact that we are always registering the new scanner now.

cmadjar:
@nicolasbrossard Thank you for the review! Your comments have been addressed, please re-review :).

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -551,16 +548,14 @@ sub determinePSC {
 
 =pod
 
-=head3 determineScannerID($tarchiveInfo, $to_log, $centerID, $NewScanner, $upload_id)
+=head3 determineScannerID($tarchiveInfo, $to_log, $centerID, $upload_id)
```
nicolasbrossard: Maybe indicate in the method doc that a new scanner will automatically be created if it does not exist?

Issue #527, opened 2020-03-12
remove -globLocation option in the scripts


cmadjar:
Discussed at the imaging meeting of January 31st.

Since the `-globLocation` flag is always turned on when calling all the script and since the archiveLocation is always a relative path, there is no need for that option anymore so we need to refactor the scripts to get rid of that option.

PR #530, opened 2020-03-17
Remove -globLocation option

cmadjar:
### Description

These changes removes the option `-globLocation` from the pipeline scripts since now the `archiveLocation` is always a relative path, hence our scripts are always being used with the `-globLocation` turned on. This gets rid of the option and keep the behaviour in the pipeline as if the option was always given (a.k.a. querying the `tarchive` table with ` WHERE ArchiveLocation LIKE basename($tarchive)`

### Link to issues or PR

#527 

### Caveat for existing projects

Ensure to remove the `-globLocation` option from any call to the scripts that used to have that option in your own scripts or cron jobs.

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
+++ b/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
@@ -440,14 +440,12 @@ sub determineSubjectID {
 
 =pod
 
-=head3 createTarchiveArray($tarchive, $globArchiveLocation)
+=head3 createTarchiveArray($tarchive)
 
 Creates the DICOM archive information hash ref.
 
 INPUTS:
   - $tarchive           : tarchive's path
-  - $globArchiveLocation: globArchiveLocation argument specified when running
-                           the insertion scripts
 
 RETURNS: DICOM archive information hash ref
 
@@ -457,11 +455,8 @@ sub createTarchiveArray {
 
     my $this = shift;
     my %tarchiveInfo;
-    my ($tarchive,$globArchiveLocation) = @_;
-    my $where = "ArchiveLocation='$tarchive'";
-    if ($globArchiveLocation) {
-        $where = "ArchiveLocation LIKE '%".basename($tarchive)."'";
-    }
+    my ($tarchive) = @_;
+    my $where = "ArchiveLocation LIKE '%" . quotemeta(basename($tarchive)) . "%'";
     my $query = "SELECT PatientName, PatientID, PatientDoB, md5sumArchive,".
                 " DateAcquired, DicomArchiveID, PatientSex,".
                 " ScannerManufacturer, ScannerModel, ScannerSerialNumber,".
--- a/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
+++ b/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
@@ -454,7 +454,8 @@ sub createTarchiveArray {
     my $this = shift;
     my %tarchiveInfo;
     my ($tarchive) = @_;
-    my $where = "ArchiveLocation LIKE '%" . quotemeta(basename($tarchive)) . "%'";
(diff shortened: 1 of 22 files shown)


uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -457,11 +453,8 @@ sub createTarchiveArray {
 
     my $this = shift;
     my %tarchiveInfo;
-    my ($tarchive,$globArchiveLocation) = @_;
-    my $where = "ArchiveLocation='$tarchive'";
-    if ($globArchiveLocation) {
-        $where = "ArchiveLocation LIKE '%".basename($tarchive)."'";
-    }
+    my ($tarchive) = @_;
+    my $where = "ArchiveLocation LIKE '%" . quotemeta(basename($tarchive)) . "%'";
```
nicolasbrossard: So if `$tarchive` is `/my/path/to/tarchive` and `ArchiveLocation` is `/x/y/z/PREFIXtarchiveSUFFIX` then where condition will evaluate to true, which is wrong IMHO.

How about

```
my $where = "CONCAT('/', ArchiveLocation) LIKE '%/" . quotemeta(basename($tarchive)) ."'";
```

The `CONCAT` makes sure that ArchiveLocation *always* contains a slash. Also, we should use a prepared statement so the where will contain a placeholder instead of `quotemeta(basename($tarchive))`

uploadNeuroDB/NeuroDB/objectBroker/MriUploadOB.pm
```diff
@@ -113,20 +109,19 @@ RETURNS: a reference to an array of array references. If C<$isCount> is true, th
...
 
-    my $select = $isCount         ? 'COUNT(*)'   : join(',', @MRI_UPLOAD_FIELDS);
-    my $where  = $isBaseNameMatch ? 'LIKE ?' : '=?';
+    my $select = $isCount ? 'COUNT(*)' : join(',', @MRI_UPLOAD_FIELDS);
 
     my $query = "SELECT $select "
         .       "FROM mri_upload "
         .       "JOIN tarchive USING(TarchiveID) "
-        .       "WHERE ArchiveLocation $where ";
+        .       "WHERE ArchiveLocation LIKE ? ";
```
nicolasbrossard: See comment made in `MRIProcessingUtility` and apply the same strategy.

uploadNeuroDB/NeuroDB/objectBroker/TarchiveOB.pm
```diff
@@ -121,15 +119,14 @@ sub getByTarchiveLocation {
     }
 
     my $query = sprintf(
-        "SELECT %s FROM tarchive WHERE ArchiveLocation %s",
+        "SELECT %s FROM tarchive WHERE ArchiveLocation LIKE ? ",
```
nicolasbrossard: See comment made in `MRIProcessingUtility` and apply the same strategy.

uploadNeuroDB/minc_insertion.pl
```diff
@@ -342,21 +332,16 @@ =head2 Methods
...
 } elsif ($tarchive) {
     # if only the tarchive path is given as an argument, find the associated UploadID
     # and check if IsTarchiveValidated is set to 1.
     $ArchiveLocation = $tarchive;
     $ArchiveLocation    =~ s/$tarchiveLibraryDir\/?//g;
 
-    my $where = "WHERE ArchiveLocation='$tarchive'";
-    if ($globArchiveLocation) {
-        $where = "WHERE ArchiveLocation LIKE '%/" . quotemeta(basename($tarchive)) . "' "
+    my $where = "WHERE ArchiveLocation LIKE '%/" . quotemeta(basename($tarchive)) . "' "
```
nicolasbrossard: See comment made in `MRIProcessingUtility` and apply the same strategy.

PR #533, opened 2020-03-27
[bug fix] Fixed Can't call method 'prepare' on unblessed reference at MRI.pm line 150

cmadjar:
### Description

this fixes the following error that appears when running the pipeline:
```
Can't call method "prepare" on unblessed reference at /data/qpn/bin/mri/uploadNeuroDB/NeuroDB/MRI.pm line 150.
```

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -147,7 +147,7 @@ sub getSessionInformation {
            . "WHERE CandID=? "
            . "AND LOWER(Visit_label)=LOWER(?) "
            . "AND Active='Y'";
-    $sth = $dbh->prepare($query);
+    $sth = $$dbh->prepare($query);
     $sth->execute($subjectIDref->{'CandID'}, $subjectIDref->{'visitLabel'});
 
     ##### if it finds an existing session it does this:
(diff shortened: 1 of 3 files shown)


uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -147,7 +147,7 @@ sub getSessionInformation {
            . "WHERE CandID=? "
            . "AND LOWER(Visit_label)=LOWER(?) "
            . "AND Active='Y'";
-    $sth = $dbh->prepare($query);
+    $sth = $$dbh->prepare($query);
```
cmadjar: this one is the real bug fix for the message

uploadNeuroDB/tarchiveLoader.pl
```diff
@@ -591,7 +591,7 @@ =head2 Methods
     #### link the tarchive and mri_upload table  with session ##
     ############################################################
     my ($sessionRef, $errMsg) = NeuroDB::MRI::getSessionInformation(
-        $subjectIDsref, $tarchiveInfo{'DateAcquired'}, $dbh, $db
+        $subjectIDsref, $tarchiveInfo{'DateAcquired'}, \$dbh, $db
```
cmadjar: I happen to find this one too when trying to resolve the bug and since it is related to the same function I included it in the same PR.

PR #534, opened 2020-03-27
[bug_fix] Fix pipelines errors when running the pipeline on a Mac OS X sandbox

cmadjar:
### Description

Fixed several errors that occurs when running the imaging pipeline on a Mac sandbox, including some Mac install updates. 

Note, the changes made do not change the behaviour of the pipeline on a Linux computer.

### DIFF
--- a/dicom-archive/dicomTar.pl
+++ b/dicom-archive/dicomTar.pl
@@ -200,10 +200,10 @@ my $system          = `uname`;
 
 
 # Remove all files starting with . in the dcm_source directory
-my $cmd = "cd " . $dcm_source . "; find -type f -name '.*' | xargs rm -f";
+my $cmd = "cd " . $dcm_source . "; find . -type f -name '.*' | xargs rm -f";
 system($cmd);
 # Remove __MACOSX directory
-my $cmd = "cd " . $dcm_source . "; find -name '__MACOSX' | xargs rm -rf";
+$cmd = "cd " . $dcm_source . "; find . -name '__MACOSX' | xargs rm -rf";
 system($cmd);
 
 # create new summary object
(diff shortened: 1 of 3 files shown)


uploadNeuroDB/NeuroDB/ImagingUpload.pm
```diff
@@ -189,7 +189,7 @@ sub IsCandidateInfoValid {
     ####Get a list of files from the folder#####################
     #############Loop through the files#########################
     ############################################################
-    my $cmd = "find -path " . quotemeta($this->{'uploaded_temp_folder'}) . " -name '__MACOSX' -delete ";
+    my $cmd = "find " . quotemeta($this->{'uploaded_temp_folder'}) . " -name '__MACOSX' -delete ";
```
cmadjar: the `-path` option of find is useful when you are looking for a filename containing `/` as the option `-name` gets rid of all `/` before the search. In this case, the string searched is `__MACOSX` so there is no need to use the option `-path` which does not exist in the default `find` command installed on Mac OS X systems.

dicom-archive/dicomTar.pl
```diff
@@ -200,10 +200,10 @@ =head2 Methods
 
 
 # Remove all files starting with . in the dcm_source directory
-my $cmd = "cd " . $dcm_source . "; find -type f -name '.*' | xargs rm -f";
+my $cmd = "cd " . $dcm_source . "; find . -type f -name '.*' | xargs rm -f";
 system($cmd);
 # Remove __MACOSX directory
-my $cmd = "cd " . $dcm_source . "; find -name '__MACOSX' | xargs rm -rf";
+$cmd = "cd " . $dcm_source . "; find . -name '__MACOSX' | xargs rm -rf";
```
nicolasbrossard: I suggest doing this in one command:

```
my @args = ($dcm_source);
push(@args, qw/-type -f -name __MACOSX -o -name .* -delete/);
system('find', @args);
```

Note that I am using a form of `system` call that bypasses shell interpolation, so you don't have to quote `$dcm_source` (it will be interpreted *as is* by the command).

PR #542, opened 2020-04-08
Missing notifications when the pipeline cannot create a visit.

### LINKED ISSUE #541: [minc_insertion.pl] Missing entries in notification_spool
When the pipeline attempts to create a visit in order to insert a scan and the visit could not be created because the project ID/subproject ID specified in the `prod` file are invalid, there should be an entry in table `notification_spool` indicating exactly why the session was not created (project ID invalid, no such subproject X for project Y, etc...) There are currently no such logs.

nicolasbrossard:
When the pipeline tries to create a visit to insert a scan and cannot do so because either the project ID or subproject ID in the `prod` file are invalid, there should be an entry in table `notification_spool` indicating so.

# To test

1. Edit your `prod` file and make sure that `createVisitLabel` is set to 1. 
2. In your `prod` file, Specify a `SubprojectID` that does not exist for your project
3. Try to insert a scan for a candidate/visit such that there are no entries in table `session` for that candidate/visit.
 
Ensure that once the pipeline finishes, there is a notification in table `notification_spool` stating that the projectID/subprojectID specifications are invalid and the session could not be created.

Fixes #541

### DIFF
--- a/uploadNeuroDB/minc_insertion.pl
+++ b/uploadNeuroDB/minc_insertion.pl
@@ -541,11 +541,6 @@ my($sessionRef, $errMsg) = NeuroDB::MRI::getSessionInformation(
     $db
 );
 
-# Copy the session info into the %$subjectIDsref hash array
-$subjectIDsref->{'SessionID'}    = $sessionRef->{'ID'};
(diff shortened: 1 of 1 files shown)


cmadjar:
Nice catch!!

PR #550, opened 2020-06-23
[python] fix bug when visits have different modalities

cmadjar:
There is a bug in the current code for the BIDS import script that does not allow sessions to import all modalities when session have different modalities.

Example of a dataset with different modalities per visits:
```
sub-01
  |__ ses-01
         |__ func
         |__ fmap
  |__ ses-02
         |__ func
         |__ fmap
  |__ ses-03
          |__ anat
```
In the case of such a dataset, only `anat` modalities would be inserted before the bug fix (the content of the latest visit)...

This PR fixes this bug and allows for the insertion of all modalities into the database even when sessions do not have the same modalities.

### DIFF
--- a/python/lib/bidsreader.py
+++ b/python/lib/bidsreader.py
@@ -201,18 +201,21 @@ class BidsReader:
         cand_session_modalities_list = []
 
         for subject, visit_list in self.cand_sessions_list.items():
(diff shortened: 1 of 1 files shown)


cmadjar:
Tested by Trevor who reported the bug. All is working so merging the PR.

PR #551, opened 2020-07-21
Get rid of master references in the docs to use the main branch instead

cmadjar:
This gets rid of references to the `master` branch from the LORIS-MRI documentations and uses the new defaults `main` branch instead.

The README instruction was also modified in favour of installing releases instead of cloning `master` as users should use official releases instead of the development branch. This mimics the behaviour from the LORIS repo.


cmadjar:
Thank you @nicolasbrossard !

PR #552, opened 2020-08-05
Add iEEG support to the import pipeline

h-karim:
This PR adds support to the pipeline by:
1. Adding a Python iEEG class to query iEEG data from a bids directory
2. adds ieeg entris to the `bids.json` file
3. Fixes a small accessing mistake in the `scanstsv.py` class 

**Note:** While the minimum iEEG support is provided, this PR does not add support to import [electrical stimulation files,](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/04-intracranial-electroencephalography.html#electrical-stimulation) which are exclusive to iEEG.


python/bids_import.py
```diff
@@ -216,6 +216,20 @@ def read_and_insert_bids(bids_dir, config_file, verbose, createcand, createvisit
                     loris_bids_root_dir    = loris_bids_root_dir
                 )
 
+            elif modality == 'ieeg':
+                Ieeg(
```
christinerogers, line 220: any chance this could be all lowercase instead of `Ieeg` ?

python/lib/bids.json
```diff
@@ -81,11 +81,11 @@
...
         "sub-{subject}[/ses-{session}]/func/sub-{subject}[_ses-{session}]_task-{task}[_acq-{acquisition}][_rec-{reconstruction}][_run-{run}][_echo-{echo}][_recording-{recording}]_{suffix<physio|stim>}.{extension<tsv\\.gz|json}",
-        "sub-{subject}[/ses-{session}]/meg/sub-{subject}[_ses-{session}]_task-{task}[_acq-{acquisition}][_run-{run}][_proc-{proc}]_meg.{extension|json}",
-        "sub-{subject}[/ses-{session}]/meg/sub-{subject}[_ses-{session}]_task-{task}[_acq-{acquisition}][_run-{run}][_proc-{proc}]_{suffix<channels>}.{extension<tsv>|tsv}",
-        "sub-{subject}[/ses-{session}]/meg/sub-{subject}[_ses-{session}]_task-{task}[_acq-{acquisition}]_{suffix<coordsystem>}.json",
-        "sub-{subject}[/ses-{session}]/meg/sub-{subject}[_ses-{session}]_task-{task}[_acq-{acquisition}]_{suffix<photo>}.jpg"
+        "sub-{subject}[/ses-{session}]/meg|ieeg/sub-{subject}[_ses-{session}]_task-{task}[_acq-{acquisition}][_run-{run}][_proc-{proc}]_meg|ieeg.{extension|json}",
```
christinerogers, line 86: can you clarify - why was it necessary to do this for ieeg but not eeg?

python/lib/scanstsv.py
```diff
@@ -68,7 +68,7 @@ def get_acquisition_time(self):
         """
 
         if 'acq_time' in self.tsv_headers:
-            eeg_acq_time = self.tsv_entries['acq_time']
+            eeg_acq_time = self.tsv_headers['acq_time']
```
christinerogers, line 71: @cmadjar 's review will be interesting to have on this.

python/lib/ieeg.py
```diff
@@ -0,0 +1,803 @@
...
+from lib.bidsreader    import BidsReader
+from lib.scanstsv      import ScansTSV
+
+
+__license__ = "GPLv3"
+
+
+class Ieeg:
+    """
+    This class reads the BIDS iEEG data structure and register the iEEG datasets
```
christinerogers, line 24:
```suggestion
    This class reads the BIDS iEEG data structure and registers the iEEG datasets
```

christinerogers: typo

python/bids_import.py
h-karim: It could, hwever since it's a class, the convention is to have it initialized in upper case, see [this line](https://github.com/aces/Loris-MRI/blob/main/python/bids_import.py#L220) for example. Let me know if you would still prefer to make it all lowercase.

python/lib/bids.json
h-karim: This change was based on the recommendation of [this issue](https://github.com/bids-standard/pybids/issues/583#issue-560546138). Given that this config file is used to detect patterns associated with a certain modality, I don't think the change will have any regressions.

christinerogers: i see! How lovely that [the pyBids issue pinpointing the ieeg gap](https://github.com/bids-standard/pybids/issues/583#issue-560546138) also pinpointed a solution/suggestion that was validated by Chris Markiewicz.

h-karim:
@christinerogers I added a little note mentioning ieeg support and the caveat with electrical stimulation files in BIDS. Let me know if you would like more information, or if it's better off mentioned someplace else in the documentation.

docs/02-Install.md
```diff
@@ -391,6 +391,9 @@ The pipeline to insert BIDS datasets into LORIS currently support the
 following BIDS modalities/entities:
   - Electroencephalography ('eeg')
   - Magnetic Resonance Imaging ('anat', 'func', 'fmap', 'dwi') 
+  - Intracranial electroencephalography ('ieeg')
+      - **Note:** optional [electrical stimulation](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/04-intracranial-electroencephalography.html#electrical-stimulation) files for ieeg modality are currently not supported in LORIS-MRI (or LORIS). 
+      If a BIDS-IEEG directory includes these files, they will be ignored during the import process.
```
cmadjar, line 396: is there a reason why this is not supported? Just wondering...

python/lib/scanstsv.py
cmadjar:
```suggestion
            eeg_acq_time = self.tsv_entries['acq_time']
```

`self.tsv_entries` contains the values present in the TSV file while `self.tsv_header` contains only the header. So here, instead of returning the actual date of acquisition, it will return the header label, a.k.a. `acq_time`

Note: both `self.tsv_entries` and `self.tsv_header` are defined in the `__init__` function of the class and available across all functions of the class.

python/bids_import.py
cmadjar, line 220: I compared the `Ieeg` and `Eeg` classes and I did not notice any major code changes (aside from renaming variables and documentation to `Ieeg` instead of `eeg`).

Therefore, I don't see the need to create a new `Ieeg` class when it is identical to the `Eeg` class. Instead, I would just call the `Eeg` class here.

If there are some specificities I missed for `Ieeg`, then, this could be resolved with if statements in the class and checking the modality in the `Eeg` class.

docs/02-Install.md
christinerogers: @cmadjar we didn't have any initial data to work with, and it's not in scope of the current project. 

@h-karim now that you have the basic PR in progress, can you research and outline a plan for adding electrostim - what changes/scripts are needed to support it?

christinerogers:
> @christinerogers I added a little note mentioning ieeg support and the caveat with electrical stimulation files in BIDS. Let me know if you would like more information, or if it's better off mentioned someplace else in the documentation.

yes please - see my other comment in cecile's review thread about a plan for electrostim coverage

cmadjar:
@h-karim you mentioned there was a necessary change to pybids, could you document it here?
 Sorry, just thought about this now...

h-karim:
@cmadjar the change was specifically to the `bids.json` file

> @h-karim you mentioned there was a necessary change to pybids, could you document it here?
> Sorry, just thought about this now...
@cmadjar 
The changes where specifically to the [`bids.json](https://github.com/aces/Loris-MRI/blob/8f6d681c0d41eb214abd5d37a6ee838bb938dc74/python/lib/bids.json) file. I had to include mentions of `ieeg` for it to detect the modality.

python/lib/scanstsv.py
cmadjar: right actually the `if` statement should be changed to that instead:

```suggestion
        if 'acq_time' in self.acquisition_data:
            eeg_acq_time = self.acquisition_data['acq_time']
```

Just tested it locally and it would work for tsv files with several entries as well as with TSV files with one entry only. 

Congratulations, you found a bug! ;)

cmadjar: @h-karim forgot to tag you

h-karim: Done!

PR #558, opened 2020-10-02
Populating Project/Subproject/Current_Stage in Loris DB

laemtl:
Populates additional fields in the db, for an EEG/IEEG entry to show in the frontend:
-  ‘registrationProjectID’ (candidate table),
- ‘projectID’  (session table) 
- ‘subProjectID’ (session table) 

Participants.tsv contains additional columns used to extract the data:

```
participant_id  age     sex     hand    site    subproject      project
sub-DCC0001     n/a     n/a     n/a     DCC     IEEG    IEEG

``` 
Note that subproject reference subproject.title and project references Project.Alias.

participants.json

```
{
    "participant_id": {
        "Description": "Unique participant identifier"
    },
    "age": {
        "Description": "Age of the participant at time of testing",
        "Units": "years"
    },
    "sex": {
        "Description": "Biological sex of the participant",
        "Levels": {
            "F": "female",
            "M": "male"
        }
    },
    "hand": {
        "Description": "Handedness of the participant",
        "Levels": {
            "R": "right",
            "L": "left",
            "A": "ambidextrous"
        }
    },
    "site": {
        "Description": "Site of the testing",
        "Levels": {
            "DCC": "Data Coordinating Center",
            "MNI": "Montreal Neurological Intitute"
        }
    },
    "subproject": {
        "Description": "Subproject of the participant",
        "Levels": {
            "IEEG": "IEEG Atlas"
        }
    },
    "project": {
        "Description": "Project of the participant",
        "Levels": {
            "IEEG": "IEEG Atlas"
        }
    }
}
```
If not provided, NULL is inserted in the DB and later addition is required in order to display the entry on the frontend.


python/2020-10-02-default_bids_stage_Config.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'default_bids_stage', 'Default stage to use in session when importing EEG/IEEG data from BIDS dataset', 1, 0, 'text', ID, 'Default stage for BIDS dataset', 25 FROM ConfigSettings WHERE Name="imaging_pipeline";
```
christinerogers, line 1: maybe this should be "default import stage" ? @cmadjar is this possibly relevant as an argument for the imaging uploader/pipeline too?

python/lib/session.py
```diff
@@ -66,10 +75,34 @@ def create_session(self, db):
...
+            []
+        )
+
+        if(len(current_stage_info) > 0):
+            current_stage = current_stage_info[0]['Value']
+
+        if current_stage:
+            column_names = column_names + ('Current_stage',)
+            values = values + (current_stage,)
+
```
christinerogers, line 101: beyond setting the session table field, should this trigger the battery population in Loris, @cmadjar ? pros/cons

christinerogers (commented):
wondering about whether the import should trigger the battery population? (comment for cecile inline)

python/lib/session.py
cmadjar: that should have the exact same behaviour as the Perl insertion pipeline, meaning setting the stage to "Not Started" and not populating the battery. That should be done by the data entry side of LORIS.

python/lib/session.py
```diff
@@ -66,10 +75,34 @@ def create_session(self, db):
...
+            column_names = column_names + ('ProjectID',)
+            values = values + (str(self.project_id),)
+
+        if self.subproject_id:
+            column_names = column_names + ('SubprojectID',)
+            values = values + (str(self.subproject_id),)
+        
+        current_stage = None
+        current_stage_info = db.pselect(
+            "SELECT Value FROM Config WHERE ConfigID = (SELECT ID FROM ConfigSettings WHERE Name='default_bids_stage')",
```
cmadjar, line 91: Should be setting that by default to "Not Started" removing the need of the config setting and this SQL query

cmadjar (changes requested):
@laemtl sorry for the delay in reviewing this. 

A few general comments:
- by default, the insertion pipeline should set the visit stage to "Not Started" (to mimic the behaviour of the perl insertion pipelines). Data entry staff should be the one to set the stage to something else when they start entering data.
- In general, SQL changes and patches should go to LORIS (although my comment above suppress the need for SQL changes in the case here)
- what happens if `project_id` cannot be found in `participants.tsv`? It looks like this is required for the pipeline to run. Two solutions here:
  - adding a check at the beginning of the pipeline so that if no project are associated to a candidate, it exits and print information on how to specify that information in the BIDS dataset
  - support project_id being null in the same way you did for subproject IDs

Let me know if you have questions or something I wrote was not clear.

And thank you for working on this!!! :-D

christinerogers:
thanks for your comments @cmadjar  - it was my request to see if we could trigger the battery population.  
We'll have to do that in a custom version of this pipeline to keep this commit consistent, should be fine.

cmadjar:
@christinerogers in that case, maybe the battery population could be a separate script that people can run. Probably more on the PHP side of things in the tools directory (if that does not already exist)?

christinerogers:
@cmadjar agreed it could be a follow-on script with 3 steps : get the Date of acquisition from the data, set the visit Stage and then run assign missing instruments (looking at tools/)

python/lib/session.py
laemtl: Good point. Is there an added value to have a NULL vs a Not Started stage? I wonder if it would not be even better to set 'Not Started' the default value of Current_Stage in the session table. It will have the advantage to not encode the value in the code, which can cause problem if the Current_Stage structure is updated.

laemtl:
@cmadjar There is no delay at all. Thank you for taking the time to review this.

* `In general, SQL changes and patches should go to LORIS`:
Totally agree, I posted them here to give a complete picture of the changes made, and also to give us a change to discuss them before suggesting those changes on Loris. So far, I think that was useful :)

* `what happens if project_id cannot be found in participants.tsv? It looks like this is required for the pipeline to run.`:
In python/lib/candidate.py, project_id is set by default to None, and is updated only if a value is found in participants.tsv. Otherwise L134), `RegistrationProjectID` remains NULL in candidate, and required manual edition later (similar to how the pipeline is behaving now)

python/lib/session.py
cmadjar: @laemtl Could you bring that up at the next LORIS meeting if it was not discussed yet? I would be in favour of having the default to `Not Started`, but there might be subtleties I don't see and would not mind other people opinion on this.

@ridz1208 @driusan Thoughts regarding changing the default value of Current_Stage from `NULL` to `Not Started`?

cmadjar:
@laemtl your PR on the LORIS side about the Current_stage has been merged. I think you can now remove the default current stage config from the code since by default, it will create the visit as Not Started now which should be the default behaviour.

Thanks! :)

laemtl:
@cmadjar @christinerogers Current_stage removed!

.gitignore
```diff
@@ -0,0 +1,4 @@
+**/__pycache__/*
+environment
```
cmadjar, line 2: @laemtl the environment file should not be in gitignore in case changes need to be made to that file. Instead, I would recommend creating your own environment file (environment_sandbox) and source that.

cmadjar (changes requested):
@laemtl I just tested your PR and it works great! There is only one last change to take care of and I am ready to merge. See my comment about the environment file. It should not be in the .gitignore file.

Let me know when this is removed. Thank you :)

.gitignore
laemtl: Oh... Good point.

laemtl:
@cmadjar Done!

cmadjar (approved):

PR #561, opened 2020-10-09
Implement strict checks for command line arguments of delete_imaging_upload.pl.

### LINKED ISSUE #559: Delete script: need more robust checks on the deleted tarchives
It is entirely possible to have records in table `mri_upload` that have the same TarchiveID. If an attempt is made to delete one (and only one) of these uploads, the delete script will issue an error message when it tries to delete the record in table `tarchive` since there are still uploads in table `mri_upload` tied to that archive. The script should detect that situation right at the start, issue an error message and abort.

nicolasbrossard:
This PR improves the validation done on the `upload_ID` argument of script `delete_imaging_upload.pl`. It will determine the `tarchive` associated to the uploads whose IDs are passed on the command line and check that *only* these uploads are associated to that archive and no other. An error message is issued if that is not the case. For example, say uploads `U1` and `U2` are tied to archive `A`. An attempt to delete only `U1` will fail: both `U1` and `U2` have to be deleted in the same command.

Resolves #559

### DIFF
--- a/docs/scripts_md/delete_imaging_upload.md
+++ b/docs/scripts_md/delete_imaging_upload.md
@@ -1,11 +1,11 @@
 # NAME
 
-delete\_mri\_upload.pl -- Delete everything that was produced (or part of what was produced) by the imaging pipeline for a given set of imaging uploads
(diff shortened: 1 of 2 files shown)


cmadjar (approved):
Finally tested the PR and it works great! Thank you @nicolasbrossard and sorry for the delay in getting that tested...

PR #563, opened 2020-11-06
Circumvent a dcm2mnc bug that causes modalities to be ignored in specific occasions.

nicolasbrossard:
This is a fix for the bug that was reported by Alfredo on the LORIS mailing list. See issue #564 for details.


cmadjar (approved):
LGTM. Tested and works as expected. Thank you @nicolasbrossard for the quick fix on this!

PR #568, opened 2020-11-19
[22.0][Bugfix] Undef EchoTime, RepTime and SliceThickness

jesscall:
Was having issues with inserting an image for DTI_TENSOR that had no values for EchoTime, RepTime and SliceThickness. 

This fix generalizes that any series type with empty values for those fields should be replaced by `undef`.


cmadjar (approved):
Thank you @jesscall :-)

PR #573, opened 2021-01-29
Fix create candidate bug in 22.0.0 when Config createCandidate is set to yes

cmadjar:
When the Config setting `createCandidates` is set to Yes, the pipeline always go into the `CreateMRICandidates` function of `MRIProcessingUtility.pm` and returns the following error:

```
Spool message is: ERROR: Cannot create candidate (MTL0001, 123456) as a candidate with PSCID=MTL0001 already exists
```

This adds an if statement before calling the function `CreateMRICandidates` to check first if a candidate with the PSCID already exists. If it exists, the call to the function `CreateMRICandidates` will be skipped.

NOTES: this bug fix does not apply to 23.0 since the code to determine candidate information has been revamped and is working well.

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
+++ b/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
@@ -416,9 +416,11 @@ sub determineSubjectID {
     );
 
     # create the candidate if it does not exist
-    $this->CreateMRICandidates(
-        $subjectIDsref, $tarchiveInfo, $User, $centerID, $upload_id
-    );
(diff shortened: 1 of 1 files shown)


cmadjar:
@nicolasbrossard if you have some time and can look at this PR. This resolves a bug that Alfredo had with 22. Note, the bug does not exist on 23 as that section of the code was revamped for that release. It would be just a bug fix for 22 for projects that create candidates via the insertion pipeline.

nicolasbrossard (approved):
LGTM. It might be a good idea to put something in the log if candidate creation is turned on and the candidate already exists (like "Skipping creation of MTL001/12345: candidate already exists"). Feel free to reject if you think that feature is not needed.

Issue #577, opened 2021-02-17
[bids_import][lib/mri.py]scan.json is wrongly parsed instead of scan.tsv in bids datasets


ycAbout:
In importing bids dataset, scan.json is wrongly parsed instead of scan.tsv.
A solution has been found. I will make a PR.

@gdevenyi

Traceback (most recent call last):
  File "/data/loris/bin/mri/python/bids_import.py", line 438, in <module>
    main()
  File "/data/loris/bin/mri/python/bids_import.py", line 75, in main
    read_and_insert_bids(bids_dir, config_file, verbose, createcand, createvisit)
  File "/data/loris/bin/mri/python/bids_import.py", line 237, in read_and_insert_bids
    loris_bids_root_dir    = loris_bids_root_dir
  File "/data/loris/bin/mri/python/lib/mri.py", line 137, in __init__
    self.register_raw_file(nifti_file)
  File "/data/loris/bin/mri/python/lib/mri.py", line 245, in register_raw_file
    inserted_nii  = self.fetch_and_insert_nifti_file(nifti_file)
  File "/data/loris/bin/mri/python/lib/mri.py", line 315, in fetch_and_insert_nifti_file
    scan_info = ScansTSV(self.scans_file, nifti_file.filename, self.verbose)
  File "/data/loris/bin/mri/python/lib/scanstsv.py", line 48, in __init__
    self.acquisition_data = self.find_acquisition_data()
  File "/data/loris/bin/mri/python/lib/scanstsv.py", line 59, in find_acquisition_data
    if os.path.basename(self.acquisition_file) in entry['filename']:
KeyError: 'filename'

PR #579, opened 2021-02-17
fix scan.json wrongly taken instead of scan.tsv

ycAbout:
This PR fix issue #577
In importing bids dataset, scan.json is wrongly parsed instead of scan.tsv

The solution has been consulted with @cmadjar , who provided input and shares the credit.
Please assign @cmadjar as one reviewer.

cc:
@gdevenyi


cmadjar (approved):
LGTM. Thanks @ycAbout :)

PR #582, opened 2021-02-17
Fixed wrong parameter name for scan acquisition date in imaging_non_minc_insertion.pl

### LINKED ISSUE #572: [imaging_non_minc_insertion] Nifti files not seen in imaging_browser
I inserted NIFTI images with script `imaging_non_minc_insertion.pl` and everything seemed to go fine but I cannot see my images listed in the landing page of the imaging browser module.

nicolasbrossard:
This bug was due to the fact that when the acquisition date is inserted in table `parameter_file`, the name of the parameter used is incorrect. This PR changes it to the correct value.

Fixes #572.


cmadjar (approved):
ooooops. Thank you for fixing this @nicolasbrossard !

PR #584, opened 2021-02-18
mass_nifti_pic.py does not handle file ID ranges well.

### LINKED ISSUE #583: mass_nifti_pic.py bug when using a range of file IDs
When using a range of FileIDs, say from F1 to F2, the script should create images for files ranging from file ID F1 to F2 (inclusive). It turns out the image for F2 is never generated. All the other images are created properly.

nicolasbrossard:
When using a file ID range with the `mass_nifti_pic.py` script, the image associated to the last (max) file ID is not generated. This PR corrects this.

Fixes #583

### DIFF
--- a/python/mass_nifti_pic.py
+++ b/python/mass_nifti_pic.py
@@ -68,7 +68,7 @@ def main():
     if (smallest_id == largest_id):
         make_pic(smallest_id, config_file, verbose)
     else:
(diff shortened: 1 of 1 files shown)


cmadjar (approved):
👍 Thank you @nicolasbrossard

cmadjar (approved):

Issue #586, opened 2021-02-18
[bids_import][lib/scantsv.py]acquisition time error in scantsv in importing bids datasets


ycAbout:
In importing bids dataset, scantsv.py is not able get acquisition time.
I am finding a solution, I will make a PR.

@gdevenyi

PR #588, opened 2021-02-18
fix bids import acquisition time error

ycAbout:
This PR fix issue #586
In importing bids dataset, scantsv.py is not able get acquisition time for mri.
The cause is that self.tsv_entries is a list of dict rather than a dict.

The solution: 
if it is a list, then
filtered out the unrelated dict objects for the current acquisition_file, and handle error in case of more than one entry exist. Then go from there.

cc:
@gdevenyi


cmadjar (approved):
Thank you @ycAbout !

PR #595, opened 2021-03-02
Fix bug occurring when multiple MRI pipeline processes are running simultaneously

### LINKED ISSUE #594: MRI pipeline fails when run in parallel by multiple processes
When the MRI pipeline performs a check on each file in an archive to see if it is a DICOM file, it first builds the list of files in the archive and writes it an a temporary file. The problem is that the path of this temporary file is always the same, irrespective of the MRI pipeline process that writes/uses it. Consequently, it can contain files that belong to multiple MRI pipeline processes if two or more are running concurrently.

nicolasbrossard:
When the MRI pipeline performs a check on each file in an archive to see if it is a DICOM file, it first builds the list of files in the archive and writes it an a temporary file. The problem is that the path of this temporary file is always the same, irrespective of the MRI pipeline process that writes/uses it. Consequently, it can contain files that belong to multiple MRI pipeline processes if two or more are running concurrently.

Fixes #594

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -1549,7 +1549,7 @@ sub isDicomImage {
     # not return one file per line but many files in one line. Writing in a
     # temporary file on which we run the command `cat` seems to be the only option
     # that works at the moment...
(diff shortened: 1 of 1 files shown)


cmadjar (approved):
👍

PR #600, opened 2021-03-15
Optimization of the delete imaging upload script.

nicolasbrossard:
Script `delete_imaging_upload.pl` takes an unusually long amount of time to complete when the number of entries in table `parameter_file` is large. This PR optimizes the query used to determine the rows to delete from table `parameter_file` so that running time is acceptable. Execution time going from an hour to a second were observed with this improvement!

### DIFF
--- a/tools/delete_imaging_upload.pl
+++ b/tools/delete_imaging_upload.pl
@@ -1046,40 +1046,56 @@ sub getParameterFilesRef {
     # Try finding a file in parameter_file that matches the -file argument
     my($query, @queryArgs);
     if($fileBaseName ne '' && !@{ $filesRef->{'files'} } && !@{ $filesRef->{'files_intermediary'} }) {
-        $query = 'SELECT ParameterFileID, FileID, Value, pt.Name FROM parameter_file pf '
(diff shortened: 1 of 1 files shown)


cmadjar (commented):
LGTM. @zaliqarosli could you test the delete script of this PR? On a session with defaced data and on a session without defaced data? 

Both situations exists in raisinbread.

Thanks!

zaliqarosli:
deleting and resourcing worked well for both defaced and non defaced sessions. however all the (raisinbread) sessions with defaced images have sql dumps that when resourced give the duplicate entry for primary key error reported in https://github.com/aces/Loris-MRI/pull/601. this still allowed for the upload ID to be restored in the imaging_upload module data table however. 

for both defaced and non defaces sessions, the files backups look good. in terms of optimization, i didn't see a drastic difference, but for 1 deletion, this PR knocked off a couple seconds for sure

nicolasbrossard:
@zaliqarosli Thanks for testing! Yep, there is indeed a bug with the delete script. See https://github.com/aces/Loris-MRI/issues/602.

cmadjar (approved):

PR #603, opened 2021-03-25
Invalid SQL backup file produced by script delete_imaging_upload.pl

### LINKED ISSUE #602: delete_imaging_upload produces incorrect SQL backup file
When script `delete_imaging_upload.pl` builds the list of rows of table `parameter_file` that it needs to delete, those that are associated to records of both `files` and files_intermediary` are listed twice. This causes 2 problems:

1. An attempt to delete these rows will be done twice. This is not so bad as the first delete will succeed and the second delete will be a no-op (i.e the statement is syntactically correct but does nothing).
2. When creating the SQL backup file, 2 `INSERT` statements will be created for all rows that were deleted twice. Consequently, when the SQL backup file is sourced in MySQL, it produces an error (attempt to insert records with the same primary key).

nicolasbrossard:
When trying to delete an upload on raisin bread that had associated records in table `files_intermediary`, the SQL backup file produced by the script contained multiple `INSERT` statements for some of the records in table `parameter_file`. Consequently, this file cannot be sourced directly in mysql to restore the deleted records.

Fixes #602

### DIFF
--- a/docs/scripts_md/delete_imaging_upload.md
+++ b/docs/scripts_md/delete_imaging_upload.md
@@ -30,7 +30,7 @@ Available options are:
                          `-backup_path` cannot be used if `-nofilesbk` and `-nosqlbk` are also used.
 
 \-basename fileBaseName : basename of the file to delete. The file is assumed to either exist in table `files` or table 
-                         `parameter_files`. This option should be used when targeting a specific (unique) file for deletion.
+                         `parameter_file`. This option should be used when targeting a specific (unique) file for deletion.
                          Note that the file will be deleted from both the database and the filesystem. This option cannot be 
                          used with options `-defaced` and `-form`.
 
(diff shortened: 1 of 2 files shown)


zaliqarosli (changes requested):
the sql output by this branch doesn't have the following insert into table statements when compared to on main:
- notification_spool
- tarchive_files
- tarchive_series
- any other potential tables that have insert statements appended at the end of the sql file after `parameter_file` insert statements

cmadjar:
@nicolasbrossard Will you have some time to integrate @zaliqarosli's  review on this PR? Just putting it back on your radar ;). Thanks!

cmadjar:
@nicolasbrossard Will you have some time to integrate @zaliqarosli's review on this PR? Just putting it back on your radar ;). Thanks!

zaliqarosli:
second round test:

i also could not reproduce what i reported before. deleting upload ID 90 worked well, SQL output had all those tables i reported missing before, and sourced well. im wondering if the issue i found before was due to some missing data in raisinbread that is no longer an issue

```
delete_imaging_upload.pl -profile prod -backup_path /data/backups/imaging_uploadid90_backup -uploadID 90
```

PR #604, opened 2021-03-26
Encoding and grep issues

laemtl:
This PR fixes 2 issues:

## BIDS parser:
- the tsv parser can't find participant_id (U+FEFFparticipant_id is read instead) when a byte-order mark (BOM) is present in participants.tsv (happened on 2 different instances)

## DICOM to MINC conversion
- for some reason the grep construct, which identifies DICOM files with identical `studyUID, seriesUID, echo and seriesDesc` was not working on a particular project: 0 file matched the pattern instead on the 321 files expected to match. Updating to use grep -P + \t fixes the issue.

### DIFF
--- a/python/lib/utilities.py
+++ b/python/lib/utilities.py
@@ -27,7 +27,7 @@ def read_tsv_file(tsv_file):
     """
 
     results = []
-    with open(tsv_file) as file:
+    with open(tsv_file, encoding='utf-8-sig') as file:
         reader = csv.DictReader(file, delimiter="\t")
 
         for row in reader:
--- a/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
+++ b/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
@@ -1347,7 +1347,7 @@ INPUTS:
 
 sub dicom_to_minc {
     my $this = shift;
-    my ($study_dir, $converter,$get_dicom_info, $exclude,$mail_user, $upload_id) = @_;
+    my ($study_dir, $converter, $get_dicom_info, $exclude,$mail_user, $upload_id) = @_;
     my ($d2m_cmd, $d2m_log, $exit_code, $excluded_regex);
     my $message = '';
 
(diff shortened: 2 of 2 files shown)


uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1387,7 +1387,7 @@ sub dicom_to_minc {
         # We need to use quotemeta here to make sure each character of a given variable is interpreted
         # as is and not as a regex special character
         # Note: '$\'\t\'' is a TAB for Unix grep
-        my $pattern = join('$\'\t\'', quotemeta($studyUID), quotemeta($seriesUID), quotemeta($echo), quotemeta($seriesDesc));
+        my $pattern = join('\t', quotemeta($studyUID), quotemeta($seriesUID), quotemeta($echo), quotemeta($seriesDesc));
```
cmadjar, line 1390: @nicolasbrossard do you agree with those changes? I remember you changed that section due to a bug in https://github.com/aces/Loris-MRI/commit/a38df855043c3833d7a94ea00df1f0d858f0e5e5

python/lib/utilities.py
```diff
@@ -27,7 +27,7 @@ def read_tsv_file(tsv_file):
     """
 
     results = []
-    with open(tsv_file) as file:
+    with open(tsv_file, encoding='utf-8-sig') as file:
```
cmadjar, line 30: @laemtl what happens if the file is not encoded utf-8-sig?

laemtl: The script fails (participant_id is not found in participants.tsv) since the field extracted is U+FEFFparticipant_id.

laemtl:
```
  File "bids_import.py", line 454, in <module>
    main()
  File "bids_import.py", line 75, in main
    read_and_insert_bids(bids_dir, config_file, verbose, createcand, createvisit)
  File "bids_import.py", line 158, in read_and_insert_bids
    bids_reader = BidsReader(bids_dir, verbose)
  File "/data/loris-mri/bin/mri/python/lib/bidsreader.py", line 62, in __init__
    self.participants_info = self.load_candidates_from_bids()
  File "/data/loris-mri/bin/mri/python/lib/bidsreader.py", line 114, in load_candidates_from_bids
    self.candidates_list_validation(participants_info)
  File "/data/loris-mri/bin/mri/python/lib/bidsreader.py", line 145, in candidates_list_validation
    row['participant_id'] = row['participant_id'].replace('sub-', '')
KeyError: 'participant_id'
```

```
> print(row)
OrderedDict([('\ufeffparticipant_id', 'sub-OTT174'), ('age', 'n/a'), ('sex', 'n/a'), ('hand', 'n/a')])
```

cmadjar:
@laemtl would it be possible to split that PR into two different PRs? Since they are touching completely different piece of code and the dcm2mnc changes might be a bit trickier, it would be best to split it. Thank you!

PR #605, opened 2021-04-06
Chunking script memory optimization

laemtl:
For huge edf files (problem observed for a 8GB file), the chunking script needs to allocate 30GB of memory and fails.
This is because the reader (parsed.get_data) loads all the data points at once. Loading the data points per channel helps to avoid a memory exhaustion error.


cmadjar (approved):
👍

PR #611, opened 2021-04-12
Grep issue

laemtl:
## DICOM to MINC conversion
ANSI-C quoting (e.g. $'/t') are not supported on all shells. On Ubuntu, the default shell on /bin/sh (used by default by Perl when running system) is dash and does not support it. Bash does. 

The grep construct, which identifies DICOM files with identical studyUID, seriesUID, echo and seriesDesc in `uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm` uses a ANSI-C quoting tab, and on some system the grep command fails: 0 files matched the pattern despite some match. Forcing perl system to execute the command on bash fixes the issue.

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
+++ b/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
@@ -1347,7 +1347,7 @@ INPUTS:
 
 sub dicom_to_minc {
     my $this = shift;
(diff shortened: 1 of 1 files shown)


laemtl:
Replace #604

nicolasbrossard (approved):
@laemtl I would only add a comment above the line where the command is executed to indicate that you are forcing execution with `bash` in order to support `$'\t'`.

laemtl:
@nicolasbrossard Done!

PR #612, opened 2021-04-15
Fix scan's IDs set to unknown but file named with a different ScanType

cmadjar:
Resolves an issue where some files where inserted into the `files` table with `AcquisitionProtocolID` linking to the `unknown` scan type but the File was actually named `<study>_<candID>_<visit>_F18tracerOSEM.mnc`, `F18tracerOSEM` being the correct scan type.

What happened is that `F18tracerOSEM` was not in `mri_scan_type` so the pipeline fetched the ID of the `unknown` scan type in `scan_type_text_to_id` but never updated the scan type to `unknown` leading to inconsistent linking into the database.

This refactors the `scan_type_text_to_id` function to remove the part where it looks for the unknown type as this should be done outside of the function.

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -639,7 +639,7 @@ INPUTS:
   - $type: scan type
   - $db  : database object
 
-RETURNS: ID of the scan type
+RETURNS: ID of the scan type or undef
 
 =cut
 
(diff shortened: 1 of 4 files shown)


cmadjar:
@laemtl I did some more testing on that and it seems to be working well on my end. Maybe when you have time you could see if there is still an issue when you test. Thanks!

cmadjar:
@laemtl just a little reminder that this PR is ready for another review when you have time. Thanks!

PR #617, opened 2021-05-11
Display more elaborate message in the error for comparison of subjects present in participants.tsv and the ones present in the BIDS layout

cmadjar:
This PR improves a little the logs of the error when doing comparison between the list of subjects present in participants.tsv and the list of subjects grepped by the BIDS reader.

### DIFF
--- a/python/lib/bidsreader.py
+++ b/python/lib/bidsreader.py
@@ -135,7 +135,7 @@ class BidsReader:
 
         mismatch_message = "\nERROR: Participant ID mismatch between " \
                            "participants.tsv and raw data found in the BIDS " \
-                           "directory/n"
+                           "directory\n"
 
         # check that all subjects listed in participants_info are also in
         # subjects array and vice versa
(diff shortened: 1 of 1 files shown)


cmadjar:
@laemtl If you have some time to review this one. Should be a relatively quick one.

python/lib/bidsreader.py
```diff
@@ -135,7 +135,7 @@ def candidates_list_validation(self, participants_info):
 
         mismatch_message = "\nERROR: Participant ID mismatch between " \
                            "participants.tsv and raw data found in the BIDS " \
```
vietdhoang, line 137:
```suggestion
        mismatch_message = ("\nERROR: Participant ID mismatch between "
                            "participants.tsv and raw data found in the BIDS "
```

python/lib/bidsreader.py
```diff
@@ -135,7 +135,7 @@ def candidates_list_validation(self, participants_info):
...
-                           "directory/n"
+                           "directory\n"
```
vietdhoang, line 138:
```suggestion
                            "directory")
```

vietdhoang (changes requested):
I added some suggestions to account for the fact that `print()` automatically adds a newline at the end.

I also added a suggestion for how a multi-line string can be formatted. This suggestion can be ignored if the current format is preferred or is the standard at LORIS

cmadjar:
@vietdhoang thank you! I committed the changes. :)

PR #608, opened 2021-04-08
EEG-BIDS pipeline - Derivatives support: annotations

laemtl:
This PR adds support for derivatives, specifically annotations.
Annotations are added using 2 files, *_annotations.json (metadata) and _annotations.tsv (data), both mandatory.
For a template:  ([tsv](https://github.com/bids-standard/bids-examples/blob/9893043f52ab1cec144d1b6069497ef48edc9570/eeg_face13/derivatives/BIDS-Lossless-EEG/sub-001/eeg/sub-001_task-faceFO_annotations.tsv)) ([json](https://github.com/bids-standard/bids-examples/blob/9893043f52ab1cec144d1b6069497ef48edc9570/eeg_face13/derivatives/BIDS-Lossless-EEG/sub-001/eeg/sub-001_task-faceFO_annotations.json)) 

To test 
The PR depends on an SQL patch from https://github.com/aces/Loris/pull/7345 

PyBIDS won't detect *_annotations.json and _annotations.tsv with validate turned on without the [force_index](https://bids-standard.github.io/pybids/generated/bids.layout.BIDSLayout.html) option, as it ignores files that do not follow a naming convention allowed by the specs. The annotation specs are currently discussed in[ BIDS Extension Proposal 21](https://docs.google.com/document/d/1PmcVs7vg7Th-cGC-UrX8rAhKUHIzOI-uIOh69_mvdlw/edit#heading=h.mqkmyp254xh6).

Related to #450

PR #625, opened 2021-05-17
BIDS import - handle multiple physiology files and derivatives

laemtl:
The current version of eeg.py handles the import of a single physiology file per subject/modality folder. This extends the current code to import all the physiology files present with their appropriate metadata.

-- note: 
~~The derivative case is broken until I revisit #608 ([code L162-165](https://github.com/aces/Loris-MRI/pull/608/files#diff-250dc8b14f8948bfc4cf003892b0da4257fc45681b3e49d5a7af615e552ea5bbR163) is already commented so it does not break anything).~~
The derivative case is fixed and needs testing.

### DIFF
--- a/python/lib/bidsreader.py
+++ b/python/lib/bidsreader.py
@@ -90,26 +90,23 @@ class BidsReader:
         if self.verbose:
             print('Loading the BIDS dataset with BIDS layout library...\n')
 
-        bids_config = os.environ['LORIS_MRI'] + "/python/lib/bids.json"
-        exclude_arr = ['/code/', '/sourcedata/', '/log/', '.git/']
-        force_arr   = re.compile("_annotations\.(tsv|json)$")
+        bids_config   = os.environ['LORIS_MRI'] + "/python/lib/bids.json"
+        exclude_arr   = ['/code/', '/sourcedata/', '/log/', '.git/']
+        force_arr     = [re.compile("_annotations\.(tsv|json)$")]
 
         # BIDSLayoutIndexer is required for PyBIDS >= 0.12.1
         bids_pack_version = list(map(int, bids.__version__.split('.')))
-        if (bids_pack_version[0] > 0
-            or bids_pack_version[1] > 12
-            or (bids_pack_version[1] == 12 and bids_pack_version[2] > 0)):
-            bids_layout = BIDSLayout(
-                root=self.bids_dir,
-                # disabled until is a workaround for https://github.com/bids-standard/pybids/issues/760 is found
-                # [file] bids_import.py [function] read_and_insert_bids [line] for modality in row['modalities']: (row['modalities'] is empty)
-                # indexer=BIDSLayoutIndexer(config_filename=bids_config, ignore=exclude_arr, force_index=force_arr)
-                config=bids_config,
-                ignore=exclude_arr,
-                force_index=force_arr
-            )
-        else:
-            bids_layout = BIDSLayout(root=self.bids_dir, config=bids_config, ignore=exclude_arr, force_index=force_arr)
+        # disabled until is a workaround for https://github.com/bids-standard/pybids/issues/760 is found
+        # [file] bids_import.py [function] read_and_insert_bids [line] for modality in row['modalities']: (row['modalities'] is empty)
+        #if (bids_pack_version[0] > 0
+        #    or bids_pack_version[1] > 12
+        #    or (bids_pack_version[1] == 12 and bids_pack_version[2] > 0)):
+        #    bids_layout = BIDSLayout(
+        #        root=self.bids_dir,
+        #        indexer=BIDSLayoutIndexer(config_filename=bids_config, ignore=exclude_arr, force_index=force_arr)
+        #    )
+        #else:
+        bids_layout = BIDSLayout(root=self.bids_dir, config=bids_config, ignore=exclude_arr, force_index=force_arr, derivatives=True)
 
         if self.verbose:
             print('\t=> BIDS dataset loaded with BIDS layout\n')
--- a/python/lib/bidsreader.py
+++ b/python/lib/bidsreader.py
@@ -287,7 +287,7 @@ class BidsReader:
     @staticmethod
     def grep_file(files_list, match_pattern, derivative_pattern=None):
         """
-        Grep a file based on a match pattern and returns it.
+        Grep a unique file based on a match pattern and returns it.
 
         :param files_list        : list of files to look into
          :type files_list        : list
@@ -297,11 +297,10 @@ class BidsReader:
                                    is a derivative file
          :type derivative_pattern: str
 
-        :return: name of the file that matches the pattern
+        :return: name of the first file that matches the pattern
          :rtype: str
         """
 
-        raw_file = None
         for filename in files_list:
             if not derivative_pattern:
                 if 'derivatives' in filename:
@@ -309,10 +308,10 @@ class BidsReader:
                     continue
                 elif re.search(match_pattern, filename):
                     # grep the file that matches the match_pattern (extension)
-                    raw_file = filename
+                    return filename
             else:
                 matches_derivative = re.search(derivative_pattern, filename)
                 if re.search(match_pattern, filename) and matches_derivative:
-                    raw_file = filename
+                    return filename
 
-        return raw_file
+        return None
--- a/python/lib/bidsreader.py
+++ b/python/lib/bidsreader.py
@@ -92,6 +92,7 @@ class BidsReader:
 
         bids_config = os.environ['LORIS_MRI'] + "/python/lib/bids.json"
         exclude_arr = ['/code/', '/sourcedata/', '/log/', '.git/']
+        force_arr   = re.compile("_annotations\.(tsv|json)$")
 
         # BIDSLayoutIndexer is required for PyBIDS >= 0.12.1
         bids_pack_version = list(map(int, bids.__version__.split('.')))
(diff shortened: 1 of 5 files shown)


cmadjar (changes requested):
@laemtl This is working well for the Face13 dataset that is on raisinbread as well as on fake datasets with multiple runs of EDF files.

When testing some of the bids-examples, I encountered two issues:
- An issue for support of SET and FDT files when testing https://github.com/bids-standard/bids-examples/tree/master/eeg_ds000117. See comment on code for line 401. 
- the inheritance of JSON file when the file is not in the eeg folder is not working. See comment on code for line 407.

cmadjar:
@laemtl final case test I did not see before:

- when there are two generic electrodes file in the eeg folder. For example:
```
sub-02_coordsystem.json
sub-02_task-faceO_electrodes.tsv
sub-02_task-faceO_run-2_eeg.edf
sub-02_task-faceO_run-2_events.tsv
sub-02_task-facerecognition_electrodes.tsv
sub-02_task-facerecognition_run-1_eeg.edf
sub-02_task-facerecognition_run-1_events.tsv
```

This one is a silent failure. Instead of inserting `task-facerecognition_electrodes.tsv` for the `task-facerecognition_run-1_eeg.edf`, the pipeline is inserting the file and data for `sub-02_task-faceO_electrodes.tsv`. Same thing for the events file. So the resulting structure inserted into LORIS is:
```
sub-02_ses-V1_task-faceO_electrodes.tsv
sub-02_ses-V1_task-faceO_run-2_eeg.edf
sub-02_ses-V1_task-faceO_run-2_eeg.tgz
sub-02_ses-V1_task-faceO_run-2_events.tsv
sub-02_ses-V1_task-facerecognition_run-1_eeg.edf
sub-02_ses-V1_task-facerecognition_run-1_eeg.tgz
```

Note: looks like the coordsystem.json file is never inserted either, which is a different issue than this PR.

cmadjar:
@laemtl I am adding the Needs work before merging label so I stop looking into that PR. Once you are done working on it, you can remove the label and tag me so I can re-review. Hope that works. Thanks!

laemtl:
@cmadjar Ready for second review. I addressed the derivatives case so I updated the PR title accordingly.

cmadjar:
@laemtl just FYI, there were conflicts with the main branch that I just resolved.

python/lib/eeg.py
```diff
@@ -238,262 +227,237 @@ def grep_bids_files(self, bids_type):
...
-            derivative_pattern = derivative_pattern
+        # TODO if derivatives, grep the source file as well as the input file ID???
+
+        # grep the raw files
+        eeg_files = self.bids_layout.get(
+            subject   = self.bids_sub_id,
+            session   = self.bids_ses_id,
+            scope     = 'derivatives' if derivatives else 'raw',
+            #datatype = self.bids_modality,
+            #suffix   = self.bids_modality,
```
cmadjar, line 315: why are those two commented?

python/lib/bidsreader.py
```diff
@@ -90,20 +88,23 @@ def load_bids_data(self):
...
+        # [file] bids_import.py [function] read_and_insert_bids [line] for modality in row['modalities']: (row['modalities'] is empty)
+        #if (bids_pack_version[0] > 0
+        #    or bids_pack_version[1] > 12
+        #    or (bids_pack_version[1] == 12 and bids_pack_version[2] > 0)):
+        #    bids_layout = BIDSLayout(
+        #        root=self.bids_dir,
+        #        indexer=BIDSLayoutIndexer(config_filename=bids_config, ignore=exclude_arr, force_index=force_arr)
+        #    )
+        #else:
+        bids_layout = BIDSLayout(root=self.bids_dir, config=bids_config, ignore=exclude_arr, force_index=force_arr, derivatives=True)
```
cmadjar, line 107: This does not work for datasets that have no derivatives. I get the following warnings trying to load the Face13 dataset with no derivatives:
```
/Users/cmadjar/Development/python_venv/python_intel_3.9.4/loris-mri-venv/lib/python3.9/site-packages/bids/layout/validation.py:131: UserWarning: Derivative indexing was requested, but no valid datasets were found in the specified locations (['/Users/cmadjar/Data/LORIS_imaging_data/incoming/EEG/Face13_small/derivatives']). Note that all BIDS-Derivatives datasets must meet all the requirements for BIDS-Raw datasets (a common problem is to fail to include a 'dataset_description.json' file in derivatives datasets).
Example contents of 'dataset_description.json':
{"Name": "Example dataset", "BIDSVersion": "1.0.2", "PipelineDescription": {"Name": "Example pipeline"}}
  warnings.warn("Derivative indexing was requested, but no valid "
```

python/lib/physiological.py
```diff
@@ -520,19 +522,27 @@ def insert_event_file(self, event_data, event_file, physiological_file_id,
...
             values_tuple = (
                 str(physiological_file_id),
                 row['onset'],
                 row['duration'],
                 row['trial_type'],
                 row['response_time'],
                 row['event_code'],
-                row['event_value']  or row['value'],
-                row['event_sample'] or row['sample'],
+                str(row['event_value']) or str(row['value']),
```
cmadjar, line 544: this is causing an error for datasets that do not have an event_value column in the events file (example: Face13 dataset that has only onset, duration and trial_type set in the TSV file). 

So in this case, row['event_value'] and row['value'] are set to `None` and when you convert it to `str` it tries to insert the string "None" in the `EventValue` field that should be a 10 digits integer causing the following failure:

```
  File "/Users/cmadjar/Development/Github/loris-mri/python/lib/database.py", line 185, in insert
    cursor.executemany(query, values)
  File "/Users/cmadjar/Development/python_venv/python_intel_3.9.4/loris-mri-venv/lib/python3.9/site-packages/MySQLdb/cursors.py", line 230, in executemany
    return self._do_execute_many(
  File "/Users/cmadjar/Development/python_venv/python_intel_3.9.4/loris-mri-venv/lib/python3.9/site-packages/MySQLdb/cursors.py", line 261, in _do_execute_many
    rows += self.execute(sql + postfix)
  File "/Users/cmadjar/Development/python_venv/python_intel_3.9.4/loris-mri-venv/lib/python3.9/site-packages/MySQLdb/cursors.py", line 206, in execute
    res = self._query(query)
  File "/Users/cmadjar/Development/python_venv/python_intel_3.9.4/loris-mri-venv/lib/python3.9/site-packages/MySQLdb/cursors.py", line 319, in _query
    db.query(q)
  File "/Users/cmadjar/Development/python_venv/python_intel_3.9.4/loris-mri-venv/lib/python3.9/site-packages/MySQLdb/connections.py", line 259, in query
    _mysql.connection.query(self, query)
MySQLdb._exceptions.OperationalError: (1366, "Incorrect integer value: 'None' for column `loris_rb`.`physiological_task_event`.`EventValue` at row 1")

During handling of the above exception, another exception occurred:

  File "/Users/cmadjar/Development/Github/loris-mri/python/bids_import.py", line 463, in <module>
    main()
  File "/Users/cmadjar/Development/Github/loris-mri/python/bids_import.py", line 76, in main
    read_and_insert_bids(bids_dir, config_file, verbose, createcand, createvisit)
  File "/Users/cmadjar/Development/Github/loris-mri/python/bids_import.py", line 220, in read_and_insert_bids
    Eeg(
  File "/Users/cmadjar/Development/Github/loris-mri/python/lib/eeg.py", line 153, in __init__
    self.register_data()
  File "/Users/cmadjar/Development/Github/loris-mri/python/lib/eeg.py", line 260, in register_data
    event_file_path     = self.fetch_and_insert_event_file(eeg_file_id, original_file_data.path, derivatives)
  File "/Users/cmadjar/Development/Github/loris-mri/python/lib/eeg.py", line 637, in fetch_and_insert_event_file
    physiological.insert_event_file(
  File "/Users/cmadjar/Development/Github/loris-mri/python/lib/physiological.py", line 551, in insert_event_file
    self.db.insert(
  File "/Users/cmadjar/Development/Github/loris-mri/python/lib/database.py", line 193, in insert
    raise Exception("Insert query failure: " + format(err))
Exception: Insert query failure: (1366, "Incorrect integer value: 'None' for column `loris_rb`.`physiological_task_event`.`EventValue` at row 1")
```

cmadjar (changes requested):
@laemtl I did an initial test with the Face13 dataset from James Desjardins that is on demo and noticed the following issues.

I also tested on a dataset that I created with multiple EEG files and it worked very nicely :).

python/lib/eeg.py
laemtl: Those line mimic the previous logic, but are commented because I think it may be good to not be too restrictive. You can let me know your though and I will adjust accordingly.

python/lib/bidsreader.py
laemtl: Weird, I think I tested the same case without any problem. Let me double-check.

laemtl: Ok, I see, it's a warning, not an error. Does it prevent the pipeline to run?

python/lib/physiological.py
laemtl: Good catch, it's easy to solve.

laemtl: @cmadjar Fixed.

python/lib/bidsreader.py
cmadjar: All good, we can keep the warning. I initially thought that was why the pipeline failed when I tested it but it was because of the task event insertion.

python/lib/eeg.py
cmadjar: I would tend to keep them to ensure you are getting only the EEG or iEEG files. For example, maybe for MEG or other electrophysiological recordings, the files will have the same extensions as EEG but their insertion process might be different than EEG? 
At least, we are sure we will not get the MRIs since the extension of images are `.nii.gz` but relying only on file extensions to grep the proper files seem risky. What do you think?

laemtl: Given your reasoning, I agree with you. Fixed!

cmadjar (approved):

PR #627, opened 2021-05-17
Fix Pybids depreciation notice

laemtl:


python/lib/bidsreader.py
```diff
@@ -10,15 +10,24 @@
 
 import lib.exitcode
 import lib.utilities as utilities
+import bids
+
 try:
-    from bids.layout import BIDSLayout
+    from bids import BIDSLayout
```
cmadjar, line 16:
```suggestion
try:
    from bids import BIDSLayout
except ImportError:
    print("Could not import BIDSLayout. Ensure you have PyBIDS >= 0.12.1 installed.")        
    exit(lib.exitcode.INVALID_IMPORT)
```

If we plan on supporting only PyBIDS > 0.12.1, we should add an error message to tell users to update their pybids package.

cmadjar (changes requested):
One comment on the new imports.

python/lib/bidsreader.py
laemtl: The change is not supposed to break Pybids < 0.12.1 since I check the version before applying the changes.

cmadjar (approved):
Tested the changes with pybids-0.11.1 and pybids-0.13 and everything worked like a charm 👍 . Thank you @laemtl :-)

PR #640, opened 2021-06-29
[Perl] Set default ScannerID value to NULL

vietdhoang:
Modified the Perl pipeline to reflect that the new default value of ScannerID is NULL in the mri_protocol table. This PR should only be merged when [#7496](https://github.com/aces/Loris/pull/7496) is ready to be merged as well.

**Testing instructions**
1. Choose one subject from the RaisinBread dataset to be deleted and re-inserted. The recommended patient is `ROM248` since this patient has not been QC'ed yet and thus the script `delete_imaging_upload.pl`
2. Delete the patient using `delete_imaging_upload.pl`.
3. Run the insertion pipeline and verify that no errors have occurred.

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -363,8 +363,8 @@ sub identify_scan_db {
         Software      => $fileref->getParameter('software_versions')
     });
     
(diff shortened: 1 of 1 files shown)


cmadjar (approved):
👍

Issue #498, opened 2019-09-03
[install] improve error handling for db, mkdir incoming/,  distro detection 


christinerogers:
Things to error-proof in the imaging installer, based on recent feedback, e.g. from Sotirios -- 

- [x] check whether `/data/incoming/` was actually created ([L123](https://github.com/aces/Loris-MRI/blob/minor/imaging_install.sh#L123)) - if not, print message about what commands to run. this happens from time to time (user unable to create new subdirectories right under `/data`

- [x] check whether database connection works - Else abort with a useful error message if can't connect (e.g. password typo)

- [x] improve handling of distro detection, per @PapillonMcGill [Loris-Core #4113](https://github.com/aces/Loris/pull/4113/files) (RM 12834)
Apply the following to imaging_install.sh for more robust detection of the distribution  -- 

```

Detecting distribution
os_distro=$(hostnamectl |awk -F: '/Operating System:/{print $2}'|cut -f2 -d ' ')
debian=("Debian" "Ubuntu")
redhat=("Red" "CentOS" "Fedora" "Oracle")

if [[ " ${debian[*]} " =~ " $os_distro " ]]; then
...
elif [[ " ${redhat[*]} " =~ " $os_distro " ]]; then
...
fi
```

cmadjar:
I think we can close this issue. @christinerogers @vietdhoang do you agree?

PR #642, opened 2021-07-07
Add database connection check in imaging_install.sh

vietdhoang:
### Description
In an effort to make `imaging_install.sh` more error-proof, a new check is added to ensure that the connection to the database is valid. If the user provides invalid credentials (username and/or password), then the script will alert them and abort. For any other connection problems, the MySQL error message is printed out instead. This address the second task in #498. 

### How to test
1. Run the `imaging_install.sh` script with incorrect username and password a check that the script catches the error and exits. The error printed out should be:
`ERROR: invalid username and/or password. Aborting...`
2. Run the script again with other incorrect credentials (i.e host, database name, etc.) and see if the script catches those errors as well.
3. Run the script with correct credentials to ensure that everything runs smoothly.

### DIFF
--- a/imaging_install.sh
+++ b/imaging_install.sh
@@ -59,7 +59,7 @@ mridir=`pwd`
 # Test the connection to the database before proceeding
 echo "Testing connection to database..."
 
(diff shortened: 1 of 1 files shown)


cmadjar (approved):
LGTM. Thank you @vietdhoang

PR #643, opened 2021-07-08
Improved distribution detection to imaging_install.sh

vietdhoang:
### Description
Improved distribution detection by adding support for all Debian-based (Ubuntu, Debian) and RedHat-based (RedHat, CentOS, Fedora, Oracle) distributions. This PR addresses the third task in #498. The added code is also based off the code provided in #498. 

### Testing instructions
1. First, run the following command in the terminal to check that it can properly detect your Linux distribution:
```bash
hostnamectl |awk -F: '/Operating System:/{print $2}'|cut -f2 -d ' '
``` 
2. Next, create a bash script containing only the block of code below. Ensure that if you are running a Debian distribution that DICOM Toolkit automatically installs:
```bash
os_distro=$(hostnamectl |awk -F: '/Operating System:/{print $2}'|cut -f2 -d ' ')
debian=("Debian" "Ubuntu")
redhat=("Red" "CentOS" "Fedora" "Oracle")

if [[ " ${debian[*]} " =~ " $os_distro " ]]; then
    #Check if apt-get is installed
    APTGETCHECK=`which apt-get`
    if [ ! -f "$APTGETCHECK" ]; then
        echo "\nERROR: Unable to find apt-get"
        echo "Please ask your sysadmin or install apt-get\n"
        exit
    fi

    echo "Installing DICOM Toolkit (May prompt for sudo password)"
    sudo -S apt-get install dcmtk

elif [[ " ${redhat[*]} " =~ " $os_distro " ]]; then
    echo "You are running ${os_distro}. Please also see Loris-MRI Readme for notes and links to further documentation in our main GitHub Wiki on how to install the DICOM Toolkit and other required dependencies for RedHat-based distributions."
fi
``` 
3. Manipulate the variable `os_distro` and manually assign distributions to test if the code works for distributions other than your own.
4. Run `imaging_install.sh` to ensure that everything runs smoothly.

### DIFF
--- a/imaging_install.sh
+++ b/imaging_install.sh
@@ -240,6 +240,7 @@ elif [[ " ${redhat[*]} " =~ " $os_distro " ]]; then
 	echo "You are running ${os_distro}. Please also see Loris-MRI Readme for notes and links to further documentation in our main GitHub Wiki on how to install the DICOM Toolkit and other required dependencies for RedHat-based distributions."
 fi
 
+######################################################################
 ###### Update the Database table, Config, with the user values #######
 ######################################################################
 echo "Populating database configuration entries for the Imaging Pipeline and LORIS-MRI code and images Path:"
--- a/imaging_install.sh
+++ b/imaging_install.sh
@@ -218,21 +218,28 @@ echo
 ################################################################################################
 #####################################DICOM TOOLKIT##############################################
 ################################################################################################
-if cat /etc/os-release | grep ^NAME | fgrep -q CentOS ; then
-    echo "You are running CentOS. Please also see Loris-MRI Readme for notes and links to further documentation in our main GitHub Wiki on how to install the DICOM Toolkit and other required dependencies."
-else
-    #Check if apt-get is installed
-    APTGETCHECK=`which apt-get`
-    if [ ! -f "$APTGETCHECK" ]; then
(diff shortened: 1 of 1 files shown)


imaging_install.sh
```diff
@@ -218,20 +218,28 @@ echo
...
-        echo "Please ask your sysadmin or install apt-get\n"
-        exit
-    fi
-
-    echo "Installing DICOM Toolkit (May prompt for sudo password)"
-    sudo -S apt-get install dcmtk
+
+# Detecting distribution
+os_distro=$(hostnamectl |awk -F: '/Operating System:/{print $2}'|cut -f2 -d ' ')
+debian=("Debian" "Ubuntu")
```
cmadjar, line 224: I get the following error here: 
```
Syntax error: "(" unexpected
```

cmadjar (changes requested):
@vietdhoang see comment below

imaging_install.sh
vietdhoang: Hi Cecile, I ran the install script and I was unable to reproduce your error. I also don't see where there shouldn't be an '('. Would it be possible for you to double check or to give me more information about the error?

cmadjar (approved):
Sorry, I just realized the issue was on my end. All good 👍

PR #645, opened 2021-07-15
Append entries to participants.tsv if participants.tsv file already exists in bids import

cmadjar:
This fixes an issue for prospective studies importing BIDS dataset one subject at a time.

Until now, the pipeline always checked that the file content in the bids_import directories matched the file content in the incoming BIDS directory. When loading one subject at a time, this will cause failures as the participants.tsv file would change for each subject loaded. Instead, in this PR we read the content of the participants.tsv file for the new upload and compare it to the participants.tsv file already existing in the bids_import directory. If there are new subjects listed in the new participants.tsv, it will append a row to the participants.tsv file present in the bids_import directory to add the subject to it.

### DIFF
--- a/python/bids_import.py
+++ b/python/bids_import.py
@@ -296,11 +296,19 @@ def create_loris_bids_directory(bids_reader, data_dir, verbose):
         )
 
     # copy the participant.tsv file to the new directory
-    lib.utilities.copy_file(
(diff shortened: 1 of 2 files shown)


cmadjar:
Tested by Alfredo in his setup so merging it.

PR #647, opened 2021-07-29
Participants.tsv: Use project and site name instead of alias

laemtl:
Adapt the logic from #558 to use project and site Names instead of Alias.

Motivations: 
  - readibility
  - uniformity (subproject's value is subproject's title)
  - follow API's logic where project and site Names are used

Note: In candidate.py, the if self.site case was deleted. To my understanding, it wil lbe populated after the first iter with the first participant's site, and assume that all subsequent participants are from the same site, which is not necessary the case.

### DIFF
--- a/python/lib/candidate.py
+++ b/python/lib/candidate.py
@@ -113,10 +113,10 @@ class Candidate:
 
         if self.verbose:
             print("Creating candidate with " + \
-                  "PSCID = "   + self.psc_id + ", " + \
-                  "CandID = "  + str(self.cand_id) + ", " + \
-                  "Site = "    + str(self.center_id)  + ", " + \
-                  "Project = " + str(self.project_id))
+                  "PSCID     = " + self.psc_id + ", " + \
+                  "CandID    = " + str(self.cand_id) + ", " + \
+                  "CenterID  = " + str(self.center_id)  + ", " + \
+                  "ProjectID = " + str(self.project_id))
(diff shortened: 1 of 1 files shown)


python/lib/candidate.py
```diff
@@ -83,42 +82,40 @@ def create_candidate(self, db, participants_info):
...
                   " participants.tsv's site column")
             sys.exit(lib.exitcode.PROJECT_CUSTOMIZATION_FAILURE)
 
         if self.verbose:
             print("Creating candidate with " + \
                   "PSCID = "   + self.psc_id + ", " + \
                   "CandID = "  + str(self.cand_id) + ", " + \
-                  "Site = "    + str(self.site)  + " and " + \
+                  "Site = "    + str(self.center_id)  + ", " + \
                   "Project = " + str(self.project_id))
```
cmadjar, line 119:
```suggestion
                  "CenterID = "    + str(self.center_id)  + ", " + \
                  "ProjectID = " + str(self.project_id))
```

cmadjar (changes requested):
@laemtl One tiny change on the log text to reflect that we are showing CenterID and ProjectID values instead of site and project name in the logs

laemtl:
@cmadjar Done!

cmadjar (approved):
Code looks good and tested locally and it works great! 👍

PR #651, opened 2021-08-12
Added a GitHub Action that will run flake8 on submitted PRs

vietdhoang:
### Summary
A new Python linter, flake8, has been added as a GitHub Action. It will run on every submitted PR and annotate all the errors that it finds. This PR addresses the issue #649. 

Documentation for the GitHub Action can be found [here](https://github.com/marketplace/actions/run-flake8-with-reviewdog). Essentially, this GitHub Action runs flake8 with [reviewdog](https://github.com/reviewdog/reviewdog). Reviewdog parses the output from flake8 and uses to it annotate code in the PR.

PR #665, opened 2021-08-17
Flake8 on bidsreader.py

vietdhoang:
### Summary
Ran flake8 and corrected some errors that were caught. In some of the Python files, I decided to ignore some errors (these errors are not on the list in `.github/workflows/flake8_python_linter.yml`) due to unique formatting that helped make the code more readable. Once #651 has been merged, you will be able to see these errors here on GitHub.

### DIFF
--- a/python/lib/bidsreader.py
+++ b/python/lib/bidsreader.py
@@ -1,10 +1,7 @@
 """Reads a BIDS structure into a data dictionary using bids.grabbids."""
 
-import csv
-import random
 import re
 import os
-import glob
 import sys
 import json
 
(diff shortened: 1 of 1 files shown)


python/lib/bidsreader.py
```diff
@@ -93,7 +91,7 @@ def load_bids_data(self):
         force_arr     = [re.compile("_annotations\.(tsv|json)$")]
 
         # BIDSLayoutIndexer is required for PyBIDS >= 0.12.1
-        bids_pack_version = list(map(int, bids.__version__.split('.')))
+        # bids_pack_version = list(map(int, bids.__version__.split('.')))
```
cmadjar, line 94: @laemtl does that change seem reasonable to you? Looks like the bids_pack_version variable is not used later in the script so I guess this is why it is commented here.

laemtl: @cmadjar I think something happen durimg a rebase since the variable is already defined L24. The commented line can be deleted. Thanks!

PR #671, opened 2021-08-24
Stop using config setting useProjects (projects are always "used" now)

### LINKED ISSUE #670: Setting useProjects has been removed since LORIS 21.0 and should not be used
Method `getUseProjects()` should be removed from the `ConfigOB.pm` class as this setting has been removed on the LORIS side since version 21.0. Projects are now used all the time.

nicolasbrossard:
Config setting `useProjects` has been removed from the LORIS code base since version 21.0 as projects are always "used" now. This PR removes from the LORIS-MRI codebase any reference or access to that setting.

How to test:
- Ensure setting `useProjects` is not in the DB
- Insert a scan that belongs to a candidate that does not exist in the DB and ensure that the profile file specifies a project ID and a subproject ID for that candidate/visit.
- Test that the scan is inserted successfully and that the candidate is created by the pipeline, along with the appropriate visit.

Fixes #670

### DIFF
--- a/dicom-archive/profileTemplate.pl
+++ b/dicom-archive/profileTemplate.pl
@@ -58,7 +58,7 @@ sub getSubjectIDs {
         # ($subjectID{'SubprojectID'}) = $patientName =~ /_(\d+)$/;
         # When createVisitLabel is set to 0, $subjectID{'SubprojectID'} is ignored.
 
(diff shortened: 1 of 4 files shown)


cmadjar (approved):
Perfect 👍

jesscall:
I think maybe this should have been sent to 23.0-release in order to have a bugfix release

PR #672, opened 2021-08-27
Start next stage

laemtl:
This PR adds support for a -n flag for bids_import to start the visit stage.
```python3 bids_import.py -d '/test/' -p loris_config.py -csvn```

This implementation uses a basic API class to make use of the next stage API endpoint (see #470) 
An API config data is required in dicom-archive/.loris-mri/loris_config.py with the following content:
``` 
api = {
    'host'     : '',
    'version'  : 'v0.0.4-dev',
    'username' : '',
    'password' : '',
}
```

### DIFF
--- a/python/bids_import.py
+++ b/python/bids_import.py
@@ -8,6 +8,7 @@ import getopt
 import re
 import lib.exitcode
 import lib.utilities
+import pandas as pd
 from lib.database   import Database
 from lib.candidate  import Candidate
 from lib.bidsreader import BidsReader
@@ -30,6 +31,7 @@ def main():
     verbose     = False
     createcand  = False
     createvisit = False
+    startvisit  = False
     profile     = ''
 
     long_options = [
@@ -45,11 +47,12 @@ def main():
         '\t-d, --directory      : BIDS directory to parse & insert into LORIS\n'
         '\t-c, --createcandidate: to create BIDS candidates in LORIS (optional)\n'
         '\t-s, --createsession  : to create BIDS sessions in LORIS (optional)\n'
+        '\t-n, --startvisit     : to start Visit stage in LORIS (optional)\n'
         '\t-v, --verbose        : be verbose\n'
     )
 
     try:
-        opts, args = getopt.getopt(sys.argv[1:], 'hp:d:csv', long_options)
+        opts, args = getopt.getopt(sys.argv[1:], 'hp:d:csvn', long_options)
     except getopt.GetoptError:
         print(usage)
         sys.exit(lib.exitcode.GETOPT_FAILURE)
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: python/bids_import.py [hunk 1]
```diff
@@ -196,14 +203,23 @@ def read_and_insert_bids(bids_dir, config_file, verbose, createcand, createvisit
...
-        # session if it does not exist yet in LORIS and the createvisit is set
-        # to true. If no visit in BIDS structure, then use default visit_label
-        # stored in the Config module)
+        visit_date = None
+        if startvisit:
+            scans_data = pd.concat([
+                scan_data.to_df()
+                for scan_data in bids_reader.bids_layout.get_collections('session', 'scans')
+            ])
+            visit_date = min(scans_data['acq_time'])
```
xlecours, line 212: In case scans_data['acq_time'] would be not defined, you could use the `get` function. 
ex: 
`scans_data.get('acq_time', datetime.datetime.now())`

cmadjar: not sure I would use time of upload as `acq_time` here since a scan can be uploaded at a different date than the upload date. Might be best keeping the acquisition time as undef is that case instead of setting a default date. Unless I am misunderstanding your suggestion @xlecours?

python/lib/api.py
```diff
@@ -0,0 +1,79 @@
+"""Allows LORIS API connectivity for LORIS-MRI python code base"""
```
cmadjar, line 1: This is fine for now but I am guessing that as we use more API functions for querying, this file is going to become gigantic. No big deal for now but we might want to think about a proper structure for API querying (a bit like now there is a database folder handling database queries with several files depending on what is being queried). Just a thought.

cmadjar (changes requested):
@laemtl I would tend to put the API config in the database_config file and modify the template file https://github.com/aces/Loris-MRI/blob/main/dicom-archive/database_config_template.py to keep all python related configs in the same file. It will also make the install easier.

Note: if we keep a separate file for the API config, it needs to be taken care of during the install process and a template needs to be provided.

laemtl:
> @laemtl I would tend to put the API config in the database_config file and modify the template file https://github.com/aces/Loris-MRI/blob/main/dicom-archive/database_config_template.py to keep all python related configs in the same file. It will also make the install easier.
> 
> Note: if we keep a separate file for the API config, it needs to be taken care of during the install process and a template needs to be provided.
It is a good option but we should probably rename that file then? loris_config?

python/bids_import.py
xlecours: I makes perfect sense to have a date explicitly provided. I just don't know how it would work. Could it be grep'ed from the BIDS files?

cmadjar: I believe this is what is being done in line 210. I guess the date could also be inside a session level file? I have not looked at the BIDS session spec in a while but it could be feasible. @laemtl thoughts?

cmadjar:
> It is a good option but we should probably rename that file then? loris_config?

@laemtl agreed. Initially, it only contained database credentials but it is becoming more of a general config file now. Could be named `config.py` or `loris_python_config.py` to make it clear that this config is only for the python scripts? Although, technically the `.py` extension sort of implies python...

PR #716, opened 2021-10-06
Fix mass_nifti_pic.py in the case there is already an entry in parameter_file for the pic path

cmadjar:
This fixes a few bugs found when testing mass_nifti_pic.py:
- fixes the warning printed
- checks whether a pic has already been created and is present in parameter_file. If so, print a warning that the pic already exists and that if the user wishes to recreate it, the option `--force`/`-f` should be used when running the script
- adds a force option to force the re-creation of the pic on the filesystem with the same path as the path already stored in the database

### DIFF
--- a/python/mass_nifti_pic.py
+++ b/python/mass_nifti_pic.py
@@ -166,9 +166,6 @@ def make_pic(file_id, config_file, force, verbose):
     # load the Imaging object
     imaging = Imaging(db, verbose)
 
-    # checks if there is already a pic for the NIfTI file
-    existing_pic_file_in_db = imaging.grep_parameter_value_from_file_id(file_id, 'check_pic_filename')
-
     # grep the NIfTI file path
     nii_file_path = imaging.grep_file_path_from_file_id(file_id)
     if not nii_file_path:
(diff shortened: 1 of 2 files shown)


python/mass_nifti_pic.py
```diff
@@ -159,17 +166,23 @@ def make_pic(file_id, config_file, verbose):
...
-        print('WARNING: no file in the database with FileID = ' + file_id)
+        print('WARNING: no file in the database with FileID = ' + str(file_id))
         return
     if not re.search('.nii.gz$', nii_file_path):
         print('WARNING: wrong file type. File ' + nii_file_path + ' is not a .nii.gz file')
         return
     if not os.path.exists(data_dir + nii_file_path):
         print('WARNING: file ' + nii_file_path + ' not found on the filesystem')
         return
+    if existing_pic_file_in_db and not force:
```
nicolasbrossard, line 183: I would move the declaration of `existing_pic_file_in_db` closer to the line where it isa first used (i.e. here). Just nitpicking...  :)

nicolasbrossard (changes requested):

python/mass_nifti_pic.py
cmadjar: @nicolasbrossard done in the last commit! Thank you! :)

nicolasbrossard (approved):

PR #722, opened 2021-11-16
uploadNeuroDB/imaging_non_minc_insertion.pl now works with all versions of Perl

### LINKED ISSUE #721: uploadNeuroDB/imaging_non_minc_insertion.pl does not work with recent versions of Perl
Script `uploadNeuroDB/imaging_non_minc_insertion.pl` has code that does not compile with recent versions of Perl. When I run the script with Perl 5.26, I get:
```
Experimental keys on scalar is now forbidden at ./uploadNeuroDB/imaging_non_minc_insertion.pl line 392.
```

nicolasbrossard:
Script uploadNeuroDB/imaging_non_minc_insertion.pl uses code that is no longer supported in earlier versions of Perl. This PR refactors the script so that it works with all Perl versions.

Fixes #721

### DIFF
--- a/uploadNeuroDB/imaging_non_minc_insertion.pl
+++ b/uploadNeuroDB/imaging_non_minc_insertion.pl
@@ -389,7 +389,7 @@ if ($metadata_file) {
     close( FILE );
 
     my $metadata = decode_json($json);
-    foreach my $parameter (sort keys $metadata) {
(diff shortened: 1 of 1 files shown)


cmadjar (approved):
Oops, merged before approving. LOL

PR #727, opened 2021-11-23
Fixes a NoneType error bug when no entry in scans.tsv for a given file and trying to look for an acquisition date

cmadjar:
This PR fixes a NoneType error that gets issued when no entry was found for a given *nii.gz or EEG file in scans.tsv to determine the acquisition date.


AlexandraLivadas (approved):
This looks great to me!

PR #731, opened 2021-11-29
get_dicom_files.pl: add support for archives that contain files with identical base names

nicolasbrossard:
Script `get_dicom_files.pl` will fail if the archive contains DICOM files with identical base names and one of these files should be extracted by the script. Example:

Let's say you have an archive which contains two DICOM files:
```
A/B/C/my_file.dcm
X/Y/Z/my_file.dcm
```
and imagine that these DICOM files are associated to two different modalities (say `M1` and `M2`). When the pipeline processes that archive, it will store in table `tarchive_files` two records with identical `File` values that are associated to different entries in table `tarchive_series`. When you run the extract script to get the DICOM files for modality M1, the script will try to retrieve file `myfile.dcm` from the archive but then it won't know which one to get since there are two...

This PR improves the extraction script so that the MD5 sum of the DICOM files are used _along_ with the file's base names when "ambiguous" extraction requests are encountered (as the one described above).

### DIFF
--- a/tools/get_dicom_files.pl
+++ b/tools/get_dicom_files.pl
@@ -9,7 +9,7 @@ get_dicom_files.pl - extracts DICOM files for specific patient names/scan types
 
 =head1 SYNOPSIS
 
-perl get_dicom_files.pl [-name patient_name_patterns] [-type scan_type_patterns] [-outdir tmp_dir] [-outfile tarBasename] 
+perl get_dicom_files.pl [-names patient_name_patterns] [-types scan_type_patterns] [-outdir tmp_dir] [-outfile tarBasename] 
            [-id candid|pscid|candid_pscid|pscid_candid] -profile profile
 
 Available options are:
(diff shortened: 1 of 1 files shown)


tools/get_dicom_files.pl
```diff
@@ -291,38 +291,94 @@ =head1 DESCRIPTION
...
         while (<LIST_TAR_CONTENT>) {
             chomp;
-            my($fileName, $dirName, $suffix) = fileparse($_);
-            print FILE_LIST "$_\n" if grep($_ eq $fileName, @{ $filesRef->{$file}->{'DICOM'} });
+            my($dicomFileBaseName, $dirName, $suffix) = fileparse($_);
+            if(grep($_ eq $dicomFileBaseName, @{ $filesRef->{$mincFile}->{'DICOM'} })) {
+                print FILE_LIST "$_\n";
+                # If there is more than one DICOM file with this basename, store it in the hash: we'll extract
+                # all files with that base name and use the MD5 sum to only retain the correct one later
+                if( keys %{ $md5sum{$dicomFileBaseName} } > 1) {
```
cmadjar, line 335:
```suggestion
                if( keys %{ $md5sum{$dicomFileBaseName} } > 0) {
```

cmadjar (changes requested):
@nicolasbrossard the script was not extracting the correct files until I made the changes suggested below. 

FYI - From the debugger when I tested that line
```
  DB<7> x keys %{ $md5sum{$dicomFileBaseName} }                                                                                        
0  'assembly/676061/V3/mri/native/demo_676061_V3_t2_003.mnc'
  DB<8> x keys %{ $md5sum{$dicomFileBaseName} } > 1                                                                                    
0  ''
  DB<9> x keys %{ $md5sum{$dicomFileBaseName} } >0                                                                                     
0  1
```

PR #744, opened 2021-12-09
Move violated files when running minc_insertion.pl

### LINKED ISSUE #739: run minc_insertion.pl on a MINC that belongs to a valid UploadID with a valid protocol (scan type identified) but an exclusionary check is identified (from mri_protocol_checks) => check that the MINC file gets moved to trashbin and an entry is added in mri_violations_log

cmadjar:
This resolves an issue where when running `minc_insertion.pl` on its own, the files that are inserted into the `mri_protocol_violated_scans`, `mri_violations_log (exclude violations only)` and `MRICandidateError` tables, the file is not moved to `trashbin` and the path inserted in those tables are not reliable as there is no files in the file system.

This is not an issue when running the main pipeline as all the files are moved to `trashbin` at the end of `tarchiveLoader` execution. 

This resolves #739

### DIFF
--- a/docs/scripts_md/MRI.md
+++ b/docs/scripts_md/MRI.md
@@ -83,10 +83,11 @@ INPUTS:
   - $db             : database object
   - $minc\_location  : location of the MINC files
   - $uploadID       : ID of the upload containing the scan
+  - $data\_dir       : path to the LORIS MRI data directory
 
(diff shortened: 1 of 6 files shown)


zaliqarosli (approved):
LGTM 👍

PR #746, opened 2021-12-10
fix coordinate space not being inserted in the files table when running imaging_non_minc_insertion.pl

cmadjar:
This fixes an issue found by a collaborator where the coordinate space provided to the script is not being used when inserting the file into the `files` table.


cmadjar:
This was tested by a collaborator and it worked as expected so merging it.

PR #756, opened 2022-01-12
Fix bug where string concatenate with int in mass_nifti_pic.py

cmadjar:
This fixes:
- string/int concatenation bug present in the mass_nifti_pic.py file.
- update or insert values into parameter_file depending on whether they are already present or not
- fix the largest number provided to the script that was not being processed


cmadjar:
This was tested by Gabriel P who reported the bugs.

PR #760, opened 2022-02-04
Bugfix - Annotation get_nearest should take subject arg

jesscall:
Current behaviour:
Takes one candidate’s annotation files and copies it to all candidates raw sub folders (and also inserts to the DB).

This PR fixes the search by adding the subject as a parameter to the search.

### DIFF
--- a/python/lib/eeg.py
+++ b/python/lib/eeg.py
@@ -317,7 +317,7 @@ class Eeg:
 
             # create data chunks for React visualization in
             # data_dir/bids_import/bids_dataset_name_BIDSVersion_chunks directory
-            physiological.create_chunks_for_visualization_on_cbrain(eeg_file_id, self.data_dir)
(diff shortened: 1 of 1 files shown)


python/lib/eeg.py
```diff
@@ -317,7 +317,7 @@ def register_data(self, derivatives=False):
 
             # create data chunks for React visualization in
             # data_dir/bids_import/bids_dataset_name_BIDSVersion_chunks directory
-            physiological.create_chunks_for_visualization(eeg_file_id, self.data_dir)
+            physiological.create_chunks_for_visualization_on_cbrain(eeg_file_id, self.data_dir)
```
jesscall, line 320:
```suggestion
            physiological.create_chunks_for_visualization(eeg_file_id, self.data_dir)
```

cmadjar (approved):
works great 👍

PR #762, opened 2022-02-07
Bugfix: Floats returning false on isdecimal() fxn

jesscall:
```
a = '1.23456'
b = '3'

a.isdecimal()
>> False !! this is problematic

b.isdecimal()
>> True
```

Extend support and make sure that strings in float format are accepted. Otherwise a null is inserted in the onset / duration field of annotation instance


cmadjar (approved):
LGTM 👍

PR #766, opened 2022-03-03
Bulletproof the retrieval of the modalities on which to compute the SNR

### LINKED ISSUE #764: Fatal error when computing the SNRs
Gabriel (project MCNRR) reported this error on LORIS 23:
```
Can't use string ("flair") as an ARRAY ref while "strict refs" in use at /data/mcnrr/bin/mri/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm line 1890.
```
The faulty line is:
```
if ( grep($_ eq $fileScanType, @$modalities) ) {
```
The problem lies with the way config setting `compute_snr_modalities` is retrieved. Fixing this bug should involve using the `ConfigOB` in order to fetch this setting, possibly adding a new method in order to do so.

nicolasbrossard:
This PR adds a method to the `ConfigOB` object broker in order to allow the retrieval of config setting `compute_snr_modalities`. It also makes use of this new feature in `MRIProcessingUtilities.pm`. 

Fixes #764 .

### DIFF
--- a/docs/scripts_md/ConfigOB.md
+++ b/docs/scripts_md/ConfigOB.md
@@ -64,12 +64,13 @@ RETURN: new instance of this class.
 Private method. This method fetches setting `$setting` from the LORIS table 
 Config. It will throw a `NeuroDB::objectBroker::ObjectBrokerException` if either
 the database transaction failed for some reason or it succeeded but returned no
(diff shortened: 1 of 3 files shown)


cmadjar (approved):
👍  Works perfectly! Thank you @nicolasbrossard :-)

PR #794, opened 2022-08-17
Set DICOM dates to undef if the date does not follow proper DICOM standard

cmadjar:
# Description

PatientBirthDate should follow a string Value Representation (a.k.a `Date (DA)`) which is of the form `YYYYMMDD`. See [Patient's Birth Date Attribute in DICOM standard Browser](https://dicom.innolitics.com/ciods/rt-dose/patient/00100030) and [descriptions of DICOM value representations](https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html).

Sometimes, some deidentifying tools might insert incorrect date format (latest example was 00/00/0000) which leads to database errors when inserting invalid dates to the `tarchive` table for example.

Proposed fix: if the date is not following the original DICOM Value Representation, then it means a tool was run on it to get rid of the date, therefore, the DICOM scripts will set that date to `NULL` in the `tarchive` table.

Discussed during the LORIS HBCD dev meeting call.

### DIFF
--- a/dicom-archive/DICOM/DCMSUM.pm
+++ b/dicom-archive/DICOM/DCMSUM.pm
@@ -1205,7 +1205,11 @@ sub date_format {
         my $diff = &Math::Round::nearest(0.01, ($Y + $M/12.0 + $D/365.0)*1) . " or $Y years, $M months $D days";
 	    return $diff;
     }
-    $first = ($first =~ m/(\d\d\d\d)(\d\d)(\d\d)/ ? s/(\d\d\d\d)(\d\d)(\d\d)/$1-$2-$3/ : undef);
+    # return undef if the date does not follow DICOM Date Value Representation (YYYYMMDD)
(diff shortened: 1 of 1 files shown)


cmadjar:
Tested on HBCD pilot data that had a DoB set to `00/00/0000` and a ScanDate set to `20220614`. See result of the insertion below after the fixed from this PR:
```
MariaDB [loris]> select TarchiveID,DateAcquired, PatientDoB, PatientName, DicomArchiveID from tarchive;
+------------+--------------+------------+--------------------+------------------------------------------------------------------+
| TarchiveID | DateAcquired | PatientDoB | PatientName        | DicomArchiveID                                                   |
+------------+--------------+------------+--------------------+------------------------------------------------------------------+
|         84 | 2022-06-14   | NULL       | XX                 | 1.2.840.113619.6.408.109357156493812671873443248051426458364copy |
+------------+--------------+------------+--------------------+------------------------------------------------------------------+
```

PR #809, opened 2022-10-19
ProjectID fix

laemtl:
Solves several problems:
- Profile `getSubjectIDs` comments/documentation for `$subjectID{'ProjectID'}`  updated
- Candidate insert statement updated with `RegistrationProjectID` instead of `ProjectID`.
- Adds new LORIS Configs: `CreateVisit` (replace the variable in the prod file), `default_project` and `default_cohort` (fallback to the variables in the prod file if not set). See https://github.com/aces/Loris/pull/8384
- Updated rules to set the default project:
   - Value in the `prod` file
   - session `ProjectID`
   - candidate `RegistrationProjectID`
   - config default_project
   - throw an error
- Updated rules to set the default cohort:
   - Value in the `prod` file
   - session `CohortID`
   - config default_cohort
   - ("UNKN", 0)

### DIFF
--- a/dicom-archive/profileTemplate.pl
+++ b/dicom-archive/profileTemplate.pl
@@ -51,8 +51,8 @@ sub getSubjectIDs {
 
         $subjectID{'createVisitLabel'} = 1;
 
-	# $subjectID{'CohortID'} = undef;
-        # When createVisitLabel is set to 1, CohortID  must also
+        # $subjectID{'CohortID'} = undef;
+        # When createVisitLabel is set to 1, CohortID must also
         # be set to the ID of the cohort that the newly created
         # visit should have. Assuming for example that all patient
         # names end with "_<myCohortID>", then we could write:
--- a/dicom-archive/profileTemplate.pl
+++ b/dicom-archive/profileTemplate.pl
@@ -59,7 +59,7 @@ sub getSubjectIDs {
         # ($subjectID{'CohortID'}) = $patientName =~ /_(\d+)$/;
         # When createVisitLabel is set to 0, $subjectID{'CohortID'} is ignored.
 
-        # $subjectID{'ProjectID'} = 1;
+        # $subjectID{'ProjectID'} = undef;
         # If config setting 'createVisitLabel' is true
         # then $subjectID{'ProjectID'} must be set to the project ID of the
         # newly created visit. Assuming for example that all patients
@@ -81,7 +81,7 @@ sub getSubjectIDs {
 
         $subjectID{'createVisitLabel'} = 0;
   
-        # $subjectID{'CohortID'} = 1;
+        # $subjectID{'CohortID'} = undef;
         # When createVisitLabel is set to 1, CohortID must also
         # be set to the ID of the cohort that the newly created
         # visit should have. Assuming for example that visits V01 and V02
@@ -91,7 +91,7 @@ sub getSubjectIDs {
         #     ? 1 : 2;
         # When createVisitLabel is set to 0, $subjectID{'CohortID'} is ignored.
         
-        $subjectID{'ProjectID'} = 1;
+        # $subjectID{'ProjectID'} = undef;
         # If config setting 'createVisitLabel' is true
         # then $subjectID{'ProjectID'} must be set to the project ID of the
         # newly created visit. Assuming for example that candidates with a
--- a/dicom-archive/profileTemplate.pl
+++ b/dicom-archive/profileTemplate.pl
@@ -40,35 +40,44 @@ sub getSubjectIDs {
      # If patientName is phantom scan or test scan
      # CandID is scanner DCCID (based on site alias)
      # visitLabel is scan patient name
-     # Set createVisitLable to 
-     #      a. 1 if imaging pipeline should create the visit label (when visit label has not been created yet in the database. 
-     #      b. 0 if imaging pipeline should not create the visit label (when visit label has not been created yet in the database. 
     if ($patientName =~ /PHA/i or $patientName =~ /TEST/i) {
 
         $subjectID{'CandID'}     = NeuroDB::MRI::my_trim(NeuroDB::MRI::getScannerCandID($scannerID, $db));
         $subjectID{'visitLabel'} = NeuroDB::MRI::my_trim($patientName);
         $subjectID{'isPhantom'} = 1;
 
-        $subjectID{'createVisitLabel'} = 1;
-
-        # $subjectID{'CohortID'} = undef;
-        # When createVisitLabel is set to 1, CohortID must also
-        # be set to the ID of the cohort that the newly created
-        # visit should have. Assuming for example that all patient
-        # names end with "_<myCohortID>", then we could write:
-        # ($subjectID{'CohortID'}) = $patientName =~ /_(\d+)$/;
-        # When createVisitLabel is set to 0, $subjectID{'CohortID'} is ignored.
+        # Override createVisitLabel here if its value needs to be different than the createVisit DB config
+        # $subjectID{'createVisitLabel'} = undef;
+        # Set createVisitLabel to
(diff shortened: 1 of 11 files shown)


dicom-archive/profileTemplate.pl
```diff
@@ -88,6 +91,7 @@ sub getSubjectIDs {
         #     ? 1 : 2;
         # When createVisitLabel is set to 0, $subjectID{'SubprojectID'} is ignored.
         
+        $subjectID{'ProjectID'} = 1;
```
nicolasbrossard, line 94: I am not sure I would give a default value to the `ProjectID`. I think it would be better to leave it undefined and have the MRI pipeline complain that if was not set instead of setting the project ID of the newly created candidate to 1 (which might not be the correct value). @cmadjar Can you weigh in on this?

cmadjar: I agree with @nicolasbrossard. Since `ProjectID` can be `NULL` I would leave the default to be `undef`.

dicom-archive/profileTemplate.pl
```diff
@@ -79,6 +81,7 @@ sub getSubjectIDs {
 
         $subjectID{'createVisitLabel'} = 0;
   
+        # $subjectID{'SubprojectID'} = 1;
```
cmadjar, line 84: here as well, I would put the default to be `undef`

dicom-archive/profileTemplate.pl
```diff
@@ -51,13 +51,15 @@ sub getSubjectIDs {
...
 
+        # $subjectID{'SubprojectID'} = 1;
         # When createVisitLabel is set to 1, SubprojectID must also
         # be set to the ID of the subproject that the newly created
         # visit should have. Assuming for example that all patient
         # names end with "_<mySubProjectID>", then we could write:
         # ($subjectID{'SubprojectID'}) = $patientName =~ /_(\d+)$/;
         # When createVisitLabel is set to 0, $subjectID{'SubprojectID'} is ignored.
 
+        # $subjectID{'ProjectID'} = 1;
```
cmadjar, line 62: here as well, I would put the default to be undef

dicom-archive/profileTemplate.pl
```diff
@@ -51,13 +51,15 @@ sub getSubjectIDs {
 
         $subjectID{'createVisitLabel'} = 1;
 
+        # $subjectID{'SubprojectID'} = 1;
```
cmadjar, line 54: here as well, I would put the default to be undef

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -1787,13 +1787,13 @@ sub CreateMRICandidates {
...
-        ProjectID            => $subjectIDsref->{'ProjectID'},
-        Sex                  => $sex,
-        RegistrationCenterID => $centerID,
-        UserID               => $User,
+        CandID                => $subjectIDsref->{'CandID'},
+        PSCID                 => $subjectIDsref->{'PSCID'},
+        DoB                   => $subjectIDsref->{'PatientDoB'},
+        Sex                   => $sex,
+        RegistrationCenterID  => $centerID,
+        RegistrationProjectID => $subjectIDsref->{'ProjectID'},
```
cmadjar, line 1795: Given that the default ProjectID would be undef, I wonder whether the if statement line 1758 should not be modified by `!defined $subjectIDsref->{'ProjectID'} and $pscid eq 'scanner'` since no ProjectID should be assigned to scanner candidates.

@laemtl thoughts?

cmadjar (changes requested):

dicom-archive/profileTemplate.pl
laemtl: I get confused by [this change](https://github.com/aces/Loris/blob/973fb261e420ae23696ddda628bfe2831260353e/SQL/Archive/24.0/2019-07-01_fix_project_in_session.sql).
Since v24.0 session ProjectID cannot be null. createVisitLabel is set to 1 by default, so setting Project to undef will trigger an INSERT error when attempting to create a session. Can we fix this by setting createVisitLabel to 0 by default?

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
laemtl: I believe the change [here](https://github.com/aces/Loris/blob/973fb261e420ae23696ddda628bfe2831260353e/SQL/Archive/24.0/2019-07-01_fix_project_in_session.sql) will not allow scanner to have no ProjectID.

cmadjar:
@laemtl Based on today's discussion at the LORIS call, the best approach to this might be to set the `RegistrationProjectID` of the `scanner` candidate to be created to the same `RegistrationProjectID` as the one used for the candidate being scanned (PatientName). 

Then, when Phantom session are being inserted into LORIS, the session `ProjectID` will be dependent on the string in the PatientName (the project info will be in the PatientName the same way the site info is).

cmadjar:
I guess, the same way we have `determinePSC` called before `determineScannerID` in `tarchive_validation.pl`, `minc_insertion.pl` and `tarchiveLoader.pl`, there should be a `determineProjectID` based on the `PatientName` which will then be passed to `determineScannerID` function the same way it is passed the `CenterID`.

@laemtl @nicolasbrossard would that make sense to you?

nicolasbrossard:
@cmadjar Makes sense to me.

cmadjar:
@laemtl ping on this one :)

cmadjar:
@laemtl I tested the PR with the new config in the config module but I get the following error:
```
ERROR: Cannot create candidate 966666/ROM999 as the profile file does not define a ProjectID for him/her.
```

Looking at the code, I realize that the config values from the Config module are not being fetched as far as I can tell. Maybe you did not push the latest version of the code?

Note to myself for upcoming testing once this PR is updated to resume testing: (test on ROM666_966666_V1 which does not exist in RB with default Project set to `Challah` and default Cohort set to `Stale`)
```
tarchiveLoader.pl -profile prod -uploadID 124 /data/demo/data/tarchive/DCM_2016-08-19_ImagingUpload-16-50-H_7N05.tar 
```

uploadNeuroDB/NeuroDB/objectBroker/ConfigOB.pm
```diff
@@ -93,6 +93,9 @@ use constant BIDS_DATASET_AUTHORS        => 'bids_dataset_authors';
 use constant BIDS_ACKNOWLEDGMENTS_TEXT   => 'bids_acknowledgments_text';
 use constant BIDS_README_TEXT            => 'bids_readme_text';
 use constant BIDS_VALIDATOR_OPTIONS_TO_IGNORE => 'bids_validator_options_to_ignore';
+use constant CREATE_CANDIDATES           => 'createCandidates';
+use constant CREATE_VISIT                => 'createVisit';
+use constant DEFAULT_COHORT              => 'default_cohort';
```
cmadjar, line 98: missing the following here:
```suggestion
use constant DEFAULT_COHORT              => 'default_cohort';
use constant DEFAULT_PROJECT             => 'default_project';
```

Which results in the following error:
```
Bareword "DEFAULT_PROJECT" not allowed while "strict subs" in use at /data/qpn/bin/mri/uploadNeuroDB/NeuroDB/objectBroker/ConfigOB.pm line 477.
```

laemtl: Thanks, it was probably introduced by a merge conflict not properly resolved 😄

laemtl: @cmadjar I will retest to make sure something else was not broken

laemtl:
@cmadjar Ready!

cmadjar (approved):
LGTM

PR #811, opened 2022-10-24
Fix imaging_install.sh

regisoc:
Fix a string that was not closed properly and made the script buggy at the end of the execution.


cmadjar (approved):
LGTM 👍

PR #823, opened 2022-11-03
Project, event validation and protobuf update

laemtl:
- add project validation
- update protobuf compiled file
- improve metadata validation

TODO
- [x] logic to set a default Site/Project when not provided in participants.tsv


python/lib/candidate.py
```diff
@@ -83,27 +83,66 @@ def create_candidate(self, db, participants_info):
...
+                if(len(db_site) > 0):
+                    self.center_id = db_site[0]['CenterID']
+
+            if self.center_id is None:
+                # try to find participant site in db
+                candidate_site_project = db.pselect(
+                    "SELECT RegistrationCenterID FROM candidate WHERE pscid = %s",
+                    [self.psc_id, ]
+                )
+                try:
```
laemtl, line 118: We can avoid a try/catch block by checking the length of the result: if(len(candidate_site_project) > 0):

python/lib/candidate.py
```diff
@@ -83,53 +84,95 @@ def create_candidate(self, db, participants_info):
...
+
             if 'site' in row:
+                # search site id in psc table by its full name
                 site_info = db.pselect(
                     "SELECT CenterID FROM psc WHERE Name = %s",
                     [row['site'], ]
                 )
                 if(len(site_info) > 0):
                     self.center_id = site_info[0]['CenterID']
-            else:
```
laemtl, line 93: I believe the logic inside this code block should work. Maybe it is skipped because the if part gets evaluated?

python/lib/candidate.py
```diff
@@ -83,53 +84,95 @@ def create_candidate(self, db, participants_info):
...
                 if(len(project_info) > 0):
                     self.project_id = project_info[0]['ProjectID']
 
+            if self.project_id is None:
+                # try to find participant project
+                candidate_site_project = db.pselect(
+                    "SELECT RegistrationProjectID FROM candidate WHERE pscid = %s",
+                    [self.psc_id, ]
+                )
+                try:
```
laemtl, line 142: Same comment, I believe a try/catch block is better to be avoided when possible.

python/lib/candidate.py
regisoc: Done, same logic as before. The same code is just below to match the third candidate site check.

cmadjar (approved):
LGTM

PR #844, opened 2022-12-02
Add chunk creation subprocess error check

### LINKED ISSUE #843: Capture chunk subprocess silent failure
Currently, if errors in subprocess call in python, errors and error logs are ignored. 
This makes the pipeline failing silently. Chunks are not created. No log nor errors raised.

- [x] capture error raised by subprocess call and raise them into the actual caller code

regisoc:
Raise error when the chunk creation subprocess fails.

Fix #843

### DIFF
--- a/python/lib/exitcode.py
+++ b/python/lib/exitcode.py
@@ -5,7 +5,7 @@ Perl ones that are in uploadNeuroDB/NeuroDB/ExitCodes.pm"""
 __license__ = "GPLv3"
 
 
-# -- Script ran successfully 
+# -- Script ran successfully
 SUCCESS = 0  # yeah!! Success!!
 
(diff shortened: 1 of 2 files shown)


cmadjar:
@regisoc thanks for the PR! Looks good to me but I wouldn't mind an extra pair of eyes to review this. @laemtl do you think you would have time to review as well? Thanks!

cmadjar:
@regisoc @laemtl should that go to the next bug fix release for 24.1 or to main? Will it be needed for HBCD? If so, maybe resend the PR to 24.1-release branch instead of main?

Let me know

regisoc:
It could be useful in the next bug fix

regisoc:
Closed to #848 -> 24.1-release

PR #855, opened 2022-12-12
Keep track of additional event data

regisoc:
## Context

Currently, we ignore additional event data from BIDS files. 
This PR allows keeping all these additional event data from BIDS import into a new table named `physiological_task_event_opt`.

Issue #835 
Requires [Loris 8237](https://github.com/aces/Loris/pull/8237)

Issue #835, opened 2022-11-14
Add support for additional event data


laemtl:
A praticular project may require to use custom event metadata. We can extend the table `physiological_task_event` with additional columns to support those.
 
- [ ] save all additional event data  present in _events.tsv if a corresponding column exists in `physiological_task_event`. 
- [ ] validate all metadata fields to match the db type.

regisoc:
The table `physiological_task_event` will get all required (base) info, and a new table e.g. `physiological_task_event_opt` should be added to include all optional and added columns in each `_event.tsv`.

PR #873, opened 2023-01-22
keep track of additional event data (24.1)

regisoc:
## Context

Rebase of #855  (main -> 24.1-release).
Currently, we ignore additional event data from BIDS files.
This PR allows keeping all these additional event data from BIDS import into a new table named `physiological_task_event_opt`.

Issue #835

Note: embeds a try/catch that should be removed when additional events are supported everywhere

Works with but should not require [LORIS 8237](https://github.com/aces/Loris/pull/8237)

### DIFF
--- a/python/lib/eeg.py
+++ b/python/lib/eeg.py
@@ -673,7 +673,8 @@ class Eeg:
                 # insert event data in the database
                 # TODO: Temporary
                 # TODO: should not have two different insert_event_file methods
-                # TODO: remove the TRY/CATCH and the legacy method when
+                # TODO: remove this TRY/CATCH and the legacy method when
+                # TODO: all LORIS code will support additional events.
                 insert_fallback = False
                 msg = ""
                 try:
@@ -681,19 +682,17 @@ class Eeg:
                         event_data, event_path, physiological_file_id, blake2
                     )
                 except NameError:
-                    # when fn does not exist
-                    msg = "WARNING: fn 'insert_event_file' not found. " \
-                          "Using fallback method."
+                    # when the insert_event_file function does not exist
+                    msg = "WARNING: function 'insert_event_file' not found. Using fallback method."
                     insert_fallback = True
                 except Exception as ex:
-                    # when table is not the good one
+                    # when event table structure is still the old one
                     if ex.args[0] and ex.args[0].startswith("Insert query failure: "):
-                        msg = "WARNING: error during DB insert." \
-                              "Using fallback method."
+                        msg = "WARNING: error during DB insert. Using fallback method."
                         insert_fallback = True
                     else:
                         # re-raise other errors from db insert
-                        raise
+                        raise ex
                 # insert fallback, call legacy method
                 if insert_fallback:
                     print(msg)
--- a/python/lib/eeg.py
+++ b/python/lib/eeg.py
@@ -671,9 +671,35 @@ class Eeg:
                 # get the blake2b hash of the task events file
                 blake2 = blake2b(event_file.path.encode('utf-8')).hexdigest()
                 # insert event data in the database
-                physiological.insert_event_file(
-                    event_data, event_path, physiological_file_id, blake2
-                )
+                # TODO: Temporary
+                # TODO: should not have two different insert_event_file methods
+                # TODO: remove the TRY/CATCH and the legacy method when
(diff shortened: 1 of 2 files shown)


python/lib/eeg.py
```diff
@@ -671,9 +671,35 @@ def fetch_and_insert_event_file(
                 # get the blake2b hash of the task events file
                 blake2 = blake2b(event_file.path.encode('utf-8')).hexdigest()
                 # insert event data in the database
-                physiological.insert_event_file(
-                    event_data, event_path, physiological_file_id, blake2
-                )
+                # TODO: Temporary
+                # TODO: should not have two different insert_event_file methods
+                # TODO: remove the TRY/CATCH and the legacy method when
```
cmadjar, line 676: I feel it is missing something after when in the comment ;)

python/lib/eeg.py
```diff
@@ -671,9 +671,35 @@ def fetch_and_insert_event_file(
...
+                    physiological.insert_event_file(
+                        event_data, event_path, physiological_file_id, blake2
+                    )
+                except NameError:
+                    # when fn does not exist
+                    msg = "WARNING: fn 'insert_event_file' not found. " \
+                          "Using fallback method."
+                    insert_fallback = True
+                except Exception as ex:
+                    # when table is not the good one
```
cmadjar, line 689:
```suggestion
                    # when event table structure is still the old one
```

python/lib/eeg.py
```diff
@@ -671,9 +671,35 @@ def fetch_and_insert_event_file(
...
+                    if ex.args[0] and ex.args[0].startswith("Insert query failure: "):
+                        msg = "WARNING: error during DB insert." \
+                              "Using fallback method."
+                        insert_fallback = True
+                    else:
+                        # re-raise other errors from db insert
+                        raise
```
cmadjar, line 696: shouldn't there be something after `raise`? Maybe `raise ex`?

python/lib/physiological.py
```diff
@@ -580,6 +586,150 @@ def insert_event_file(self, event_data, event_file, physiological_file_id,
...
+         :type physiological_file_id: int
+        :param blake2               : blake2b hash of the task event file
+         :type blake2               : str
+        """
+
+        # event_file_id = self.physiological_event_file_obj.insert(
+        #     physiological_file_id,
+        #     'tsv',
+        #     event_file
+        # )
```
cmadjar, line 616: those should be removed if commented.
```suggestion
```

python/lib/physiological.py
```diff
@@ -580,6 +586,150 @@ def insert_event_file(self, event_data, event_file, physiological_file_id,
...
+            duration = 0
+            if isinstance(row['duration'], (int, float)):
+                duration = row['duration']
+            else:
+                try:
+                    # try casting to float
+                    duration = float(row['duration'])
+                except ValueError:
+                    # value could be 'n/a'
+                    # put it at 0
```
cmadjar, line 663: it does not look like duration is set to 0 here instead of n/a. Unless I am misunderstanding the comment?

cmadjar (changes requested):
@regisoc a first review. Once addressed, @laemtl would you mind testing this PR?

python/lib/eeg.py
regisoc: Yes, it is! Good catch :)

python/lib/eeg.py
regisoc: By default, I think the exception handled by the bloc catching it will automatically be raised again if nothing is specified in the `raise`, but I can add it to be sure.

python/lib/physiological.py
regisoc: My comment is not clear. The variable `duration` is set to 0 before the if/else.

cmadjar (commented):
@laemtl would you have time to test this PR?

jeffersoncasimir:
Works great both with [LORIS PR #8237](https://github.com/aces/Loris/pull/8237) and without it (fallback/legacy implementation)

jeffersoncasimir (approved):
Works great and tested both with [LORIS PR #8237](https://github.com/aces/Loris/pull/8237) and without it (fallback/legacy implementation)

cmadjar (approved):

Issue #841, opened 2022-11-21
HDF5 support (Matlab files) - scipy 


christinerogers:
(placeholder @cmadjar don't hesitate to edit)

Matlab 7.3+ files are not supported by SciPy -  no HDF5 Python library is integrated. 
Loris-MRI has a dependency on SciPy in the pipeline, and it look like it fails on .mat files - likely to impact all projects trying to load .mat data straight out of many EEG suites. 
Since it's not easy (or ideal) to switch out libraries for these functions -- reaching out to SciPy (Nov 2022) is the likely next step. 
There doesn't seem to be any chatter in the SciPy forum/list about this yet. 

reported November 4 by Steven Yeu, Scott Makeig's NEMAR development group, based at UCSD: 

> I updated SciPy to 1.9.3 but still ran into the same error. I was reading the [scipy docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.loadmat.html) and it said that SciPy doesn’t support reading MATLAB 7.3 files. It says a HDF5 Python library is needed to load them and SciPy doesn’t have one. Does that mean the Loris pipeline scripts don’t support this as well?
> 
> I am using Loris-MRI-24.0.2. For Loris, I am using v24.0.3.

cmadjar:
see issue created in the scipy repo for reference: https://github.com/scipy/scipy/issues/17509

christinerogers:
I think JC might have some time to explore this in Jan-Feb since we expect to handle similar (recent matlab) files for EEGNet.  Might be able to quickly contribute a fix to Loris if it's straightforward.

christinerogers:
update (fyi @cmadjar) - in progress at the link below, jc has a fix basically, it's just a question of when/where it should go, cheching for side effects, and having the NEMAR folks test it thoroughly: https://github.com/aces/Loris-MRI/compare/main...jeffersoncasimir:Loris-MRI:hdf5_temporary_solution

christinerogers:
in testing with NEMAR team

christinerogers:
update per MCIN EEG meeting - likely this will be pushed as a little fix on the 24 branch and added in the next bug-fix release.   
Status/priority will be clearer in very early February when we have news on its success from S.Makeig's NEMAR development team.

cmadjar:
@christinerogers @jeffersoncasimir should this be closed?

PR #880, opened 2023-02-03
MATLAB 7.3 HDF5 Workaround

jeffersoncasimir:
This workaround allows MATLAB 7.3 HDF5 files to work with the pipeline despite not being supported by SciPy.
It assumes that the file path fields are already correctly set.

Related to #841 and in testing by NEMAR team, per MCIN-EEG meeting Jan.27

# Caveat for existing projects

- make sure to install the `mat73` python library using `pip install mat73` after having source the LORIS-MRI `environment` file

### DIFF
--- a/python/lib/utilities.py
+++ b/python/lib/utilities.py
@@ -172,6 +172,8 @@ def update_set_file_path_info(set_file, fdt_file):
 
     # grep the basename without the extension of set_file
     basename = os.path.splitext(os.path.basename(set_file))[0]
+    set_file_name = numpy.array(basename + ".set")
+    fdt_file_name = numpy.array(basename + ".fdt")
 
     try:
         # read the .set EEG file using scipy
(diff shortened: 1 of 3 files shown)


python/lib/utilities.py
```diff
@@ -173,22 +173,25 @@ def update_set_file_path_info(set_file, fdt_file):
...
+        if 'EEG' in dataset.keys():
+            dataset['EEG'][0][0][1] = numpy.array(basename + ".set")
+        if fdt_file and 'EEG' in dataset.keys():
+            dataset['EEG'][0][0][15] = numpy.array(basename + ".fdt")
+            dataset['EEG'][0][0][40] = numpy.array(basename + ".fdt")
+
+        # write the new .set file with the correct path info
+        scipy.io.savemat(set_file, dataset, False)
+    except NotImplementedError:     # Thrown for matlab v7.3 files
+        pass    # Assumes path fields are already set
```
cmadjar, line 194: @jeffersoncasimir Thank you for the PR!!! Would it be worth to have a warning printed here mentioning that the pipeline assumes that the path fields are already set?

jeffersoncasimir: @cmadjar Yes, but upon reconsideration, I think I should change it so that we aren't assuming that the fields are properly set, and we should instead use a library that can read the fields and only allow it on that condition. This could otherwise potentially introduce a vector for referencing unauthorized data.

christinerogers: Thanks, I'll add to the Loris-EEG agenda for this Friday - Jefferson could you bring an estimate of how long will this approach take?

cmadjar:
@jeffersoncasimir thank you for the PR!!!

@christinerogers the fix should be in the next bug fix release of 24.1. I will most likely do a release early next week.

christinerogers:
fantastic! thanks everyone.

PR #882, opened 2023-02-09
No more Subproject

regisoc:
Following renaming in [LORIS #7817](https://github.com/aces/Loris/pull/7817)
Change "subproject" to "cohort" naming in LORIS-MRI.

# Caveat for existing projects

`SubprojectID` has been renamed `CohortID` (and table `subproject` was renamed `cohort`). This refactoring implies the following:
- `participants.tsv` file of incoming BIDS datasets should have column `subproject` renamed to `cohort`
- update your `prod` and `database_config.py` files if `SubprojectID` was determined by the get subject ID function so that it uses `CohortID ` instead.


regisoc:
We need to also change BIDS import references (i.e. column `subproject` in `participants.tsv` file). 
The ref that are linked to this file are not changed but listed here:
- [ ] docs/05-PipelineLaunchOptions.md
- [ ] python/lib/mri.py
- [ ] python/lib/eeg.py
- [ ] python/bids_import.py

python/bids_import.py
```diff
@@ -271,8 +273,8 @@ def create_loris_bids_directory(bids_reader, data_dir, verbose):
 
     # determine the root directory of the LORIS BIDS and create it if does
     # not exist
-    name = re.sub("[^0-9a-zA-Z]+", "_", bids_reader.dataset_name) # get name of the dataset
-    version = re.sub("[^0-9a-zA-Z\.]+", "_", bids_reader.bids_version) # get BIDSVersion of the dataset
+    name = re.sub(r"[^0-9a-zA-Z]+", "_", bids_reader.dataset_name)       # get name of the dataset
```
ridz1208, line 276: @regisoc was this change and the one below it on purpose?

ridz1208:
@cmadjar I went through and checked that the casing is consistent with the LORIS changes. Looks good other than the one comment on my end

python/bids_import.py
regisoc: Hmm, no. I added on another branch but I am pretty sure I created this one based on main.

ridz1208: okai but they are not accidental, thats the important part. If you can rebase and remove this commit `rebase -i` it would be better I think but @cmadjar will do the final review so it's her call

regisoc: Ok, there are 2 things here: 1/ the "r" intends to put the following string as a regex to nullify the backslash (else, flake8 raises a `W605` error) and 2/ it should be at least 2 spaces after for an inline comment (else, flake8 raises a `W261` error). But, neither `W605` nor `W261` are in the flake8 workflow exclusion https://github.com/aces/Loris-MRI/blob/06aa4125ac19f9f4601a8b1e7c0225f3140b8cae/.github/workflows/flake8_python_linter.yml#L38 So, do we add them?

ridz1208: 👀 huh?

@cmadjar ?

cmadjar: yeah, those 2 changes are fine and compliant with flake 8 (the phpcbf of python)

cmadjar (changes requested):
@regisoc There are a lot of spaces changes that should not be present in the PR. Could you please rebase or resubmit without all those space changes. It makes the PR big and difficult to review properly. Thanks!

regisoc:
Yes, I believe this is my editor that auto remove them. I will change that.

cmadjar (changes requested):
@regisoc see below for review comments.

docs/05-PipelineLaunchOptions.md
```diff
@@ -158,13 +158,13 @@ The BIDS dataset to import has to:
   ```
 
 
-In cases where the `site`, `project` or `subproject` are missing from the `participant.tsv` file, `NULL` will be populated for those fields in the candidate and session tables. In those cases, later manual addition is required in order to display the entry on the frontend.
+In cases where the `site`, `project` or `cohort` are missing from the `participant.tsv` file, `NULL` will be populated for those fields in the candidate and session tables. In those cases, later manual addition is required in order to display the entry on the frontend.
```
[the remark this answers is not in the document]
regisoc: Agreed. I think we can go for Cohort everywhere, else it will be messy even for us.
But BIDS -> `participants.tsv` files should also be noted to change their columns for `cohort` instead of `subproject`.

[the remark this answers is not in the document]
regisoc: Seems ok for @laemtl too! I will apply the changes.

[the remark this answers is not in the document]
cmadjar: @regisoc see note that I added in the PR description. Let me know if you think I may have missed something. Thanks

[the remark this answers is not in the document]
regisoc: LGTM!

cmadjar:
Tested BIDS import pipeline and python dcm2bids pipeline. It all works as expected.

Did not have a chance yet to test the perl pipeline though...

cmadjar (approved):
LGTM 👍 Thank you @regisoc for submitting the PR!!

cmadjar:
Tested the perl pipeline as well.

PR #886, opened 2023-02-13
[bids_import] Revised .set insertion logic

jeffersoncasimir:
The .set file must be copied before updating the file paths


cmadjar (approved):
LGTM

PR #888, opened 2023-02-15
Merge 24.1.8 into main

cmadjar:
This merges the latest bug fix release of 24.1 into the main branch.


regisoc:
Tested for BIDS import: do not know if it is worth mentioning but got some repeated warnings during execution.

```bash
/opt/hbcd/bin/mri/python/react-series-data-viewer/eeglab_to_chunks.py:9: RuntimeWarning: Data will be preloaded. preload=False or a string preload is not supported when the data is stored in the .set file
  return mne.io.read_raw_eeglab(path, preload=False)
/opt/hbcd/bin/mri/python/react-series-data-viewer/eeglab_to_chunks.py:9: RuntimeWarning: The data contains 'boundary' events, indicating data discontinuities. Be cautious of filtering and epoching around these events.
  return mne.io.read_raw_eeglab(path, preload=False)
```

regisoc:
Only execution error [found here](https://github.com/aces/Loris-MRI/blob/0c1d6abe46ddbb2a379dd6f209472c51c592c03f/python/lib/eeg.py#L754)
Change `physiological_file_id` by `str(physiological_file_id)`.

cmadjar:
> Tested for BIDS import: do not know if it is worth mentioning but got some repeated warnings during execution.
> 
> ```shell
> /opt/hbcd/bin/mri/python/react-series-data-viewer/eeglab_to_chunks.py:9: RuntimeWarning: Data will be preloaded. preload=False or a string preload is not supported when the data is stored in the .set file
>   return mne.io.read_raw_eeglab(path, preload=False)
> /opt/hbcd/bin/mri/python/react-series-data-viewer/eeglab_to_chunks.py:9: RuntimeWarning: The data contains 'boundary' events, indicating data discontinuities. Be cautious of filtering and epoching around these events.
>   return mne.io.read_raw_eeglab(path, preload=False)
> ```

@laemtl are those warnings expected on main?

laemtl:
> @laemtl are those warnings expected on main? I don't remember...

Since preload is not supported on eeglab files, I will set it to true to prevent this warning.
The boundary one is ok.

PR #900, opened 2023-03-08
Keep additional events - rename Task to Property

regisoc:
Change the `TaskName` and `TaskValue` to `PropertyName` and `PropertyValue` for additional event integration to match BIDS naming.

Linked to [LORIS 8432](https://github.com/aces/Loris/pull/8432)

### DIFF
--- a/python/lib/physiological.py
+++ b/python/lib/physiological.py
@@ -704,8 +704,8 @@ class Physiological:
                 # table cols
                 add_event_fields = (
                     'PhysiologicalTaskEventID',
-                    'TaskName',
-                    'TaskValue'
+                    'PropertyName',
+                    'PropertyValue'
                 )
                 # each additional fields is a new entry
                 add_event_values = []


jeffersoncasimir (approved):

python/lib/physiological.py
```diff
@@ -704,8 +704,8 @@ def insert_event_file(self, event_data, event_file, physiological_file_id,
                 # table cols
                 add_event_fields = (
                     'PhysiologicalTaskEventID',
```
christinerogers, line 706: Could `PhysiologicalTaskEventID` be changed too?

christinerogers (changes requested):
see my comments on #8432 as well in Loris repo

cmadjar:
Blocked by https://github.com/aces/Loris/pull/8432 which need to be merged at the same time as this PR on main.

cmadjar (changes requested):
@regisoc isn't this supposed to go on main since the SQL changes go to main on the LORIS side?

regisoc:
Yes, but we agreed on the `try/catch` bloc in `LORIS-MRI` that will make the routing depending on the code version in `LORIS`: `24.1` does not have the SQL change on LORIS, but `main` has. I was told to push this on `24.1` but I can change it if that is not convenient anymore.

cmadjar:
@regisoc right, so if SQL is the main version, the code changed here will be run, otherwise it will still run the old code. Sounds good then

cmadjar:
@christinerogers is that OK with you to get merged? I see you requested changes but the sentence says `looks ok to me...`. If all good, could you please edit your review and approve the PR? Thanks!

christinerogers (approved):

PR #912, opened 2023-04-03
candID subject_id support

laemtl:
- Adds support for the CandID in the participants.tsv's subject_id fields
- Adds support for the CAND_ID/PSCID validation (match) in the folder name if the flag -i is used


cmadjar (changes requested):
@laemtl One tiny change, otherwise works great.

cmadjar (approved):
LGTM 👍 Thanks for implementing this @laemtl!

PR #915, opened 2023-04-14
Blake2b fix

### LINKED ISSUE #894: [BIDS Import] Validate hash  behaviour
For each file, a hash is generated to allow for future data integrity check. At the moment, it seems like the hash function is called on the file path, not the file data.

Example:
https://github.com/aces/Loris-MRI/blob/b9d7ee4f8a8d353673d9e1653855d3a40f5b4a7c/python/lib/eeg.py#L451

Check if this is the case, and correct if necessary.

cmadjar:
# Description

This fixes the Blake2b and MD5 hashing in the python scripts. Previously, the code was hashing the path instead of the data content. Now the data content is hashed as it should have been.

Resolves #894

### DIFF
--- a/python/lib/bidsreader.py
+++ b/python/lib/bidsreader.py
@@ -129,7 +129,11 @@ class BidsReader:
             else:
                 continue
 
-        self.candidates_list_validation(participants_info)
+        if participants_info:
+            self.candidates_list_validation(participants_info)
+        else:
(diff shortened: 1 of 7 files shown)


cmadjar:
Rebase of this PR is https://github.com/aces/Loris-MRI/pull/916

PR #916, opened 2023-04-14
Fix the blake2b and md5 hashing in python bids_import and dcm2bids pipelines

cmadjar:
# Description 

This fixes the Blake2b and MD5 hashing in the python scripts. Previously, the code was hashing the path instead of the data content. Now the data content is hashed as it should have been.

Resolves https://github.com/aces/Loris-MRI/issues/894

# Notes for existing projects

Since the hashes were not created properly by the python pipelines (A.K.A. bids_import.py and dcm2bids pipelines), the hashes stored in the database are incorrect and need to be corrected. In order to do that, a script has been made available in the tools directory. To correct all the incorrect MD5 and BLAKE2b hashes, run the following command:

```
python tools/correct_blake2b_and_md5_hashes_in_database.py -p database_config.py
```


cmadjar:
@laemtl finally ready for you if you want to test the EEG pipeline :-)

I tested the dcm2bids pipeline and the tool script on my sandbox and it seems to behave as expected. Let me know if you have any questions.

laemtl:
I tested the fix and works well. The hash generated is now reproducible.

PR #966, opened 2023-04-28
[petupload_cron_prod] Automatically create dirs

laemtl:
- Create directory if they do not exist
- append time to the file name to avoid conflicts if the script run more than once a day


cmadjar (approved):
LGTM 👍

PR #968, opened 2023-05-01
[bids_import] Added check for 'HED' field to prevent KeyError

jeffersoncasimir:
Checking first for the HED field is necessary because the 'HED' field is optional as per the BIDS specification.


cmadjar (approved):
LGTM

PR #984, opened 2023-05-08
[imaging_non_minc_insertion] Error message obtained when an invalid upload ID is specified on the command lineline

nicolasbrossard:
This PR removes the following error message obtained when running script `imaging_non_minc_insertion.pl` with an invalid argument to the `-upload_id` option:
```
DBD::mysql::st execute failed: Incorrect integer value: '<whatever_invalid_upload_id>' for column 'ProcessID' at row 1 at /data1/loris/bin/mri/uploadNeuroDB/NeuroDB/Notify.pm line 130.
```

The validation of the upload ID needs to happen **before** calling `NeuroDB::Notify->spool()` as this call requires a valid upload ID


laemtl:
@nicolasbrossard Can you rebase?

PR #993, opened 2023-06-14
Fix MriScanner::determine_scanner_information description

### LINKED ISSUE #992: register_new_scanner config not used
The `determine_scanner_information` method in the class `MriScanner` has the following description:
```
Select a ScannerID based on the scanner information gathered from the headers of the
DICOM archive. If a ScannerID is not found for the scanner but register_new_scanner
is set to True in the Config table, a new entry will be inserted into the mri_scanner table.
```
but the config `register_new_scanner` does not exist (see https://github.com/aces/Loris/pull/5954).

laemtl:
Resolve #992


regisoc (approved):
LGTM

PR #989, opened 2023-05-10
[register_processed_data] Validation of input file ID list

nicolasbrossard:
This PR adds a validation step in script `register_processed_data.pl` to ensure the argument for option `inputFileIDs` is a list of existing `FileID`s found in the `files` table.

Fixes #990

PR #999, opened 2023-07-05
[register_processed_data] Add more validation for options -sourceFileID and -inputFileIDs

nicolasbrossard:
This PR adds validation to the `-sourceFileID` and `-inputFileIDs` options of script `register_processed_data.pl`:

1. Validates that the argument for `-sourceFileID` is a file ID that exists in table `files`.
2. Validates that the argument for `-inputFileIDs` is a list of file IDs separated by `;` and that each file ID exists in table `files`.
3. Validates that each file ID contained in the list described above is not equal to the source file ID (see 1).

Fixes #989


laemtl (approved):
Changes look good, TYSM!

PR #1015, opened 2023-08-11
Add method to check and skip duplicate content uploads to S3

breen0074:
# Description

The changes here are intended to check to see if the content of file that would be uploaded to S3 has already been uploaded.  It does this by checking to see if the hash of a file content is already available at the targeted S3 object key location before attempting to upload new content.  If it already exists, it will skip it. 

This helps to resolve an issue where sometimes the same content would be uploaded to an S3 bucket, even if that file already existed.  Normally this would be fine, but in versioning enabled buckets this creates duplicate copies of the files when no changes are needed.

This does not appear to cause any breaking changes.

### DIFF
--- a/python/lib/aws_s3.py
+++ b/python/lib/aws_s3.py
@@ -1,9 +1,9 @@
 """This class interacts with S3 Buckets"""
 
 import boto3
+import lib.utilities
 import os
 from botocore.exceptions import ClientError, EndpointConnectionError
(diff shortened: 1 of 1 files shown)


cmadjar:
Closing in favour of https://github.com/aces/Loris-MRI/pull/1032 (which is identical to this PR but rebased on 24.1-release branch instead of main)

PR #1017, opened 2023-10-10
[bids_import] Add single electrode value instead of all electrode values

jeffersoncasimir:
This fixes a bug in electrode insertion, where more than one electrode was being inserted every time.

### DIFF
--- a/python/lib/physiological.py
+++ b/python/lib/physiological.py
@@ -368,7 +368,6 @@ class Physiological:
             'Impedance',
             'FilePath'
         )
-        electrode_values = []
         electrode_ids = []
         optional_fields = ('type', 'material', 'impedance')
         for row in electrode_data:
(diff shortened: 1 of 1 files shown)


christinerogers:
@jeffersoncasimir want to request a reviewer?  maybe laetitia or regis review this, if it's straightforward
Mark it high priority for the next release if that makes sense too.

regisoc (commented):
Makes sense. If that is the case, lines containing electrode values ([L371](https://github.com/aces/Loris-MRI/blob/4e19fd9420921894a0a0983e7c5a0872a8c82906/python/lib/physiological.py#L371), [L412](https://github.com/aces/Loris-MRI/blob/4e19fd9420921894a0a0983e7c5a0872a8c82906/python/lib/physiological.py#L412C13-L412C29)) are not useful anymore and can be removed

jeffersoncasimir:
Good point. I just removed those two references.

regisoc (approved):
LGTM

christinerogers:
this PR is a `High Priority` `Bug fix` for the 26 release.

PR #1023, opened 2023-10-16
Aws upload dir subdir fix

laemtl:
Fix 
==> tmp_path could not be uploaded to the S3 bucket. Error was /subdir/file processing failure, must be a s3 url.

--> Cause: os.path.join() called on a string with leading / will return that string.
--> Solution: remove the leading string.


cmadjar (approved):
Makes sense! Thanks @laemtl

PR #1024, opened 2023-10-17
Fixes for parameter file value lists 

breen0074:
# Description

The changes here are intended to help address two different issues surrounding bids_import and parameter files.

- When importing parameter_file values that are a list, do not convert the entire list to a string before join(d57969b75875284e69b9f39cd27f5c66f87e768d).  This ensures that the list is joined correctly instead of creating a list of every character from the original value.
- Check to ensure parameter_file is not empty before attempting to make any changes to int, or float lists(c5b411d0614b08d23367465fbaea0f8e7301eaf6).  This prevents the `bids_import.py` script from failing if a parameter value contains an empty list.


breen0074:
Closing to create a new pull request to the proper branch per request.

PR #1045, opened 2024-03-07
[EEG BIDS] Support assembly_bids source for datasets

laemtl:


python/bids_import.py
```diff
@@ -218,8 +275,9 @@ def read_and_insert_bids(bids_dir, config_file, verbose, createcand, createvisit
         grep_candidate_sessions_info(
             bids_sessions, bids_id,    cand_id,       loris_bids_root_dir,
             createvisit,   verbose,    db,            default_bids_vl,
-            center_id,     project_id, cohort_id
+            center_id,     project_id, cohort_id,     nocopy
         )
+        # TODO: Is this deadcode?
```
cmadjar, line 280: fascinating. It sounds like dead code indeed.

python/lib/bidsreader.py
```diff
@@ -1,13 +1,12 @@
...
 
 import re
-import os
 import sys
 import json
 
 import lib.exitcode
 import lib.utilities as utilities
-import bids
+# import bids
```
cmadjar, line 9: shouldn't this be removed instead of commented out? Or at list moved in the `load_bids_data` function?

python/lib/bidsreader.py
```diff
@@ -19,12 +18,12 @@
...
-        or bids_pack_version[1] > 12
-        or (bids_pack_version[1] == 12 and bids_pack_version[2] > 0)):
-    
-	from bids import BIDSLayoutIndexer
+# bids_pack_version = list(map(int, bids.__version__.split('.')))
+# if (bids_pack_version[0] > 0
+#     or bids_pack_version[1] > 12
+#    or (bids_pack_version[1] == 12 and bids_pack_version[2] > 0)):
+
+# 	from bids import BIDSLayoutIndexer
```
cmadjar, line 26: shouldn't this be removed instead of commented out? It seems to be one in the `load_bids_data` function now

python/lib/dcm2bids_imaging_pipeline_lib/nifti_insertion_pipeline.py
```diff
@@ -431,7 +431,6 @@ def _determine_new_nifti_assembly_rel_path(self):
         bids_cand_id = 'sub-' + self.subject_id_dict['CandID']
         bids_visit = 'ses-' + self.subject_id_dict['visitLabel']
         bids_subfolder = self.bids_categories_dict['BIDSCategoryName']
-        new_nifti_rel_dir = os.path.join('assembly_bids', bids_cand_id, bids_visit, bids_subfolder)
```
cmadjar, line 434: I think that is going to break the dcm2bids pipeline. For dcm2bids pipeline, we need to have that rel path for projects not using S3

python/requirements.txt
```diff
@@ -8,7 +8,7 @@ google
 matplotlib
 nose
 nilearn
-sklearn
+scikit-learn
```
cmadjar, line 11: why is that changed?

cmadjar (changes requested):
A few comments.

python/lib/bidsreader.py
laemtl: It will be probably uncommented whenever I spent time on this

python/lib/dcm2bids_imaging_pipeline_lib/nifti_insertion_pipeline.py
laemtl: It was flagged as an unused var by flake8. Where is it used?

python/requirements.txt
laemtl: The 'sklearn' PyPI package is deprecated, use 'scikit-learn'
      rather than 'sklearn' for pip commands.

PR #1117, opened 2024-04-25
New DICOM study import script

maximemulder:
# Description

This PR is a rewrite in Python (+ cleaning up / refactor) of the DICOM archiving scripts in Python, with a few added features.

In more details, the changes are the following:
* `dicom-archive/dicomTar.pl` → `python/scripts/import_dicom_study.py`
* `dicom-archive/dicomSummary.pl` → `python/scripts/summarize_dicom_study.py`
* `dicom-archive/DICOM/DCMSUM.pm` (and a bunch of other code) → `python/lib/import_dicom_study/*`

This PR aims to be feature-compatible with the current Perl scripts, such that it can eventually replace it seemlessly. It removes some arguments that were unused (`-centerName`, `-today` and `-addMriUpload`), sorts the DICOM archive in a year subdirectory automatically (this was previously done later in the pipeline), and automatically use the `tarchive` directory as a target.

This PR also adds the `--session` argument to determine the DICOM archive session based on the Python configuration file (and fail if no session is found).

# Testing

Tested all options successfully with Raisinbread and MPN data. The data looks identical with both the old DICOM archiver and this new script.

# Potential future work

- Replace the old DICOM archiver by the new script within the existing MRI pipeline.
- Clean up the `tarchive` table.
- Add integration test for CI.

# Notes for existing projects

- need to install `pydicom` library `pip install pydicom`


MaximeMulder:
Blocked on Python 3.10, as well as #1141 and #1142 to see if this PR should be updated with new libraries.

python/dicom_archive.py
```diff
@@ -0,0 +1,325 @@
...
+import lib.exitcode
+
+
+def print_error_exit(message: str, code: int):
+    print(f'ERROR: {message}', file=sys.stderr)
+    sys.exit(code)
+
+
+def print_warning(message: str):
+    print(f'WARNING: {message}', file=sys.stderr)
```
cmadjar, line 27: Did you check the log.py library for the messaging?

python/dicom_archive.py
```diff
@@ -0,0 +1,325 @@
...
+                (
+                    f'File or directory \'{path}\' already exists. '
+                    'Use option \'--overwrite\' to overwrite it.'
+                ),
+                lib.exitcode.TARGET_EXISTS_NO_CLOBBER,
+            )
+
+
+# Modified version of 'lorisgetopt.load_config_file'.
+# We use argparse to parse the command line options in this script,
```
cmadjar, line 45: I think it would be best to use the `lorisgetopt.py` library to parse arguments since it checks and has a few other loris specific functionalities. Also, it avoids the duplicated code for `load_config_file.py`. It also avoids having to add another dependency.

python/lib/dicom/dicom_database.py
```diff
@@ -0,0 +1,231 @@
+from datetime import datetime
```
cmadjar, line 1: This file should go under the `database_lib` directory. This is where we store the DB interaction files. The files are organized by table names and holds the table specific queries. It would be best to keep that organization.

python/dicom_archive.py
```diff
@@ -0,0 +1,325 @@
...
+check_create_file(log_path)
+
+print('Extracting DICOM information (may take a long time)')
+
+summary = lib.dicom.summary_make.make(source, verbose)
+
+if db is not None:
+    print('Checking database presence')
+
+    archive = lib.dicom.dicom_database.get_archive_with_study_uid(db, summary.info.study_uid)
```
cmadjar, line 205: Maybe rename to db_archive to make is clear that this is the archive found in the DB?

```suggestion
    db_archive = lib.dicom.dicom_database.get_archive_with_study_uid(db, summary.info.study_uid)
```

python/dicom_archive.py
```diff
@@ -0,0 +1,325 @@
...
+
+print('Copying into DICOM tar')
+
+with tarfile.open(tar_path, 'w') as tar:
+    for file in os.listdir(source):
+        tar.add(source + '/' + file)
+
+print('Calculating DICOM tar MD5 sum')
+
+tarball_md5_sum = lib.dicom.text.make_hash(tar_path, True)
```
cmadjar, line 234: You should use the function `compute_md5_hash` in lib.utilities to compute the hash.

python/dicom_archive.py
```diff
@@ -0,0 +1,325 @@
...
+
+print('Zipping DICOM tar (may take a long time)')
+
+with open(tar_path, 'rb') as tar:
+    with gzip.open(zip_path, 'wb') as zip:
+        shutil.copyfileobj(tar, zip)
+
+print('Calculating DICOM zip MD5 sum')
+
+zipball_md5_sum = lib.dicom.text.make_hash(zip_path, True)
```
cmadjar, line 244: You should use the function `compute_md5_hash` in `lib.utilities` to compute the hash.

python/dicom_archive.py
```diff
@@ -0,0 +1,325 @@
...
+
+print('Getting DICOM scan date')
+
+if not today and summary.info.scan_date is None:
+    print_warning((
+        'No scan date found for this DICOM archive, '
+        'consider using argument \'--today\' to use today\'s date instead.'
+    ))
+
+scan_date = date.today() if today else summary.info.scan_date
```
cmadjar, line 254: I don't think we should use the date the script is run at a DICOM date if there is no date information found in the DICOMs. @nicolasbrossard thoughts on this?

I understand that this is to organize the DICOMs in the year subfolder but if you rerun dicomTar.pl a year later, you might end up with duplicated tar.gz file in the file system. Sounds like a recipe for confusion and duplicated files.

python/dicom_archive.py
```diff
@@ -0,0 +1,325 @@
...
+    if not os.path.exists(dir_path):
+        print(f'Creating directory \'{dir_path}\'')
+        os.mkdir(dir_path)
+    elif not os.path.isdir(dir_path) or not os.access(dir_path, os.W_OK):
+        print_error_exit(
+            f'Path \'{dir_path}\' exists but is not a writable directory.',
+            lib.exitcode.CREATE_DIR_FAILURE,
+        )
+else:
+    dir_path = target
```
cmadjar, line 275: I am not super confortable using the date the pipeline is run to organize the DICOM.tar.gz files. What happens if the data gets run again through the pipeline on a different year? You would end up with duplicated files on the file system and that might cause confusions. Also, DICOMs are organized by acquisition date year later on in the pipeline, so that might mess up later code.

@nicolasbrossard, @laemtl  thoughts in this?

python/lib/dicom/dicom_log.py
```diff
@@ -0,0 +1,97 @@
+from datetime import datetime
+import os
+import socket
+from lib.dicom.text_dict import DictWriter
+
+
+class Log:
```
cmadjar, line 7: there is already a class `Log` that exists. I would rename this one `DicomLog` instead to make it clear this is for the DICOM log file.

python/dicom_summary.py
```diff
@@ -0,0 +1,36 @@
...
+import lib.dicom.summary_write
+import lib.exitcode
+
+parser = argparse.ArgumentParser(description=(
+        'Read a DICOM directory and print the DICOM summary of this directory '
+        'in the the console.'
+    ))
+
+parser.add_argument(
+    'directory',
```
cmadjar, line 17:
```suggestion
    '--directory',
```

However, it might be best to use the lorisgetop library we already have instead of adding a new dependency

python/lib/dicom/summary_make.py
```diff
@@ -0,0 +1,218 @@
...
+    """
+    if os.path.isdir(prefix + '/' + path):
+        files = []
+        for file in os.listdir(prefix + '/' + path):
+            files += get_dir_files(prefix, path + '/' + file)
+
+        # Flatten the lists of files
+        return files
+
+    return [path]
```
cmadjar, line 94: This function could be moved to `lib.utilities` and could be used by other processes since it is not specific to DICOMs.

python/dicom_archive.py
```diff
@@ -0,0 +1,325 @@
...
+
+print('Calculating DICOM tar MD5 sum')
+
+tarball_md5_sum = lib.dicom.text.make_hash(tar_path, True)
+
+print('Zipping DICOM tar (may take a long time)')
+
+with open(tar_path, 'rb') as tar:
+    with gzip.open(zip_path, 'wb') as zip:
+        shutil.copyfileobj(tar, zip)
```
cmadjar, line 240: Could `copy_file` from `lib.utilities` be used here? Maybe not since you are using copyfileobj. Just wondering

cmadjar (changes requested):
@MaximeMulder I read through the code and have a few comments/refactoring suggestions. Let me know your thoughts.

Note: I did not test yet, just went through the code.

python/dicom_archive.py
MaximeMulder: I did this to match the `-today` option that currently exists in `dicomTar.pl`. I am fine with removing this option in the new script, but then it would make it impossible to insert DICOMs that do not have a scan date (unless they are patched manually to add a scan date), is this an outcome we want ?

MaximeMulder: I added this `--today` option to match the `-today` option in the `dicomTar.pl`. But you are probably right that it might be mistake. I guess if there is no scan date we just put the archive in the root `tarchive` directory ? If the `--year` option is used in this case, should we just give a warning or abort the script execution ?

python/dicom_archive.py
MaximeMulder: I am not a fan of `lorisgetopt.py` design, but you are probably right that it is best to avoid duplication, I'll do it.

python/lib/dicom/summary_make.py
```diff
@@ -0,0 +1,218 @@
...
+        return None
+
+    return dicom[tag].value or None
+
+
+def cmp_int_none(a: int | None, b: int | None):
+    """
+    Order comparison between two nullable integers.
+    """
+    match a, b:
```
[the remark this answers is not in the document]
MaximeMulder: Both options are fine with me, tell me if you have a preference. If we want to be really rigorous though, it might be good to update as 3.8 should (theoretically) soon become EOL.

https://devguide.python.org/versions/

python/lib/dicom/summary_make.py
```diff
@@ -0,0 +1,218 @@
...
+        case None, None:
+            return 0
+        case _, None:
+            return -1
+        case None, _:
+            return 1
```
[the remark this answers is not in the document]
MaximeMulder: `_` is the wildcard pattern, which kind of means "anything works" (so any `Some(x)` such as `Some(0)`, `Some(1)`... work here, since `None` was already matched before).
The returns `-1`, `1`... are because this is a comparison function, a negative number means that `a < b`, `0` that `a == b`, and a positive number that `a > b`. This is required by `cmp_to_key` in `summary_make` but I'll add some documentation.

python/dicom_summary.py
MaximeMulder: Not using `--` means that it is a positional argument instead of a named one. I would think this is better here `dicom_archive.py <source_dir> <target_dir>` ?

MaximeMulder:
Thanks for the review @cmadjar ! I wrote some comments on a few points. Points with no comments mean that I agree or will look into it.

cmadjar:
@MaximeMulder I think if you rebase this one, it will pass the flake8 now that your PR with the python install update has been merged. Thanks!

Also, regarding the `-mri_upload_update`, this option is used by projects that do not use the imaging uploader to upload their scan and run directly `dicomTar.pl`. I think we should still support that functionality as some projects might still be using the old way of importing DICOMs. Otherwise, the script to create the MINC or BIDS dataset would break because there is no entry in the `mri_upload` table. Actually, before you do this change, could you bring it up during the next imaging call? It should be next week on August 8th. Thank you :)

python/dicom_archive.py
MaximeMulder: I tried to do that and encountered two problems:
- The behaviour of `dicomTar.pl` is to not only write the hash, but also the filename (`"filename    hash"`). `make_hash` emulates this behaviour but not `compute_md5_hash` (although I would not be against writing just the hash).
- `compute_md5_hash` also works on non-existing files, while `make_hash` does not. This makes typing (rightfully) unhappy because I need a `str`, not a `str | None` (`compute_md5_hash` docstring is incorrect with regards to its real return type).
- which makes typing (rightfully) unhappy

python/dicom_archive.py
MaximeMulder: Same comment as previous point.

python/lib/dicom/dicom_log.py
MaximeMulder: That makes sense, renamed to `DicomArchiveLog` !

python/lib/dicom/summary_make.py
MaximeMulder: Good point, done !

python/dicom_archive.py
MaximeMulder: That makes sense, done !

MaximeMulder:
Ok I replied to all your (very good !) comments @cmadjar. The only ones I have intentionally left out are these ones :
1. Move SQL queries to `database_lib` (waiting on a decision on SQLAlchemy at the next imaging meeting).
2. Add the `--db-upload` parameter (done for a while on my SQLAlchemy branch)
3. Use `log.py` (full implementation depends on `--db-upload`)

If you like the rest of the PR, I can add these very quickly after the next imaging meeting.

laemtl:
Waiting for the first test implementation to polish it.

python/dicom_archive.py
maximemulder: Changed to the new logging abstraction !

python/dicom_archive.py
maximemulder: Adopted in the main script ! Not in `summarize_dicom_study` since it is a very small script that does not use the database where using the "big" argument parser over the standard library is not worth it IMO.

python/lib/dicom/dicom_database.py
maximemulder: Adopted the new LORIS-MRI database abstraction !

python/dicom_archive.py
maximemulder: Well, named `dicom_archive` but it still kind of works when compared with the other variable names IMO.

python/dicom_archive.py
maximemulder: Removed the `-today` option.

python/dicom_archive.py
maximemulder: Removed the `-today` option. Also, if no scan date is found, there is now no date at all in the DICOM archive name.

python/dicom_archive.py
```diff
@@ -0,0 +1,325 @@
...
+print('Removing temporary files')
+
+os.remove(tar_path)
+os.remove(zip_path)
+os.remove(summary_path)
+os.remove(log_path)
+
+print('Calculating DICOM tar MD5 sum')
+
+log.archive_md5_sum = lib.dicom.text.make_hash(log.target_path, True)
```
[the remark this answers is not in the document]
maximemulder: Same points as before.

python/lib/dicom/dicom_database.py
```diff
@@ -0,0 +1,231 @@
...
+
+from lib.database import Database
+from lib.dicom.summary_type import Summary
+from lib.dicom.dicom_log import Log
+import lib.dicom.text
+import lib.dicom.summary_write
+import lib.dicom.dicom_log
+
+
+def select_dict(db: Database, fields: list[str], table: str, conds: dict[str, Any]):
```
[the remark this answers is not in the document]
maximemulder: Adopted new database abstraction.

python/lib/dicom/dicom_log.py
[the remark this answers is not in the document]
maximemulder: Renamed (that class and a bunch of others) to more descriptive names.

python/lib/dicom/summary_make.py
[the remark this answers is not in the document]
maximemulder: We updated our Python version since this comment.

python/lib/dicom/summary_make.py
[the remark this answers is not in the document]
maximemulder: Added a comment on the return value.

python/lib/dicom/dicom_log.py
maximemulder: Acutally renamed to `DicomStudyImportLog`.

python/lib/dicom/summary_make.py
maximemulder: Acutally, this function has been improved and moves to `lib.util`, which is a typed version of `lib.utilities`.

maximemulder:
@ridz1208 @cmadjar

I finally took some time to update and test this PR (I also edited the PR description). It has been rebased, adapted to the new ✨fancy✨ LORIS-MRI Python abstractions, and generally improved (notably with regards to control flow and variable naming).

I will use this PR to ingest the MPN DICOM scans into C-BIG, notably the `--session` feature. Note that this feature asks for the session ID directly instead of using the Python configuration. I don't think it should be the responsibility of the DICOM study importer to determine the session ID from the configuration. However, this script could be wrapped up in another script that does that and provides the session ID to the DICOM study importer.

python/scripts/import_dicom_study.py
```diff
@@ -0,0 +1,314 @@
...
+            env,
+            "Argument '--source' must be a readable directory path.",
+            lib.exitcode.INVALID_ARG,
+        )
+
+    if args.insert and args.update:
+        log_error_exit(
+            env,
+            "Arguments '--insert' and '--update' cannot be used both at the same time.",
+            lib.exitcode.INVALID_ARG,
```
cmadjar, line 131: Why make the distinction between `--insert` and `--update` here? I think a simple --database option is enough.

If --overwrite is used, then, it would update, if not, it would insert after having check that there is not already an entry in the DB for the StudyUID.

cmadjar:
@maximemulder very nice work overall!!

Two summary points:
- issue with a Philips odd B1 images (discussion on slack with details about the issue found and suggestion of improvements)
- issue with enhanced DICOMs (I want to take a closer look at this one on Monday to see if it would be possible to read enhanced DICOMs parameters). That last one is not a blocker to merge the PR though and can be implemented later on.

python/scripts/import_dicom_study.py
cmadjar: ok. Interesting. I have to think about this one. I would have a tendency to have a simple database flag and then figure out if data should be updated or inserted in the database when want to update the database instead of having to do that outside of the script in order to properly call import_dicom_study.py... I will sleep on it and get back to you.

cmadjar:
> @cmadjar Hhhhm, none of the scanner fields in the `tarchive` table are nullable, so I guess I will store `None` in the code, and insert an empty string in the database for now (we can still change the schema and insert `NULL` later). Should I do that for all the scanner fields (manufacturer, model, serial number, software version) or only some of them ?

Good catch! Looks like the current dicomTar.pl is indeed inserting with an empty string so probably best to use the same strategy here. And I think it is a good idea to apply that on all the Scanner fields, in case other anonymization software completely removes those.

Thanks!

cmadjar (approved):
Great job!!!

PR #1123, opened 2024-05-06
[minc_insertion] Improve command line arguments validation

nicolasbrossard:
This PR improves the validation done on the command line parameters for script `minc_insertion.pl`. More specifically:
- When the argument to `-upload_id` refers to a non-existent upload, the script displays a message indicating that the upload ID is invalid.
- When the upload ID passed on the command line refers to an upload that does not have an associated `tarchive` (i.e value of `TarchiveID` for that upload is `NULL` in table `mri_upload`), an error message to that effect is displayed to the user.


MaximeMulder (approved):
Worked on my end.

PR #1138, opened 2024-06-20
integration test

kongtiaowang:
This PR will run the sql files in RB_SQL folder without running other integration tests from core.


.github/workflows/flake8_python_linter.yml
```diff
@@ -10,15 +10,19 @@ jobs:
   
   flake8-lint:
     runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
```
cmadjar, line 16:
```suggestion
    strategy:
      matrix:
        python-version: ["3.11", "3.12"]
```

cmadjar: @kongtiaowang any chance you could update the above? Thank you :)

.github/workflows/loristest.yml
```diff
@@ -0,0 +1,92 @@
...
+      run: |
+        cd Loris
+        make dev
+
+    - name: Run Test Suite
+      run: |
+          # set sandbox to 1 before running the tests
+          # since some tests only run in sandbox environments
+          sed -i 's/<sandbox>0<\/sandbox>/<sandbox>1<\/sandbox>/g' Loris/test/config.xml
+          cd Loris
```
laemtl, line 86: Why do we run the LORIS tests?

cmadjar: @kongtiaowang? See Laetitia's question above. I don't understand either why we would need to run the LORIS tests here. Thank you :)

test/RB_SQL/RB_Config.sql
```diff
@@ -0,0 +1,116 @@
...
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (23,27,'/data-raisinbread/');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (24,28,'/var/www/loris/');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (26,30,'/PATH/TO/EXTERNAL/LIBRARY/');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (28,32,'/data/');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (29,33,'tools/logs/');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (30,34,'/data/incoming/');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (31,35,'/data/demo/bin/mri/');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (32,36,'/data/incoming/');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (33,37,'/PATH/TO/Genomic-Data/');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (34,38,'/data/uploads/');
```
cmadjar, line 32: ConfigID 35 (MRICodePath), 27 (imagingPath) are essentials and not set. Would need to set them here.

test/RB_SQL/RB_Config.sql
```diff
@@ -0,0 +1,116 @@
...
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (12,13,'false');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (13,14,'false');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (15,16,'false');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (16,20,'false');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (17,21,'true');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (18,22,'true');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (19,23,'0');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (20,24,'Customize this text with your citation policy (via Configuration module)');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (22,12,'Ymd');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (23,27,'/data-raisinbread/');
```
cmadjar, line 23: Config 27 is there @laemtl

test/RB_SQL/RB_Config.sql
cmadjar: in the end, found the configs.

.github/workflows/loristest.yml
```diff
@@ -0,0 +1,92 @@
...
+
+    - name: Clone Loris repo
+      run: git clone https://github.com/aces/Loris.git
+
+    - name: Override Test Files
+      run: |
+        cp test/Dockerfile.test.php8 Loris/
+        cp test/docker-compose.yml Loris/
+        cp test/phpunit.xml Loris/test/
+        cp test/RB_SQL/* Loris/raisinbread/RB_files/
```
cmadjar, line 38: Could we select here the RB SQL tables we want to source? Technically, we only need the following tables:

- Config
- ConfigSettings
- ImagingFileTypes
- Project
- Visit_Windows
- all the `bids-*` tables
- candidate
- cohort
- mri_protocol_* tables
- mri_scan_type
- parameter_type (however, this could be populated automatically when running the pipeline)
- permissions* (I would guess this would be needed when using the API)
- project_cohort_rel
- psc
- session
- user_perm_rel
- user_project_rel
- user_psc_rel
- users
- visit
- visit_project_cohort_rel


Purposely skipping the following tables:
- feedback_mri_* (don't need feedbacks for LORIS-MRI)
- files (would use a very lightweight version of this table)
- files_intermediary
- files_qcstatus (QC so not really used by LORIS-MRI)
- electrophysiology_uploader (not sure if this is needed)
- hed_schema (not sure if this has data or has set up information, would need @jeffersoncasimir to decide)
- hed_schema_nodes (same, @jeffersoncasimir)
- hrrt_* (empty tables in RB)
- mri_parameter_form ( I don't think there is any use of that table for LORIS-MRI)
- mri_processing_protocol (empty table in RB)
- mri_scanner (has scanner data but the pipeline is supposed to create that if it does not exist so maybe we can ensure the logic of creating the scanner works if we skip this table?)
- mri_upload* (could have a lightweight version of it or keep it empty)
- mri_violations_log (could have a lightweight version of it or keep it empty)
- notification_spool (we don't really need all of that data)
- parameter_files (could have a lightweight version of it or keep it empty)
- physiological* tables (could have a lightweight version of it or keep it empty and the script will populate those)
- server_processes (could probably keep that empty)
- tarchive* (would have a lightweight version of it with a couple of files)
- violations_resolved (could have a lightweight version of it or keep it empty)

.github/workflows/loristest.yml
kongtiaowang: yes, we need to install Loris first, it will include the all tables from raisinbread data. like config ... , we don't have python testsuit yet on gitactions, right?

.github/workflows/loristest.yml
kongtiaowang: yes, you can do it override any files under the RB_SQL folder.

.github/workflows/loristest.yml
cmadjar: all good. Thanks!

PR #1144, opened 2024-07-30
Validate subject IDs early in the imaging pipeline

MaximeMulder:
This PR validates the subject IDs at the very beginning of the imaging pipeline, preventing the insertion of any MRI upload / DICOM archive that would not be associated to a session because of a typo or configuration mistake.

This PR uses the Python configuration function `get_subject_ids`, even in the Perl code (it calls the Python code from the Perl code). It also includes a factorization and de-duplication of some existing Python code regarding the obtentation and validation of subject IDs.

# Caveats for projects

- need to have python dependencies installed if not already installed + datatabase_config.py configured.


batch_uploads_imageuploader.pl
```diff
@@ -63,8 +63,8 @@ =head2 Methods
 
 
 my $profile   = '';
-my $upload_id = undef; 
-my ($debug, $verbose) = (0,1);
+my $upload_id = undef;
+my ($debug, $verbose) = (0,0);
```
cmadjar, line 67: why has this changed?

batch_uploads_imageuploader.pl
```diff
@@ -229,13 +229,34 @@ =head2 Methods
...
+            $patientname,
+        );
+
+        if ($verbose) {
+            $command .= " --verbose";
+        }
+
+        if (system($command) != 0) {
+            # The error is already printed by the Python script, just exit
+            exit $NeuroDB::ExitCodes::CANDIDATE_MISMATCH;
```
cmadjar, line 255: This would fail silently. I think the data should still be inserted in the imaging uploader but with a failure status instead.

python/lib/database_lib/candidate_db.py
```diff
@@ -35,23 +35,16 @@ def __init__(self, db, verbose):
...
+        the database.
         """
 
-        query = "SELECT c1.CandID, c2.PSCID AS PSCID " \
-                " FROM candidate c1 " \
-                " LEFT JOIN candidate c2 ON (c1.CandID=c2.CandID AND c2.PSCID = %s) " \
-                " WHERE c1.CandID = %s"
+        query = 'SELECT PSCID ' \
+            'FROM candidate ' \
+            'WHERE CandID = %s'
```
cmadjar, line 46: you should revert back to the old SQL. There is a reason why there are those joins. It allows checking the the PSCID and DCCID refers to the same candidate.

python/lib/database_lib/visit_windows.py
```diff
@@ -44,6 +44,5 @@ def check_visit_label_exits(self, visit_label):
         """
 
         query = 'SELECT Visit_label FROM Visit_Windows WHERE BINARY Visit_label = %s'
-        results = self.db.pselect(query=query, args=(visit_label,))
-
-        return results if results else None
+        results = self.db.pselect(query=query, args=[visit_label])
```
cmadjar, line 47: I think that would break since pselect args expects a tuple.

python/lib/database_lib/visit_windows.py
```diff
@@ -44,6 +44,5 @@ def check_visit_label_exits(self, visit_label):
...
+        return bool(results)
```
cmadjar, line 48: Have you checked that changing the return value type does not break any other code?

python/lib/dcm2bids_imaging_pipeline_lib/base_pipeline.py
```diff
@@ -105,10 +106,11 @@ def __init__(self, loris_getopt_obj, script_name):
...
         # ---------------------------------------------------------------------------------
         if self.dicom_archive_obj.tarchive_info_dict.keys():
-            self.subject_id_dict = self.imaging_obj.determine_subject_ids(self.dicom_archive_obj.tarchive_info_dict)
-            if 'error_message' in self.subject_id_dict:
+            try:
+                self.subject_id_dict = self.imaging_obj.determine_subject_ids(self.dicom_archive_obj.tarchive_info_dict)
+            except DetermineSubjectException as exception:
                 self.log_error_and_exit(
-                    self.subject_id_dict['error_message'],
+                    exception.message,
```
cmadjar, line 113: nice the exception refactoring! Have you checked that determine_subject_ids is not used anywhere else? To make sure nothing gets broken by this refactoring. Thanks!

cmadjar (changes requested):
a few comments.

The main change would be that we need to still insert the DICOMs in the imaging uploader and mark the insertion as failed to keep a trace on the frontend. I know that for CBIGR, you don't want to insert but most studies would still want to see the failure.

A tradeoff would be to create a config under the imaging pipeline section with default to always insert in imaging uploader that you can turn off for CBIGR.

batch_uploads_imageuploader.pl
MaximeMulder: Oh, maybe I should have proposed this change in a separate PR indeed.
The current line makes the imaging pipeline always verbose, no matter if the `--verbose` argument is provided or not. When calling the Python script from Perl, it becomes very noticeable since all database queries are logged.

python/lib/database_lib/candidate_db.py
MaximeMulder: Whoops, I forgot that the PSCID is not unique, my mistake !

python/lib/database_lib/visit_windows.py
MaximeMulder: The code worked fine (on my machine™) with the lists and I saw a few places that use lists (in `lib/candidate.py`) so I went with it. But it seems that even the MySQL docs uses tuples so you're probably right that I should go with it yes. (although I must say I do not get why one would design this API with tuples over lists here)

python/lib/database_lib/visit_windows.py
MaximeMulder: Yep ! The function only has one reference in the codebase (the place I changed).

python/lib/dcm2bids_imaging_pipeline_lib/base_pipeline.py
MaximeMulder: Yep ! I did change the error management in all the places this function is called. There is just one place which had no error management at all, so I did not any:
https://github.com/aces/Loris-MRI/blob/main/python/lib/dcm2bids_imaging_pipeline_lib/nifti_insertion_pipeline.py#L84

MaximeMulder:
> The main change would be that we need to still insert the DICOMs in the imaging uploader and mark the insertion as failed to keep a trace on the frontend. I know that for CBIGR, you don't want to insert but most studies would still want to see the failure.
> 
> A tradeoff would be to create a config under the imaging pipeline section with default to always insert in imaging uploader that you can turn off for CBIGR.

This sounds reasonable to me, and I wonder if it is not what C-BIG should be doing as well. @ridz1208 do you have an opinion ?

I understand we want to avoid `NULL` session IDs as much as possible, but would it really be a problem to have `NULL`-sessioned MRI uploads here ? IMO this seems like the only way to have failed insertion logs on the front-end, and that these entries should not be causing any harm since there is no data linked to it.

python/lib/imaging.py
```diff
@@ -510,93 +512,61 @@ def grep_cand_id_from_file_id(self, file_id):
...
-            subject_id_dict['PatientName'] = dicom_value
-        except AttributeError:
-            message = 'Config file does not contain a get_subject_ids routine. Upload will exit now.'
-            return {'error_message': message}
-
+        subject_id_dict = self.determine_subject_ids_from_name(dicom_value, scanner_id)
         return subject_id_dict
 
-    def validate_subject_ids(self, subject_id_dict):
+    def determine_subject_ids_from_name(self, subject_name: str, scanner_id: Optional[int] = None) -> dict[str, Any]:
```
cmadjar, line 538: duplicated function name. See line 515

MaximeMulder: Oh, there is no duplicate name, on line 515 there is `determine_subject_ids`, and on 538 there is `determine_subject_ids_from_name`, the former calls the latter. The diff is a bit messy here.

ridz1208:
> A tradeoff would be to create a config under the imaging pipeline section with default to always insert in imaging uploader that you can turn off for CBIGR.
@MaximeMulder @cmadjar This point was brought up by @nicolasbrossard the first time we discussed this, he was asking how would we handle a failed upload if its not inserted anywhere and a few suggestions came up I believe but now that its more concrete we can maybe rediscuss on the imaging meeting?

A few suggestions that came up (but we had not setteled on any) were:
1. keep a local log where the scans are being uploaded from to store unsuccessful scans
2. use crontab's email system to email specifically designed coordinators (by project) when a scan fails to upload
3. use the new DICOM API to have proper feedback from an upload failure

new: 
4. have a config option for accepting or not NULL session IDs

Briefly my thoughts are that even for CBIG if NULL sessionID is not accepted we need a way to notify the scan "owners" that it failed to upload so solutions 1,2 and/or 3 will need to be implemented at least for CBIG even if 4 is implemented. As I explained previously my issue with NULL sessionIDs is technical not procedural. The problem is that a NULL sessionID means that we were not able to match the "patient_name" to a session for a candidate because of whatever reason... soooo on the front end who sees it?

- Option 1: noone sees it, its filtered out - thats pretty useless we may as well not insert it
- Option 2: everyone sees it - problematic in an institutional instance given that patient name may contain if we dont want it to, after all it failed to match with a candidate/session otherwise it would have not had SessionID=NULL
- Option 3: a subset of users with a "special" permission can see only those scans... - sure but then we still need to know the source of the scan to know who to send it to to fix the name and reupload... we are back to the file system of origin

Unless I'm missing something here thats my view of the issue. We can certainly discuss at the meeting.

MaximeMulder:
Closed in favor of using a new script that calls the DICOM upload API to insert scans.

PR #1145, opened 2024-07-31
Update Python installation

MaximeMulder:
As discussed in https://github.com/aces/Loris-MRI/pull/1117#discussion_r1697285437, I believe we should update our Python installation. This PR does that, with a few other improvements to our Python environment.

The Python version used in this PR is Python 3.11, which is a relatively recent stable Python version and is likely the best target imo:
- 3.9 and 3.12 require to add APT repositories
- 3.10 also works but is less recent

More precisely, I did not reinstall a full LORIS-MRI instance during my testing, but I did test the PR changes on my Ubuntu 22 dev VM and Debian 12, by doing the following :
- Install the specific Python version and its dependencies.
- Create our virtual environment.
- Install our package requirements in the environment.

Another change made in this PR is that it installs a fixed Python version, which is not the case in our current setup (for instance, I had Python 3.10 installed on my dev VM although we use 3.8 in our testing).

I advise to ignore whitespace for reviewing (trailing spaces removed with autoformatting).

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ This Readme covers release 26.0.* of the LORIS Imaging Insertion Pipeline for Ub
 This repo accompanies the [LORIS neuroimaging data platform main repo](https://github.com/aces/Loris/releases)</b>, release 24.1.*.<br>
 For documentation and detailed setup information, please see the [LORIS-MRI documentation](docs/) for your installed version</b>.
 
-This repo can be installed on the same VM as the main LORIS codebase, or on a different machine such as a designated fileserver where large imaging filesets are to be stored. 
+This repo can be installed on the same VM as the main LORIS codebase, or on a different machine such as a designated fileserver where large imaging filesets are to be stored.
 
 # System Requirements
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -54,21 +54,22 @@ sudo chown -R lorisadmin:lorisadmin /opt/$projectname
...
+[releases page](https://github.com/aces/Loris-MRI/releases)
 and extract it to `/opt/$projectname/bin/mri`
 
 #### 2. Install Python 3 with `pip` and `virtualenv`
 
 ```bash
-sudo apt install python3 
-sudo apt install python3-dev
+sudo apt install software-properties-common
+sudo add-apt-repository ppa:deadsnakes/ppa
```
MaximeMulder, line 65: This repository allows us to not rely on the default distro repositories, which may be missing some packages we need (on Ubuntu 22 I only have 3.10 and 3.11 for instance).

This repository seems to be the standard for Python setups, example: 
https://askubuntu.com/questions/1398568/installing-python-who-is-deadsnakes-and-why-should-i-trust-them

README.md
```diff
@@ -54,21 +54,22 @@ sudo chown -R lorisadmin:lorisadmin /opt/$projectname
...
+sudo apt update
+sudo apt install python3.12 python3.12-dev python3.12-distutils python3.12-venv
```
MaximeMulder, line 67: Some Python versions had problems with `pip` in the virtual environment without `python3.x-distutils`.

install/imaging_install.sh
```diff
@@ -115,7 +115,7 @@ echo
 echo "Creating loris-mri Python virtualenv in $mridir/python_virtualenvs/loris-mri-python/"
 # create a directory in $mridir that will store python 3 virtualenv
 sudo -S su $USER -c "mkdir -m 770 -p $mridir/python_virtualenvs/loris-mri-python"
-virtualenv $mridir/python_virtualenvs/loris-mri-python -p `which python3`
+python3.12 -m venv $mridir/python_virtualenvs/loris-mri-python
```
MaximeMulder, line 118: This method for creating virtual environments is the recommended one:
https://docs.python.org/3/library/venv.html#creating-virtual-environments

The current command did not work with Python 3.12.

MaximeMulder:
Ok I thought the PR was done, but then I made changes, but now I think it is done for real.

cmadjar (approved):
LGTM. Thank you @MaximeMulder

PR #1146, opened 2024-08-08
Improve Python subject determination and validation error handling

MaximeMulder:
Changes extracted from #1144.

Here are the changes:
- Remove function `validate_subject_ids` in `lib/imaging.py` as it was dead code.
- Add a file `lib/validate_subject_ids` to isolate subject IDs validation.
- Add a directory `python/lib/exception` to contain exception classes.
- Add two exceptions `DetermineSubjectException` and `ValidateSubjectException` to handle subject IDs errors, and use these exceptions where dictionary entries were previously used.
- Simplify the PSCID / CandID combination checking by using a simpler SQL query.

### DIFF
--- a/python/lib/database_lib/candidate_db.py
+++ b/python/lib/database_lib/candidate_db.py
@@ -35,23 +35,16 @@ class CandidateDB:
         self.db = db
         self.verbose = verbose
 
-    def check_candid_pscid_combination(self, psc_id, cand_id):
+    def get_candidate_psc_id(self, cand_id: str | int) -> str | None:
         """
-        Checks whether the PSCID/CandID combination corresponds to a valid candidate in the `candidate` table.
-
-        :param psc_id: PSCID of the candidate
-         :type psc_id: str
-        :param cand_id: CandID of the candidate
-         :type cand_id: int
-
-        :returns: the valid CandID and PSCID if the combination corresponds to a candidate, None otherwise
+        Return a candidate PSCID and based on its CandID, or `None` if no candidate is found in
+        the database.
         """
 
-        query = "SELECT c1.CandID, c2.PSCID AS PSCID " \
-                " FROM candidate c1 " \
-                " LEFT JOIN candidate c2 ON (c1.CandID=c2.CandID AND c2.PSCID = %s) " \
-                " WHERE c1.CandID = %s"
+        query = 'SELECT PSCID ' \
+            'FROM candidate ' \
+            'WHERE CandID = %s'
 
-        results = self.db.pselect(query=query, args=(psc_id, cand_id))
+        results = self.db.pselect(query, args=[cand_id])
 
-        return results if results else None
+        return results[0]['PSCID'] if results else None
--- a/python/lib/database_lib/candidate_db.py
+++ b/python/lib/database_lib/candidate_db.py
@@ -41,9 +41,7 @@ class CandidateDB:
         the database.
         """
 
-        query = 'SELECT PSCID ' \
-            'FROM candidate ' \
-            'WHERE CandID = %s'
+        query = 'SELECT PSCID FROM candidate WHERE CandID = %s'
 
         results = self.db.pselect(query, args=(cand_id,))
 
--- a/python/lib/database_lib/candidate_db.py
+++ b/python/lib/database_lib/candidate_db.py
@@ -45,6 +45,6 @@ class CandidateDB:
             'FROM candidate ' \
             'WHERE CandID = %s'
 
-        results = self.db.pselect(query, args=[cand_id])
+        results = self.db.pselect(query, args=(cand_id,))
 
         return results[0]['PSCID'] if results else None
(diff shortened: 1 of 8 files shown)


python/lib/database_lib/candidate_db.py
```diff
@@ -35,23 +35,16 @@ def __init__(self, db, verbose):
...
+        the database.
         """
 
-        query = "SELECT c1.CandID, c2.PSCID AS PSCID " \
-                " FROM candidate c1 " \
-                " LEFT JOIN candidate c2 ON (c1.CandID=c2.CandID AND c2.PSCID = %s) " \
-                " WHERE c1.CandID = %s"
+        query = 'SELECT PSCID ' \
+            'FROM candidate ' \
+            'WHERE CandID = %s'
```
MaximeMulder, line 46: This was a comment in the old PR, but I still don't see how a join is necessary. If I just query the candidate by CandID, and that the of the candidate PSCID matches the PSCID of the file name, it is sufficient to prove that everying is fine to me.

python/lib/database_lib/candidate_db.py
```diff
@@ -35,23 +35,16 @@ def __init__(self, db, verbose):
...
 
-        results = self.db.pselect(query=query, args=(psc_id, cand_id))
+        results = self.db.pselect(query, args=(cand_id,))
 
-        return results if results else None
+        return results[0]['PSCID'] if results else None
```
MaximeMulder, line 50: This was a comment in the old PR, but I still don't see how a join is necessary here. If I query the candidate by CandID, and that the PSCID of the candidate matches the PSCID of the file name, it is a sufficient proof that everying is fine to me.

cmadjar: fair enough. That select statement is derived from the perl code actually. @nicolasbrossard any reason why that query is used instead of just selecting the PSCID based on a CandID (and later on checking that the PSCID in the filename is the same as the PSCID in the database). I have a feeling there was a reason for it but my memory fails me. Hopefully, yours is in a better shape ;)

python/lib/imaging.py
```diff
@@ -510,93 +512,44 @@ def grep_cand_id_from_file_id(self, file_id):
...
 
         config_obj = Config(self.db, self.verbose)
         dicom_header = config_obj.get_config('lookupCenterNameUsing')
-        dicom_value = tarchive_info_dict[dicom_header]
+        subject_name = tarchive_info_dict[dicom_header]
 
         try:
-            subject_id_dict = self.config_file.get_subject_ids(self.db, dicom_value, scanner_id)
-            subject_id_dict['PatientName'] = dicom_value
+            subject_id_dict = self.config_file.get_subject_ids(self.db, subject_name, scanner_id)
```
cmadjar, line 536: I am wondering if we should check also that the `subject_id_dict` from the config file is returning the proper dictionary structure. AKA a key for `PSCID`, a key for `CandID` a key for `visitLabel`, a key of `isPhantom` and a key for `createVisitLabel`. Since the scripts relies on that structure, it could be nice to double that the config file follows that structure.

python/lib/database_lib/candidate_db.py
cmadjar, line 46: That could probably be just in one line given how short the query is.

cmadjar (commented):
@MaximeMulder see my comments attached. One of them is tagging Nicolas because I want to be sure I did not forget something important before we go with the simpler query to get PSCID.

python/lib/imaging.py
MaximeMulder: I AGREE TO THE FULLEST EXTENT. This is actually the next PR that I wanted to discuss. A dictionary is not the right data structure here, because we know exactly what fields we want to be filled. I was thinking of using specific dataclasses that contain exactly the fields we need. This means that current projects need to rewrite their Python configuration but it is a very straightforward change to make and makes this configuration easier and clearer IMO.

MaximeMulder:
### Summary of today's design meeting

This PR design is fine. The details on the unanswered are the following:
- Put the `api.py` file in the `lib.dataclass` namespace to avoid conflict with the `lib.api` namespace.
- Keep the types in the route files for now.
- Have a default API version in the `Api` class.
- Be careful about nullable types in the structured API.
- Do not use a library to structure replies.

I looked into Swagger codegen in more details. It certainly generates a lot of the boilerplate to us, which is more practical to create the API client, but the generated code seems to be of lesser quality, and notably does not seem to have type hints (big no-no for me).

A final not really related to this PR, but another one that should arrive next week, is how to pass user credentials to a script that calls the API. One solution is simply to store a single user's credentials in the config file, as @cmadjar does it. However, I need a script which can be ran by multiple users, and the options we considered for the password are these ones:
- Use a CLI argument (likely unsecure)
- Use an environment variable (liked by @laemtl and others)
- Use the standard input (more cumbersome to use)
- Use a specific config file located at a hardcoded path somewhere in the Linux user home directory.
- Use a specific config file passed as a CLI argument.
I will ask Sylvain about this as he might have some insights or an opinion.

cmadjar (approved):
LGTM and tested locally to make sure all worked.

PR #1151, opened 2024-08-20
(preliminary to SQLAlchemy) Rename `database.py`

MaximeMulder:
I'd like to use the `lib.database` namespace for the upcoming SQLAlchemy PR proposal. Thus, I propose moving `lib/database.py` to `lib/database_lib/__init__.py` (but any other name is fine with me) to free the namespace. This is done in a separate PR for better separation and ease of review.

IMO, `database_lib` should be depreciated and eventually replaced by SQLAlchemy in the medium to long term. Given the size of the codebase, I do not expect this migration to be long, and can even do it myself if wanted !


MaximeMulder:
I don't know why there are whitespace changes although I did rebase on `upstream/main`, but it do be like that I guess :thinking:.

MaximeMulder:
Closed since SQLAlchemy will use the `lib.db` namespace instead, for more backwards compatibility to ease the transition of exisiting projects.

PR #1167, opened 2024-08-28
Small linting improvements

MaximeMulder:
While looking at linting for the  LORIS-MRI automated testing initiative, I noticed that flake8 currently ignores rule [E114](https://www.flake8rules.com/rules/E114.html), which states that tabulations must be 4 spaces. We obviously want this rule, so I un-ignored it and fixed the code that was not compliant.

There are also a few whitespace changes I missed in my last linting PR (didn't notice, my fault !). It should be good now.


cmadjar (changes requested):
@MaximeMulder some of the changes in this PR are still tabulated instead of 4 spaces.

MaximeMulder:
@cmadjar There was a tabulation in a comment that I just removed, was it the one you were talking about ? I don't see any other one in our code.

image

cmadjar (approved):
LGTM. In the end, I don't know where I saw the tabs. Hallucinations!!! 🤣 

Thanks!

PR #1172, opened 2024-08-30
run pytest in gitaction

kongtiaowang:


test/Dockerfile.test.py
```diff
@@ -0,0 +1,22 @@
+# Dockerfile
+FROM python:3.11.3-slim
+
```
cmadjar, line 3: @kongtiaowang you get those lint errors because flake8 things that this file is a python file. To fix that, you should probably rename it to `Dockerfile.test.python`. That should fix the lint issue.

test/Dockerfile.test.py
```diff
@@ -0,0 +1,22 @@
+# Dockerfile
+FROM python:3.11.3-slim
```
laemtl, line 2: Maybe we can add this extra line to remove this error. Otherwise looks good.

laemtl (approved):

cmadjar (approved):

PR #1175, opened 2024-08-30
Remove php 8.1 and 8.2 from the LORIS tests to keep only test on PHP 8.3

cmadjar:
# Description

We should probably just be testing the latest PHP version tested by LORIS in the LORIS-MRI testing. This PR removes call to PHP 8.1 and 8.2 in the testing and keeps only tests for PHP 8.3.

### DIFF
--- a/.github/workflows/loristest.yml
+++ b/.github/workflows/loristest.yml
@@ -11,7 +11,7 @@ jobs:
         fail-fast: false
         matrix:
             testsuite: ['integration']
-            php: ['8.1','8.2', '8.3']
+            php: ['8.3']
             ci_node_index: [0]
 
             include:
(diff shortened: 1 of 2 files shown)


kongtiaowang (approved):
LGTM

MaximeMulder (changes requested):
If we use a single version of PHP, shouldn't we just get rid of `strategy.matrix` altogether and inline the variables ? We may want to declare variables for clarity but `strategy.matrix` is not the place for that here IMO.

- https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow
- https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables

Feel free to either change and merge, merge this and do the simplification in another PR, or close and do everything in another PR.

laemtl:
@MaximeMulder What about if we want to add another PHP version later? Would it be easier to keep it?

PR #1182, opened 2024-09-09
LORIS-MRI tooling configuration (Docker version)

maximemulder:
Alternative to #1170. The differences are the following :
- In #1170, each static test is ran in a GitHub action runner, which builds the Python environment with the PIP dependencies being cached.
- In this PR, a first GitHub action will build the Python environment in a Docker image, which is cached in the GitHub Container Registry. Then, each static test is ran in this image on their GitHub action runner.

The workflow in this PR is arguably more complex than #1170, but it provides a better isolation and reproducibility for the testing environment. Although this PR caches the whole Docker image instead of just the dependencies, the performance looks similar, as the time gained in the environment setup is lost in launching and stopping the container.


maximemulder:
Okay, ignoring the messy Git history (kinda too lazy to fix it now) this PR is done. Although this approach has some advantages, it is also slower than the current Python tests setup, so I am not sure if we should merge it. Anyway, I learned quite a bit about Docker and GitHub Actions which will be useful for integration tests. This PR in itself will be a subject of discussion for another meeting.

maximemulder:
We concluded in the meeting that the dockerization of our Python checks is not worth the complexity and speed hit (~2min) of this PR. Closing.

PR #1192, opened 2024-10-01
bunch of mysql definitions

maximemulder:
This is bunch of SQLAlchemy definitions I extracted from my various PRs in order to make them easier to review.

These definitions do not interact with the current code without the related PRs. Nevertheless, the (few) unit tests work, and the manual tests I did on each PR worked as well. I don't think there is any testing to do on this PR as it will be done when we review the related PRs instead, as well as with the integration testing when possible.


python/lib/db/model/session.py
```diff
@@ -35,10 +39,13 @@ class DbSession(Base):
...
+    scan_done                : Mapped[Optional[bool]]     = mapped_column('Scan_done', YNBool)
     mri_qc_status            : Mapped[str]                = mapped_column('MRIQCStatus')
-    mri_qc_pending           : Mapped[str]                = mapped_column('MRIQCPending')
+    mri_qc_pending           : Mapped[bool]               = mapped_column('MRIQCPending', YNBool)
     mri_qc_first_change_time : Mapped[Optional[datetime]] = mapped_column('MRIQCFirstChange')
     mri_qc_last_change_time  : Mapped[Optional[datetime]] = mapped_column('MRIQCLastChange')
     mri_caveat               : Mapped[str]                = mapped_column('MRICaveat')
     language_id              : Mapped[Optional[int]]      = mapped_column('languageID')
+
+    candidate : Mapped['db_candidate.DbCandidate'] = relationship('DbCandidate', back_populates='sessions')
```
cmadjar, line 50: I think it is missing the CohortID foreign key here.

python/lib/db/model/candidate.py
```diff
@@ -15,10 +19,10 @@ class DbCandidate(Base):
     dete_of_death           : Mapped[Optional[date]] = mapped_column('DoD')
     edc                     : Mapped[Optional[date]] = mapped_column('EDC')
     sex                     : Mapped[Optional[str]]  = mapped_column('Sex')
-    registration_site_id    : Mapped[int]            = mapped_column('RegistrationCenterID')
+    registration_site_id    : Mapped[int]            = mapped_column('RegistrationCenterID', ForeignKey('psc.CenterID'))
```
cmadjar, line 22: missing FK for ProjectID and Project table definition

python/lib/db/query/config.py
```diff
@@ -0,0 +1,16 @@
...
+def get_config_with_setting_name(db: Database, name: str):
+    """
+    Get a single configuration entry from the database using its configuration setting name, or
+    raise an exception if no entry or several entries are found.
+    """
+
+    return db.execute(select(DbConfig)
+        .join(DbConfig.setting)
+        .where(DbConfigSetting.name == name)
+    ).scalar_one()
```
cmadjar, line 16: I think it is possible that a config is not set in which case this would return None.

cmadjar: I guess that if the config is not existent it will raise an exception but do we catch those? What is the error message when that happens?

python/lib/db/query/notification.py
```diff
@@ -0,0 +1,14 @@
+from sqlalchemy import select
+from sqlalchemy.orm import Session as Database
+from lib.db.model.notification_type import DbNotificationType
+
+
+def get_notification_type_with_name(db: Database, name: str):
+    """
+    Get a notification type from the database using its configuration setting name, or raise an
+    exception if no notification type is found.
```
cmadjar, line 9: do we want to raise an exception if notification type is not existant or do we want to create the notification type?

python/lib/db/query/site.py
```diff
@@ -0,0 +1,20 @@
...
+from sqlalchemy.orm import Session as Database
+
+from lib.db.model.candidate import DbCandidate
+from lib.db.model.session import DbSession
+from lib.db.model.site import DbSite
+
+
+def try_get_site_with_psc_id_visit_label(db: Database, psc_id: str, visit_label: str):
+    """
+    Get a session from the database using its candidate CandID and visit label, or return `None`
```
cmadjar, line 11: the doc says it uses the candID but the query uses the PSCID.

python/lib/db/query/notification.py
maximemulder: The current code does not create the notification type, but I thought about it and I agree creating it would probably be the better option.

python/lib/db/query/site.py
maximemulder: Whoops, that is indeed an oversight on my part, I'll be fixing that !

python/lib/db/query/config.py
maximemulder: For now it raises an exception, but in the future I was planning to have a `python/lib/config.py` module to make the queries and handle the exceptions by exiting the program.

cmadjar: I think some Config could be not set (like the S3 configs) so probably best to support those cases.

python/lib/db/query/notification.py
cmadjar: yes, that is what I assumed. I think for now it is fine but we could probably add that in the future. The same way we create the parameter_type when it does not exist. I don't think this is high priority though.

cmadjar:
@maximemulder I thought about the Project table because it can be useful when handling sessions potentially. I realize that the other imaging tables are not there yet but it can be done in a separate PR.

cmadjar (approved):
👍

PR #1194, opened 2024-10-03
Stricter global type checking

maximemulder:
Improve the global type checker by adding rules for a few things that can obviously be checked as errors even without the presence of types (unbound variables, useless type checks, unused expressions...).


cmadjar (approved):
LGTM

PR #1195, opened 2024-10-03
Optimize integration Docker image

maximemulder:
Merge some easily mergeable steps and remove some unused dependencies in the dockerfile to reduce the number of Docker layers, which should free some space in our GitHub actions cache.

The "variables should be present before their use" warning was already present before. We'll look into removing it once we have actual integration tests running (still waiting on the S3 bucket for that).


kongtiaowang (approved):
LGTM

PR #1196, opened 2024-10-04
Move Python scripts to `python/scripts` directory

maximemulder:
Title. This PR isolates the Python executable scripts in their own directory.

I tested it by starting a new SSH session with these changes enabled and running a script `run_dicom_archive_validation.py`.

I do not know how other projects are organized, but that might be a breaking change if they add scripts directly in the `python` directory, if not then it should be a transparent change.


cmadjar (approved):
👍

PR #1150, opened 2024-08-14
Improve subject configuration structure

maximemulder:
This PR makes the Python configuration file (database, S3, and subject IDs) more explicit by wrapping it in typed dataclasses instead of dicts, and adds a few comments.

I discussed some suspicious phantom-related code with @cmadjar, and it appears that phantom scans are not deeply tested in LORIS-MRI Python so there are likely a few bugs in our code. Therefore, I removed the suspicious lines as we probably want the scanner/scanner candidate to be created as early as possible so that the rest of the code does not have to worry about it.

Phantom configuration is still possible provided that the scanner and candidate exist in the database, which the user can add themselves since there is access to the database in the configuration function. More work is probably needed on that front (small quality-of-life improvements or the infamous phantom refactor) but that is a subject for other PRs.

**NOTE FOR EXISTING PROJECTS**: To update your project, change your configuration file (usually `dicom-archive/.loris_mri/database_config.py`) from the dict-shaped template to the typed one by imitating `dicom_archive/database_config_template.py`.

PR #1200, opened 2024-10-10
Remove mysql-connector dependency

maximemulder:
While testing #1150, @cmadjar encountered an issue when connecting to the database through SQLAlchemy. I suspect this is because this connection is using a deprecated version of [`mysql-connector`](https://pypi.org/project/mysql-connector/). This dependency was already present in LORIS-MRI before adding SQLAlchemy although it was unused, and I wrongfully used it for the SQLAlchemy connection not knowing the version we had was deprecated (the non-depecrated version is [`mysql-connector-python`](https://pypi.org/project/mysql-connector-python/).).

Solution: Remove the `mysql-connector` dependency and use `mysqlclient` instead, which is the client we use for the non-SQLAlchemy database connection.

P.S. 1: I manually recreated my virtual environment to make sure the dependency was completely removed, but it was detected by git, so I added the directory the gitignore, I have no idea why it was ignored before.

P.S. 2 : Some whitespace auto-trimming, ignore whitespaces in review.

# Notes for existing projects

- need to remove `mysql-connector` via `pip uninstall mysql-connector`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 **/__pycache__/*
 *.pyc
 .loris_mri/
+python_virtualenvs/
--- a/install/INSTALL.Mac.md
+++ b/install/INSTALL.Mac.md
@@ -1,18 +1,18 @@
 # LORIS-MRI Mac Install Guide
 
-### Mac systems are Not supported for LORIS. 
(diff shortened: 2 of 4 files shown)


cmadjar (approved):
Works with the `quote()` 👍

PR #1201, opened 2024-10-18
Remove unused attribute in NIFTI script

maximemulder:
The `bids_mapping_dict` attribute in `nifti_insertion_pipeline.py` is unused, and `get_bids_to_minc_terms_mapping` does not mutate its parameters. Hence the attribute and function call can be removed.

The script works by calling `get_bids_to_minc_terms_mapping` later in `imaging.py`.


maximemulder:
Sucessfully tested with `run_dicom_archive_loader`

PR #1203, opened 2024-10-18
Integration test for `run_dicom_archive_loader.py`

maximemulder:
This PR mounts the S3 bucket in CI and adds an integration test for `run_dicom_archive_loader.py` !


maximemulder:
Still kind of barebones but better than nothing for now IMO.

.github/workflows/integration.yml
```diff
@@ -11,6 +11,10 @@ env:
   DATABASE_NAME: TestDatabase
   DATABASE_USERNAME: TestUsername
   DATABASE_PASSWORD: TestPassword
+  BUCKET_URL: https://ace-minio-1.loris.ca:9000
+  BUCKET_NAME: loris-rb-data
+  BUCKET_ACCESS_KEY: lorisadmin-ro
+  BUCKET_SECRET_KEY: Tn=qP3LupmXnMuc
```
cmadjar, line 17: @maximemulder could this be secrets variables? I know this is for read-only data but still, not ideal to get credentials in the code.

cmadjar (dismissed):

.github/workflows/integration.yml
maximemulder: According to this doc:
https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-encrypted-secrets-in-a-workflow
> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.
So I guess our options are either:
- Keep the variables secret, but the integration tests must be triggered manually.
- Make the variables public.
Still looking into the above options, I may edit this message in the future.

maximemulder:
There is now a test to check that a file has indeed been inserted in the database !

maximemulder:
Wait, this is not Samir's brain, someone lied to me !!!

image

.github/workflows/integration.yml
maximemulder: After the setting up of a bucket without keys taking too long to set up with IT, it was decided to hard-code the keys for now.

laemtl (approved):

PR #1191, opened 2024-09-27
Unify LORIS-MRI logging

maximemulder:
# Description of the PR

This PR is a rewrite of the logging and I/O modules of LORIS-MRI.

# Problems of the current architecture

There are several problems with the current logging and I/O architecture of LORIS-MRI:
1. It is fragmented. The core logging class is the `Log` class. The `BasePipeline` and `ImagingIO` classes build on top of `Log` provide better abstractions, with some amount of duplication between those two classes (some scripts simply use `print` but they are not the subject of this PR).
2. The `BasePipeline` class is too coupled with the DCM2BIDS pipeline to be generalized.
3. The `BasePipeline` and `ImagingIO` classes handle both logging and file system operations, which are two different responsibilities.
4. The logging abstractions are verbose, notably because all logging functions require two keyword arguments  `is_error = 'Y' | 'N'` and `is_verbose = 'Y' | 'N'`.
5. The logging functions sometimes uses stdout to print errors, whereas they should use stderr.
6. The code does not integrate the latest LORIS-MRI architecture improvements, namely static typing and the SQLAlchemy integration.

# Description of the new architecture

This PR adds the following:
1. An `lib.env.Env` dataclass that stores generic information about a running script.
2. The `lib.logging` module that handles log operations.
3. The `lib.file_system` module that handles file system operations.
4. The `make_env` function that allows to build an `Env` from a `LorigGetOpt` object (note that the environment does not directly depend on `LorisGetOpt`, it could also be built manually).

Those abstractions aim to replace the `Log` and `ImagingIO` classes, as well as the `BasePipeline` logging methods. They are script-agnostic, remove the existing duplication, are statically typed, use the SQLAlchemy abstractions, and provide more convenient interfaces IMO. They also use simple functions instead of classes, which I think is better for simplicity, reusability and testability.

# Backwards compatibility

This PR is backwards compatible as the old abstractions are not replaced by this PR. They are however annotated as depcreated to let some time for external scripts to move to the old abstractions. All the uses of the old abstractions in LORIS-MRI are replaced by the new abstractions in this PR, so the deprecated files remain only for external scripts compatibility.

Ultimately (once the migration is finished), I would like the new logging module to take the `lib.log` namespace though (the same way I would like the new database module to use the `lib.database` module).

# Final notes

[...]

PR #1205, opened 2024-10-24
Fix Perl dependencies (& CI)

maximemulder:
I noticed the LORIS-MRI Docker image was no longer building in CI for #1191. After some investigation, it appears to be because of an update in [Module::Pluggable](https://metacpan.org/pod/Module::Pluggable) (the first in 9 years !), which some of our required packages depend on (notably [DateTime](https://metacpan.org/dist/DateTime), but also others I think). This PR fixes the issue by pinning the version of Module::Pluggable to version 5.2.

During the investigation of this PR, I also changed our pipeline to use `cpan` instead of `cpanm`, which seems to be [the recommended package installer by the Perl community](https://www.cpan.org/modules/INSTALL.html). It provides both a much clearer output (no longer literal tens of thousands of lines) and is faster by approximately 5 minutes.

### DIFF
--- a/install/INSTALL.Mac.md
+++ b/install/INSTALL.Mac.md
@@ -35,23 +35,27 @@ sudo ln -s /usr/local/mysql/lib/*.dylib .
 ```
 
 ```
-sudo -S cpan install Math::Round
-sudo -S cpan install Getopt::Tabular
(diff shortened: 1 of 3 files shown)


maximemulder:
Sometimes, you gotta do what you gotta do.

PR #1206, opened 2024-11-08
The return of the Python LORIS API client

maximemulder:
Some time ago, I made #1148 which drafts an architecture to call the LORIS API inside LORIS-MRI. However, I closed this PR as the scripts that effectively call the LORIS API will live on the BIC server and not inside of LORIS-MRI. Nevertheless, I think having a library to call the LORIS API in LORIS-MRI is a good thing. It can be used to test the API (which notably came in handy with the API DICOM upload PR), can be re-used by other projects in their scripts, and generally provide a code template to call the LORIS API.

There are a few improvements in this PR compared to the old PR. First, it is updated to fit within the new LORIS-MRI architecture improvements. Second, it uses two very popular Python libraries to improve the code robustness and ergonomy:
- `pydantic`: A library to define, parse and serialize models, which is useful to structure the API results.
- `requests`: A library to make HTTP requests, with better ergonomics than the Python standard library.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ line-length = 120
 preview = true
 
 [tool.ruff.lint]
-ignore = ["E202", "E203", "E221", "E241", "E251", "E272"]
+ignore = ["E202", "E203", "E221", "E241", "E251", "E271","E272"]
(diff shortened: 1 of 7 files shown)


MaximeBICMTL:
Superseded by #1307

PR #1207, opened 2024-11-11
Add file, file_parameter and scanner SQLAlchemy models

maximemulder:
In order to improve the integration tests for `run_dicom_archive_loader.py`. I want to query the `files` table to make sure the files are inserted in the database. This PR adds SQLAlchemy model definitions for this table (and `parameter_file` and `mri_scanner` because why not).

The integration tests pass therefore the models are correct !

### DIFF
--- a/python/lib/db/model/file.py
+++ b/python/lib/db/model/file.py
@@ -6,7 +6,7 @@ from sqlalchemy.orm import Mapped, mapped_column
 from lib.db.base import Base
 
 
-class DbMriFile(Base):
+class DbFile(Base):
     __tablename__ = 'files'
 
     id                             : Mapped[int]             = mapped_column('FileID', primary_key=True)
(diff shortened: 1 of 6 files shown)


maximemulder:
Just one quick question. I named the model for the `files` table `mri_file.py`. Is this name correct ? Or can there be files that are not MRI @cmadjar ?

cmadjar:
@maximemulder there can be files in the files table that are not MRI. I would change to `imaging_file.py` instead so that any imaging modality is covered (PET, MRI, MRS etc..)

maximemulder:
Ahok makes sense. Do you prefer just `file.py` or `imaging_file.py` ? The latter seems a little long to me but if you prefer it I'll use it instead.

maximemulder:
Okay, I changed the name ! By the way I used `file_parameter.py` instead of `parameter_file.py` because it makes more sense to group the parameters of a file it with the files than with the parameter types to me. But this is a weak opinion, if you think `parameter_file.py` is a better name that's fine by me.

cmadjar:
Probably best to stick with `parameter_file` like the table name to avoid confusions.

maximemulder:
Updated ! Once this is merged I will update the integration test PR to also ensure the files have been inserted in the database.

cmadjar (approved):
LGTM

PR #1190, opened 2024-09-27
Factorize get subject session

maximemulder:
This PR extracts the "get subject session" operation out of `BasePipeline`, with the goal of making it easily reusable by all scripts, including those that do not use `BasePipeline` (new DICOM archive and eventually BIDS import). It also modernizes the database access by using the new SQLAlchemy abstractions.

Is depended on by:
- #1117 (to implement #1044)

PR #1216, opened 2024-12-23
Migrate "get all sites from the database" to the SQLAlchemy abstraction

maximemulder:
Depends on #1190
Is depended on by #1211

Migrate "get all sites from the database" to the SQLAlchemy abstraction. I'd like to use the same function in the incremental BIDS import. This PR also deprecate a few functions that already have alternatives in the SQLAlchemy abstraction.


maximemulder:
Same reasong as #1190

PR #1224, opened 2025-01-08
Migrate `ImagingUpload` and `MriUploadDB` to the new database abstraction

maximemulder:
Migrate `ImagingUpload` and `MriUploadDB` to the new database abstraction.

The current input checking is very complex and this should simplify it a little. The previous code seems to sometimes accept or forget to check if the DICOM archive or MRI upload is null in a few specific places, but I do not see how it can work as a whole in these conditions. I therefore assume the DICOM archive and MRI upload must not be null.

The integration test will check if this works broadly, but since this touches the argument validation, I will need to test the pipeline manually with:

* :white_check_mark: Correct DICOM archive path and correct MRI upload ID.
* :white_check_mark: Correct DICOM archive path.
* :white_check_mark: Correct MRI upload ID.
* :white_check_mark: Incorrect DICOM archive path.
* :white_check_mark: Incorrect MRI upload ID.
* :white_check_mark: Correct DICOM archive path but incorrect MRI upload ID.
* :white_check_mark: Correct MRI upload ID but incorrect DICOM archive path.
* :white_check_mark: Correct DICOM archive path but no associated MRI upload.
* :white_check_mark: Correct MRI upload ID but no associated DICOM archive.
* :white_check_mark: No DICOM archive path or MRI upload ID.

Tested and there does not seem to be any regression: `run_dicom_archive_loader.py` expects either (exclusive) `--upload_id` or `--tarchive_path`, and requires the DICOM archive to be associated with exactly one MRI upload.

### DIFF
--- a/python/lib/dcm2bids_imaging_pipeline_lib/base_pipeline.py
+++ b/python/lib/dcm2bids_imaging_pipeline_lib/base_pipeline.py
@@ -185,7 +185,7 @@ class BasePipeline:
                 case []:
                     log_error_exit(
                         self.env,
-                        f"Did not find an entry in mri_upload associated with 'TarchiveID' {self.dicom_archive.id}",
+                        f"Did not find an entry in mri_upload associated with DICOM archive ID {self.dicom_archive.id}",
                         lib.exitcode.SELECT_FAILURE,
                     )
                 case [mri_upload]:
@@ -193,7 +193,10 @@ class BasePipeline:
                 case _:
                     log_error_exit(
                         self.env,
-                       f"Found {len(mri_uploads)} rows in mri_upload for 'TarchiveID' {self.dicom_archive.id}",
+                        (
+                            f"Found {len(mri_uploads)} rows in mri_upload for DICOM archive ID {self.dicom_archive.id}."
+                            f" MRI upload IDs are {', '.join(map(lambda mri_upload: str(mri_upload.id), mri_uploads))}."
+                        ),
                         lib.exitcode.SELECT_FAILURE,
                     )
 
--- a/python/lib/dcm2bids_imaging_pipeline_lib/base_pipeline.py
+++ b/python/lib/dcm2bids_imaging_pipeline_lib/base_pipeline.py
@@ -5,13 +5,13 @@ import lib.exitcode
 import lib.utilities
 from lib.database import Database
 from lib.database_lib.config import Config
-from lib.db.queries.dicom_archive import try_get_dicom_archive_with_archive_location, try_get_dicom_archive_with_id
+from lib.db.queries.dicom_archive import try_get_dicom_archive_with_archive_location
+from lib.db.queries.mri_upload import try_get_mri_upload_with_id
 from lib.db.queries.session import try_get_session_with_cand_id_visit_label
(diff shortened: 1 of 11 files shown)


python/lib/dcm2bids_imaging_pipeline_lib/base_pipeline.py
```diff
@@ -78,109 +75,127 @@ def __init__(self, loris_getopt_obj, script_name):
...
-                        err_msg += f"Could not load tarchive dictionary for TarchiveID {tarchive_id}"
+            mri_upload = try_get_mri_upload_with_id(self.env.db, upload_id)
+            if mri_upload is None:
+                log_error_exit(
+                    self.env,
+                    f"Did not find an entry in mri_upload associated with 'UploadID' {upload_id}",
+                    lib.exitcode.SELECT_FAILURE,
+                )
+
+            self.mri_upload = mri_upload
```
cmadjar, line 161: Couldn't this be simplified this way? Is there a reason `self.mri_upload` is not assigned right away?

```suggestion
            self.mri_upload = try_get_mri_upload_with_id(self.env.db, upload_id)
            if self.mri_upload is None:
                log_error_exit(
                    self.env,
                    f"Did not find an entry in mri_upload associated with 'UploadID' {upload_id}",
                    lib.exitcode.SELECT_FAILURE,
                )
```

python/lib/dcm2bids_imaging_pipeline_lib/base_pipeline.py
```diff
@@ -78,109 +75,127 @@ def __init__(self, loris_getopt_obj, script_name):
...
                 )
 
-                if not success:
-                    err_msg += new_err_msg
-            else:
-                err_msg += f"Could not load tarchive dictionary for ArchiveLocation {archive_location}"
-
-        if not success and not self.force:
-            log_error_exit(self.env, err_msg, lib.exitcode.SELECT_FAILURE)
+            self.dicom_archive = dicom_archive
```
cmadjar, line 181: same comment here. Is there a reason `self.dicom_archive` is not assigned directly? Since the script exits if dicom_archive is None, I think it is OK to assign it in line 173.

python/lib/dcm2bids_imaging_pipeline_lib/base_pipeline.py
```diff
@@ -78,109 +75,127 @@ def __init__(self, loris_getopt_obj, script_name):
...
+                        self.env,
+                        f"Did not find an entry in mri_upload associated with 'TarchiveID' {self.dicom_archive.id}",
+                        lib.exitcode.SELECT_FAILURE,
+                    )
+                case [mri_upload]:
+                    self.mri_upload = mri_upload
+                case _:
+                    log_error_exit(
+                        self.env,
+                       f"Found {len(mri_uploads)} rows in mri_upload for 'TarchiveID' {self.dicom_archive.id}",
```
cmadjar, line 196: might be worth printing the different UploadIDs matching to not have to query it later when debugging.

cmadjar (commented):
@maximemulder very nice!! A few minor suggestions/questions. I did not test it yet.

PR #1225, opened 2025-01-10
Pull 26.0.1 and 26.0.2 bug fixes into main

cmadjar:


maximemulder:
Nit: Probably `delete_physiological_file.py` should go into `python/scripts` ? The Python tooling (linter, type checker) are only configured to look for Python files in the `python` directory.

PR #1232, opened 2025-02-16
Migrate `mri_scanner` to the new database abstraction

maximemulder:
Builds on top of #1224. This PR migrates `lib.database_lib.mri_scanner` to the new database abstraction, and migrates the existing code that uses it to the new abstraction.

This PR does not migrate the configuration file that still indirectly uses this code, but this can be done in a follow-up PR (or this PR if wanted).


cmadjar:
@maximemulder branch has conflicts after merging #1224

PR #1233, opened 2025-02-24
Pull 26.0.1 and 26.0.2 bug fixes into main

laemtl:
Replace #1225


laemtl:
Will test and fix the static errors soon!

laemtl:
> * The blocking type check is ` Error: Import "delete_physiological_file" could not be resolved`. You can either use `from scripts.delete_physiological_file import xxx` or (preferably) put the function you need somewhere in the `lib` modules.

I left it in script because delete_physiological_file is a script, not a lib.

PR #1236, opened 2025-03-04
Fix large file hashing memory consumption

maximemulder:
Title. This is for the Python pipelines (BIDS import and DICOM to BIDS).


cmadjar:
@maximemulder this is failing two of the tests. Could you fix those?

PR #1238, opened 2025-03-06
Add `lib.util` Python module

maximemulder:
To continue gradually adding typing to the LORIS-MRI codebase, and to prevent bloating `lib` with new files, this PR adds a new `lib.util` module divided into domain-specific submodules. The `lib.util` module should stay LORIS-agnostic, although using the LORIS-specific logging functions is authorized (I might make a PR later to decouple the logging interface from the LORIS architecture but that is not a priority).

I moved the new `lib.file_system` module (which was itself extracted from `lib.imaging_io`) to `lib.util.fs` (I believe `fs` is a conventional name for file system operations). I also extracted some functions that I use in my other PRs here. Finally, I used `@deprecated` to not create a breaking change for the LORIS-MRI 27 migration.


cmadjar (approved):
LGTM

PR #1242, opened 2025-03-14
Add integration test for run_dicom_archive_validation.py

cmadjar:


test/RB_SQL/RB_mri_upload.sql
```diff
@@ -1,6 +1,6 @@
 SET FOREIGN_KEY_CHECKS=0;
 TRUNCATE TABLE `mri_upload`;
 LOCK TABLES `mri_upload` WRITE;
-INSERT INTO `mri_upload` (`UploadID`, `UploadedBy`, `UploadDate`, `UploadLocation`, `DecompressedLocation`, `InsertionComplete`, `Inserting`, `PatientName`, `number_of_mincInserted`, `number_of_mincCreated`, `TarchiveID`, `SessionID`, `IsCandidateInfoValidated`, `IsTarchiveValidated`, `IsPhantom`) VALUES (126,'cecile','2024-08-30 14:26:07','/data/incoming/MTL001_300001_V2_t1w.tgz','/data/not_backed_up/ImagingUpload-14-30-FoTt1K',0,0,'MTL001_300001_V2',NULL,NULL,74,NULL,1,0,'N');
+INSERT INTO `mri_upload` (`UploadID`, `UploadedBy`, `UploadDate`, `UploadLocation`, `DecompressedLocation`, `InsertionComplete`, `Inserting`, `PatientName`, `number_of_mincInserted`, `number_of_mincCreated`, `TarchiveID`, `SessionID`, `IsCandidateInfoValidated`, `IsTarchiveValidated`, `IsPhantom`) VALUES (126,'cecile','2024-08-30 14:26:07','/data/incoming/MTL001_300001_V2_t1w.tgz','/data/not_backed_up/ImagingUpload-14-30-FoTt1K',0,0,'MTL001_300001_V2',NULL,NULL,74,NULL,0,0,'N');
```
cmadjar, line 4: @maximemulder I changed the column `IsCandidateInfoValidated` to set it to `0` instead of `1` because subject validation is supposed to happen later on. But on the test, it keeps complaining that `IsCandidateInfoValidated` is set to True. Do you have any idea why? Is there some sort of caching happening?

python/tests/integration/scripts/test_run_dicom_archive_validation.py
```diff
@@ -0,0 +1,40 @@
+import subprocess
+
+from lib.db.queries.mri_upload import get_mri_upload_with_patient_name
+from tests.util.database import get_integration_database_session
+
+
+def test():
+
+    test_missing_upload_id_arg()
```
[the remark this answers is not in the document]
cmadjar: good to know. I will change that. Thanks!

test/RB_SQL/RB_mri_upload.sql
cmadjar: right, I had that revelation in the middle of the night. I guess I just needed to get away from my computer and stop obsessing about this 🤣 Thanks for the prompt reply though!

cmadjar:
@maximemulder HEEEEELP please :) Do you know why suddenly the dicom_loader test might be failing. Maybe I will have a revelation during the night but in case that does not come, I figured I would ask ;)

cmadjar:
Mmmm.. The file is there. Could this be an issue with S3 connection? Rerunning the tests now to check if that resolved. If so, then I guess something was wrong with S3...

PR #1244, opened 2025-03-20
Add more tests for dicom_archive_loader

cmadjar:
# Description

This adds the following tests to `test_run_dicom_archive_loader.py`:
- test invalid argument provided to the script
- test tarchive_path that does not exist
- test invalid UploadID provided
- addition of some checks post successful run:
    - check `mri_upload.number_of_minc_inserted` is correct
    - check `mri_upload.number_of_minc_created` is correct
    - check `dicom_archive.session_id` has been set
    - check location of the archive after it has been moved to year subfolder
    - check `mri_upload.insertion_complete` is set to True

It also modifies the DICOMs on the S3 bucket and the entries in the `tarchive` tables to add the localizer to the T1W to check that the localizer is indeed skipped because it is part of the series to exclude in the configuration setting `excluded_series_description`.


maximemulder (approved):
LGTM 👍, nice comments on the asserts as well. Though you don't have to use `run_integration_script(command=[...])` each time, `run_integration_script([...])` should be enough.

cmadjar:
@maximemulder why did you force push something on this branch? Now there are a lot of conflicts and 178 files changed.

cmadjar:
Replaced by https://github.com/aces/Loris-MRI/pull/1247 which does not contain that last force-push done by mistake

PR #1249, opened 2025-03-22
More ORM definitions

maximemulder:
Extracted a few changes from my other WIP PRs (refactor `get_subject_info` and incremental BIDS importer) here.

List of changes:
* Add ORM models for `cohort` and `visit`.
* Add some common queries of the shape `try_get_x_with_id` and `try_get_x_with_name`.
* Remove some unwanted newlines that Ruff introduced when removing `Optional` imports.
* Add explicit return types for all queries (NGL I liked inferred return type at first, but now I prefer to have them explicit).
* Use (properly formatted) nested function calls instead of an intermediate `query` variable.

### DIFF
--- a/python/lib/db/models/cohort.py
+++ b/python/lib/db/models/cohort.py
@@ -1,0 +1,13 @@
+from sqlalchemy.orm import Mapped, mapped_column
+
+from lib.db.base import Base
+
+
+class DbCohort(Base):
+    __tablename__ = 'cohort'
+
+    id                 : Mapped[int]         = mapped_column('CohortID', primary_key=True)
+    name               : Mapped[str]         = mapped_column('title')
+    use_edc            : Mapped[bool | None] = mapped_column('useEDC')
+    window_difference  : Mapped[str | None]  = mapped_column('WindowDifference')
+    recruitment_target : Mapped[int | None]  = mapped_column('RecruitmentTarget')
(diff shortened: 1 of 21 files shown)


cmadjar (commented):
@ridz1208 @maximemulder is it true that a visit will always be listed in the `visit` table?

In HBCD, we can see some visits in `Visit_windows` that are not in `visit` but those are all future potential visits not yet implemented for instruments where we know there will be MRI data at some point. I can just create the visit label in the visit table instead. I am just not fully sure of how things work on the BHV side regarding those tables. Basically, is it ever possible to have a `visit_label` in the `session` table that is not listed in the `visit` table (not talking about phantoms here)? 

Let me know

cmadjar:
@maximemulder right, but in the code changed, the expected list of visit labels is queried from `visit` instead of `visit_windows` (in earlier code). I am fine with that change if we can guaranty that the `visit` table always contains all visit labels we find in the `session` table (I am not referring to phantom visit labels cause those are never in visit nor visit_windows). But for the subject data, I just want to be sure nothing will break if we switch the querying from `visit_windows` to `visit`. I am actually hoping that the labels listed in `visit` is more robust than the data in `visit_windows` but I honestly don't know. @ridz1208 @driusan  would you know?

cmadjar (approved):

cmadjar:
ooooh, oops, I have been tricked by the diff visual.

I'd still like the answer to that question though @ridz1208 @driusan when you get a chance:

- is it more reliable to check the `visit` or the `visit_windows` table to determine whether a visit label is valid before creating a new session for a candidate on the MRI side?

Issue #1231, opened 2025-02-14
Python configuration file refactor


maximemulder:
After several discussions these last few weeks on the LORIS-MRI configuration files, I will summarize here the current situation of the LORIS-MRI Python configuration file, or more precisely the [`get_subject_info`](https://github.com/aces/Loris-MRI/blob/ba8ca1dc602398d74618160a3c65a1535631f825/dicom-archive/database_config_template.py#L26-L53) function, and what we would likely want to change for LORIS-MRI 27 or after.

This function is used by the LORIS-MRI scripts to get information about the session some data is attached to when processing said data. Currently, it is only used in the DICOM study to BIDS pipeline (`run_dicom_archive_loader` & co) and the [new DICOM study importer](https://github.com/aces/Loris-MRI/pull/1117).

Input:
- Database handler
- Patient name
- Scanner ID

Output:
- Candidate information (CandID, PSCID)
- Visit label
- Session creation information (project ID, cohort ID) if enabled.

This function can also return `None`  if no session is found for the given patient name.

In LORIS 26, this function returned all this information as a dictionary, in the main branch, it has been changed to use typed dataclasses instead, whose definition can be found in [this file](https://github.com/aces/Loris-MRI/blob/ba8ca1dc602398d74618160a3c65a1535631f825/python/lib/config_file.py). As such, LORIS-MRI 27 is a breaking change for the user configuration no matter what.

This function suffers from several problems:
- It does not allow the user to return the site ID to create a session, and instead uses some heuristics hard-coded in the LORIS-MRI code (see [this function](https://github.com/aces/Loris-MRI/blob/ba8ca1dc602398d74618160a3c65a1535631f825/python/lib/dcm2bids_imaging_pipeline_lib/base_pipeline.py#L203-L237)).
- The checks and uses of the output are scattered throughout the LORIS-MRI code, although it has been mainly reduced to the [`lib.validate_subject_info`](https://github.com/aces/Loris-MRI/blob/ba8ca1dc602398d74618160a3c65a1535631f825/python/lib/validate_subject_info.py) and [`lib.get_subject_session`](https://github.com/aces/Loris-MRI/blob/ba8ca1dc602398d74618160a3c65a1535631f825/python/lib/get_subject_session.py) modules.
- The `scanner_id` parameter is unreliable, it is passed at only one call site out of three.
[...]

PR #1250, opened 2025-03-25
Refactor `get_session_config`

maximemulder:
Notable changes:
* Consolidate session configuration system code into the `lib.get_session_info` module, remove previous session configuration modules.
* Add a new `lib.config` module to factorize in-database configuration values.
* Add a new `lib.bids` module to factorize pipeline-agnostic BIDS utilities.
* Separate `SessionConfig` and `SessionInfo` objects for unchecked session information (project-provided and potentially erroneous) and checked (correct and in-database) session information.
* Rename the configuration function `get_subject_info` to `get_session_config` to better describe its purpose.
* Use the new database abstraction in `get_session_config`.
* Remove the (unreliable) scanner ID parameter from `get_session_config`.
* Add support for phantoms to the Python pipeline (there was some code before but it was clearly not tested nor working). Although the back-end for phantoms is the same as the Perl pipeline, the "scanners are candidates" part is completely hidden from the configuration.
* Add support for automatic session creation for the Python pipeline (same thing for the code).
* Adapt the current pipelines to use this new configuration system.

Solves: #1231


python/lib/get_session_info.py
```diff
@@ -0,0 +1,328 @@
...
+            session_config,
+            f"No candidate found for CandID {session_config.cand_id}."
+        )
+
+    if candidate.psc_id != session_config.psc_id:
+        raise_session_config_error(
+            session_config,
+            (
+                f"Candidate {session_config.cand_id} PSCID does not match the session configuration PSCID.\n"
+                f"The candidate PSCID is '{candidate.psc_id}', the session configuration PSCID is"
```
cmadjar, line 142: instead of talking about the session configuration PSCID, maybe simplify by saying that the PSCID stored in the DICOM header does not match the CandID stored in the DICOM header. More intuitive for users not knowing the code in depth.

python/lib/get_session_info.py
```diff
@@ -0,0 +1,328 @@
...
+
+    site    = get_session_site(env, session_config, session_config.create_session.site_id)
+    project = get_session_project(env, session_config, session_config.create_session.project_id)
+    cohort  = get_session_cohort(env, session_config, session_config.create_session.cohort_id)
+
+    project_cohort = try_get_project_cohort_with_project_id_cohort_id(env.db, project.id, cohort.id)
+    if project_cohort is None:
+        raise_session_config_error(
+            session_config,
+            f"No assiciation found for project '{project.name}' and cohort '{cohort.name}'."
```
cmadjar, line 192:
```suggestion
            f"No association found for project '{project.name}' and cohort '{cohort.name}'."
```

python/lib/get_session_info.py
```diff
@@ -0,0 +1,328 @@
...
+    Raise a session configuration error for a given session configuration.
+    """
+
+    match session_config:
+        case SessionCandidateConfig():
+            session_name = f"{session_config.psc_id} / {session_config.cand_id} / {session_config.visit_label}"
+        case SessionPhantomConfig():
+            session_name = session_config.name
+
+    raise SessionConfigError(f"Error in session configuration ({session_name}):\n{message}")
```
cmadjar, line 328: same, "session configuration" terminology could be simplified to point the user to check their configuration file.

install/templates/database_config_template.py
```diff
@@ -23,31 +23,26 @@
...
-            # database yet, create it in this function.
-            cand_id      = imaging.get_scanner_candid(scanner_id),
-            visit_label  = subject_name.strip(),
-            create_visit = CreateVisitInfo(
-               project_id = 1, # Change to relevant project ID
-               cohort_id  = 1, # Change to relevant cohort ID
-            ),
+        return SessionPhantomConfig(
+            name       = patient_id,
+            site_id    = 1,  # Change to the relevant site ID.
```
cmadjar, line 37: Replace here by Alias instead as discussed during today's call.

cmadjar (changes requested):
@maximemulder great work! A few comments to maybe have more intuitive error messages + the alias thing we talked about.

cmadjar (approved):
LGTM. Thanks @maximemulder for the updates :)

PR #1251, opened 2025-03-27
small test type improvement

maximemulder:
Just noticed something that I missed during review. We checked `mri_upload.dicom_archive is not None` several times, but we only need to check it once ! The type checker is smart enough to remember that information in the following statements. I also changed a `dicom_archive.session_id` to a `dicom_archive.session` because I like it.
@cmadjar


cmadjar (approved):
Nice! 👍

PR #1263, opened 2025-04-04
Finish to adopt `lib.util.crypto`

maximemulder:
Some other PR created `lib.util.crypto` to have typed hash functions for the DICOM study importer, but these functions were not adopted in the rest of the codebase. This PR adopts them. A small difference is that the new functions raise an exception if the file does not exist, whereas the old ones return `None`, so I had to check whether there was a possibility that the files do not exist while replacing the old function calls.

This reduces the number of depreciation warnings in CI and makes it possible to delete the untyped functions after LORIS-MRI 27.

### DIFF
--- a/python/lib/aws_s3.py
+++ b/python/lib/aws_s3.py
@@ -5,7 +5,7 @@ import os
 import boto3
 from botocore.exceptions import ClientError, EndpointConnectionError
 
-import lib.utilities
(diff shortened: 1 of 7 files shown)


cmadjar (approved):
ooops, sorry I missed that detail. Was not very awake when I looked at your PR...

PR #1264, opened 2025-04-04
Adopt new DICOM study importer in the MRI pipeline

maximemulder:
Title. Replace the uses of `dicomTar.pl` by `import_dicom_study.py` in the LORIS MRI pipeline. I would prefer to have this merged for LORIS-MRI 27, but if someone has problems with it 28 is ok too.

Succesfully tested on my local LORIS VM.

### DIFF
--- a/dicom-archive/dicomTar.pl
+++ b/dicom-archive/dicomTar.pl
@@ -1,7 +1,7 @@
 #!/usr/bin/perl
 # J-Sebastian Muehlboeck 2006
 # sebas@bic.mni.mcgill.ca
-# Archive your dicom data using DICOM::DCMSUM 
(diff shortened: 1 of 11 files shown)


cmadjar:
@maximemulder could you add this to the LORIS-MRI agenda to make sure everyone is onboard with changing this? I am definitely ok with it but I'd like to be sure it would be ok for CCNA and IBIS or any other projects.

cmadjar (changes requested):
@maximemulder Could you run the following script to update the perdoc MD files?

```
mass_perldoc_md_creation.pl -profile prod
```

cmadjar (approved):

PR #1275, opened 2025-04-17
Fix ingest_eeg_bids_datasets.py

jeffersoncasimir:
Fixed bugs found in v27 testing (no issue opened)

This PR fixes the subject ID used, so that it's uniform and works both when derived from CandID and PSCID (failed test with PSCID)

It also changes the `UPDATE  electrophysiology_uploader` query to `SET Status` to values part of its `enum('Not Started','Extracted','Failed Extraction','In Progress','Complete','Failed','Archived')`

### DIFF
--- a/python/scripts/ingest_eeg_bids_datasets.py
+++ b/python/scripts/ingest_eeg_bids_datasets.py
@@ -127,7 +127,7 @@ def main():
         if os.path.isdir(
             os.path.join(assembly_bids_path, 'sub-' + str(candid))
         ):
-            subjectid = 'sub-{}'.format(candid)
+            subjectid = f'sub-{candid}'
 
         # Try the pscid, case insensitive
         if not subjectid:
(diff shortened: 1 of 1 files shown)


python/scripts/ingest_eeg_bids_datasets.py
```diff
@@ -127,13 +127,13 @@ def main():
         if os.path.isdir(
             os.path.join(assembly_bids_path, 'sub-' + str(candid))
         ):
-            subjectid = str(candid)
+            subjectid = 'sub-{}'.format(candid)
```
cmadjar, line 130:
```suggestion
            subjectid = f'sub-{candid}'
```

(@jeffersoncasimir to fix the test errors)

jeffersoncasimir: thanks

cmadjar (approved):

PR #1289, opened 2025-05-22
Fix various little bugs preventing the insertion of a MINC file without DICOMs

cmadjar:
# Description

Currently, there are a few bugs preventing from inserting a MINC file in the database without DICOMs. This PR resolves most of them.


nicolasbrossard:
Works perfectly on my test VM.

PR #1304, opened 2025-07-01
Extract `util.fs` functions from incremental BIDS importer

MaximeBICMTL:
## Description

Extract some `lib.util.fs` improvements and new functions from #1211. 

## Changelog

- Simplify the signature of `lib.util.fs.iter_all_dir_files`.
- Add some utility functions to `lib.util.fs`.

## Future works

- One of these days it would be nice to migrate our code from `os.path` to `pathlib` but that is really not my priority right now.

### DIFF
--- a/python/lib/util/fs.py
+++ b/python/lib/util/fs.py
@@ -1,8 +1,9 @@
 import os
+import re
 import shutil
 import tarfile
 import tempfile
-from collections.abc import Generator
+from collections.abc import Iterator
 from datetime import datetime
 
 import lib.exitcode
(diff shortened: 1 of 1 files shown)


python/lib/util/fs.py
```diff
@@ -77,3 +78,39 @@ def remove_empty_directories(dir_path: str):
...
+    """
+    Get the extension (including multiple extensions) of a file name or path without the leading
+    dot.
+    """
+
+    parts = file_name.split('.', maxsplit=1)
+    if len(parts) == 1:
+        return ''
+
+    return parts[1]
```
jeffersoncasimir:
```suggestion
    return parts[1:]
```

We definitely have .tar.gz in the pipeline for events

PR #1307, opened 2025-07-03
LORIS API client for real this time

MaximeBICMTL:
## Description

This is the LORIS API client I am using on C-BIG for MPN.
This PR targets LORIS-MRI 26 as this is the LORIS version C-BIG is using, another PR for 27 or main will likely come later.
The client should be ready to use as-is, it is a completely standalone project that does not need the rest of LORIS-MRI to work (unlike the last PRs), see the README for installation and usage instructions.
There is very little documentation in the project itself, but it is typed and the user can usually refer to the broader LORIS API documentation.

I hesitated on whether to place this client in this repo, the LORIS repo, or a new repo, but I figured here would probably be the best, we will probably discuss this in a meeting.

Supersedes #1206

### DIFF
--- a/loris_api_client/README.md
+++ b/loris_api_client/README.md
@@ -1,0 +1,33 @@
+# LORIS Python API client
+
+## Description
+
+This repository contains the LORIS Python API client, a Python API client to interact with the LORIS public HTTP API.
+
+## Installation
+
+To install this project, use the following command:
+
+```sh
+pip install git+https://github.com/aces/loris-mri.git@26.0-release#subdirectory=loris_api_client
+```
+
+## Example
+
+To use the LORIS Python API client, use the `LorisApiClient.connect` method to connect to the API and pass the client to the desired endpoints.
+
+```py
+from loris_api_client.client import LorisApiClient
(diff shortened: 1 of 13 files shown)


jeffersoncasimir:
@MaximeBICMTL @cmadjar @ridz1208 

Can this client instead be a standalone repo?

jeffersoncasimir:
Sorry to be picky but I would recommend a rename to `loris-python-client` or `loris-python-api-client`.

This would be in line with `aces/cbrain-php-client` (used by EEGNet) and `aces/cbrain-python-client`.

MaximeBICMTL:
Thanks for the repo @ridz1208, will push and close this PR once I have the rights on it.

I personally have no stake on the topic of the repository name. It is true that with `loris-api-client`, we kind of indicate that the Python client is the "canonical" LORIS API client (which I guess is kind of true). I guess we should discuss that in a meeting while no one depends on it yet.

MaximeBICMTL:
The new repository has been created [here](https://github.com/aces/loris_api_client). We will discuss the naming in a meeting.

Issue #1309, opened 2025-07-21
Make LORIS-MRI Python installable via `pip install`


MaximeBICMTL:
It would be nice to make LORIS-MRI Python installable with `pip install`. By "LORIS-MRI Python", I only mean the Python scripts and libraries. This does not include the Perl scripts, system dependencies, or file system structure.

Steps:
1. Add project metadata to `pyproject.toml` (PR: #1308)
2. Move dependencies from `requirements.txt` to `pyproject.toml`, and split production from development dependencies.
3. Move scripts path from `environment` to `pyproject.toml`, and possibly rename scripts (related issue: #1219).
4. Add license file and centralize license information in `pyproject.toml`.

Step 2. and 3. also require to update the installation and CI scripts.

PR #1311, opened 2025-07-26
Move dependencies to PyProject (packaging PR 3)

MaximeBICMTL:
Builds on: #1308 #1310 
Contributes to: #1309

## Description

Move the LORIS-MRI Python dependencies from `requirements.txt` to `pyproject.toml`, which is the recommended modern way of handling dependencies. This allows Python packaging tools to know what are the dependencies of LORIS-MRI Python and will further allow us to make LORIS-MRI installable as a real Python package.

## Changelog

- Moved dependencies from `requirements.txt` to `pyproject.toml`.
- Split main dependencies from dev dependencies.
- Update the installation script and CI pipelines to install from `pyproject` instead of `requirements.txt`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,10 @@ dev = [
 [project.urls]
 Homepage = "https://github.com/aces/loris-mri"
 
+# TODO: Remove this statement once a real build backend has been configured.
+[tool.setuptools]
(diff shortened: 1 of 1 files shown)


jeffersoncasimir (approved):
Looks good

PR #1318, opened 2025-09-14
Factorize Python configuration loading and add default configuration value (configuration PR 3)

MaximeBICMTL:
## Description

Factorize the "Python configuration loading" code (which is not decoupled from `LorisGetOpt`) and use the default `config.py` value if no argument is given to the script.


jeffersoncasimir (approved):
LGTM

PR #1319, opened 2025-09-21
Add `IntBool` ORM decorator (ORM PR 2)

MaximeBICMTL:
## Description

In the LORIS SQL schema, many boolean fields are actually represented as `TINYINT(1)` instead of `BOOL`. Until now, the LORIS-MRI Python ORM used these as `BOOL` fields, which is compatible with `TINYINT(1)` in MariaDB and MySQL. This PR improves the preciseness of the ORM by adding a decorator to accurately represent these "boolean integers" as integers in the SQL side, and booleans in the Python side.

## Changelog

- Add a `IntBool` ORM decorator.
- Remove the special case in integration tests that handles boolean integers.
- Update the ORM models with the new decorator.


jeffersoncasimir (approved):
Approved!

PR #1324, opened 2025-10-30
Batch Maxime PRs

MaximeBICMTL:
This PR batches several of my PRs for easier testing and rebasing. The code of all the PRs included in this branch have been reviewed and approved by @jeffersoncasimir.

**DO NOT SQUASH WHEN MERGING** so that each individual PR appears in the git history.

### DIFF
--- a/python/lib/config_file.py
+++ b/python/lib/config_file.py
@@ -2,7 +2,6 @@
 This module stores the classes used in the Python configuration file of LORIS-MRI.
 """
 
-import importlib
 import importlib.util
 import os
 import sys


cmadjar:
@jeffersoncasimir @MaximeBICMTL Two questions coming from testing the BIDS import on CBIGR today, not sure if that has implications on the PRs sent to this repo:

- even if the bids validation is turned off, it requires dataset_description.json in the BIDS dataset to import. Is that expected?
- got an error that -b option does not exist when using `-b` short cut option. We had to use `--nobidsvalidation` long option instead
- the creation of session via the BIDS importer did not work (@jeffersoncasimir did you try the creation of visits and candidates when testing the PR?)

 Maybe those are just happening on CBIGR but just in case, figured I would mention it here.

MaximeBICMTL:
Hi @cmadjar, thank you for your feedback on the BIDS importer!

Though note that this PR does **not** contain changes to the BIDS importer, the BIDS importer used in C-BIG and that I want to merge here is #1325. This PR only contains repository-wide changes that do not affect the behavior of scripts.

I answered your points in the relevant BIDS importer PR.

PR #1338, opened 2025-11-20
Make the EEG chunker into a standalone Python package

MaximeBICMTL:
## Description

Make the LORIS EEG chunker into its own Python package. This allows to better isolate this pipeline from the rest of LORIS Python by having it declare its own dependencies and make sure it does not depend on the rest of LORIS Python. It also makes the LORIS EEG chunker installable without the rest of LORIS Python using `pip install`.

## Details

- Added the `loris-eeg-chunker` package metadata.
- Structure the `loris-eeg-chunker` package using the conventional `src`-layout.
- Make sure the main LORIS Python package depends on `loris-eeg-chunker`.

## Testing instructions

If you want to test this PR manually, either use:
- `pip install .` in the LORIS-MRI root directory, which will install both LORIS Python along with the LORIS EEG chunker as a dependency.
- `pip install python/loris_eeg_chunker` which will only install the LORIS EEG chunker (the main codebase can still rely on environment variables if it is not installed as a package).

## Migration

Note that the LORIS EEG chunker scripts have been renamed using the `[project.scripts]` Python feature (instead of the `PATH` environment variable):
- `edf_to_chunks.py` to `edf-to-chunks`
- `eeglab_to_chunks.py` to `eeglab-to-chunks`

The LORIS Python scripts that call these subscripts have been updated, so this is only notable if a project calls these scripts directly.


jeffersoncasimir (changes requested):
Tested via `bids_import.py`. Works seamlessly.

The only issue is that the `EEGChunksPath` config is not in core, but in an [unmerged draft PR](https://github.com/aces/Loris/pull/9523).

jeffersoncasimir (changes requested):
#1342 (which I approved) changed the expected chunk path. It seems that the `os.path` replacement must not have been equivalent.

My `EEGChunksPath`: `/data/Loris-MRI/data/bids_imports_chunks/`. Chunks ended up in `{EEGChunksPath}/sub-PIDCC0821_ses-V03_task-FACE_acq-eeg_eeg.chunks` but should have ended up in `{EEGChunksPath}/{DatasetName}_chunks/sub-PIDCC0821_ses-V03_task-FACE_acq-eeg_eeg.chunks`

My `physiological_parameter_file.Value` for `electrophysiology_chunked_dataset_path` ended up as `sub-PIDCC0821_ses-V03_task-FACE_acq-eeg_eeg.chunks`, but should have been `bids_imports/{DatasetName}_chunks/sub-PIDCC0821_ses-V03_task-FACE_acq-eeg_eeg.chunks` to match RB examples

So the front-end is looking at the wrong place for the `index.json` and the chunks are also not saved in the right place

Let me know how your setup looks and ends up, if you think the problem is on my end. The PR is otherwise ready, but it seems like the right time to get the above straightened out.

MaximeBICMTL:
Oooohokay I think I understand better thanks @jeffersoncasimir. In my tests I did not have `EEGChunksPath` configured and it worked well. But in your case you seem to have it configured and it causes some issues with the code. I will look into this second setup thanks.

jeffersoncasimir (approved):
Tested with https://github.com/aces/Loris-MRI/pull/1345 and works great!

Ready to merge

PR #1342, opened 2025-11-21
Use `os.path` instead of string operations in EEG

MaximeBICMTL:
Extracted from #1325 

## Description

The BIDS importer refactor changes path manipulation from string concatenation (which is a very fragile way of doing things, notably as it is very sensible to trailing slashes) to `pathlib.Path` or `os.path` functions and methods. Unfortunately, doing so also requires to update EEG code to this new technique as string-based path manipulation does not work with modern paths (because of the absence of trailing slashes).

## Details

- Use `os.path.join` in EEG code instead of string concatenation.
- Use `os.path.relpath` in EEG code instead of string replacement.
- Store paths in a list instead of a growing tuple.

### DIFF
--- a/python/lib/eeg.py
+++ b/python/lib/eeg.py
@@ -310,21 +310,21 @@ class Eeg:
             )
 
             # archive all files in a tar ball for downloading all files at once
-            files_to_archive = (os.path.join(self.data_dir, eeg_file_path),)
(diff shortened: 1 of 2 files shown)


jeffersoncasimir (approved):
Tested and works as expected!

PR #1346, opened 2025-12-05
Add database checks to the EEG BIDS import integration test

MaximeBICMTL:
## Description

This PR does what the title says it does. The code is rather straightforward. The only somewhat tricky part is that the tests failed because the imported file was already present in Raisinbread. I added some overrides to remove existing physiological files, which is a little dirty IMO but should do the job for now.

### DIFF
--- a/python/lib/db/queries/physio_file.py
+++ b/python/lib/db/queries/physio_file.py
@@ -1,11 +1,7 @@
-from collections.abc import Sequence
-
 from sqlalchemy import select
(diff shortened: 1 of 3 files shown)


MaximeBICMTL:
I must say I am puzzled by the integration test failure, it seems like there is some `OTT167` files in the test files ?

jeffersoncasimir (approved):
I approve these changes and new helper functions

PR #1359, opened 2026-01-08
Miscellaneous utility improvements

MaximeBICMTL:
## Description

Small improvements to the utilities module extracted from #1335.

## Details

- Move path-related functions (= no file system operation) out of `lib.util.fs` into a new `lib.util.path` module.
- Re-order parameters of iterator/collection functions such that the iterator/collection is always the first parameter, which is arguably more intuitive (can expand on this if needed).
- Note that all the changed/moved functions in this PR are actually not used in the codebase yet(they were extracted from a previous version of #1335), so there is no need to change the call sites . However, I'd like to extract other things from #1335 that build upon those, hence this PR first.

### DIFF
--- a/python/lib/util/fs.py
+++ b/python/lib/util/fs.py
@@ -82,37 +82,14 @@ def remove_empty_directories(dir_path: str):
             os.rmdir(subdir_path)
 
 
-def get_file_extension(file_name: str) -> str:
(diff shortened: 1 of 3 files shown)


jeffersoncasimir (approved):
Approved! 

Tested with `bids_import.py`

PR #1360, opened 2026-01-09
Typed database configuration getters for EEG/physiology

MaximeBICMTL:
## Description

Extract the EEG/phyisology-related typed database configuration getters from #1335 and adopt them in the current EEG/phyisology pipelines.

## Details

- Add an `env` parameter to the `Eeg` and `Physiological` constructors to access the typed database abstraction.
- Add EEG/physiology-related getters to `lib.config`.
- Use `lib.config.get_*` instead of the old `Config` objects to get the related configuration values.

## Future works

The next PR (which builds upon this one and #1359) will be to factorize the chunking subscript calling.

### DIFF
--- a/python/lib/config.py
+++ b/python/lib/config.py
@@ -79,6 +79,84 @@ def get_dicom_archive_dir_path_config(env: Env) -> Path:
     return dicom_archive_dir_path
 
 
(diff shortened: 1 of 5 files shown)


jeffersoncasimir (approved):
Manually tested and worked perfectly!

PR #1363, opened 2026-01-16
BIDS dataset description reader (LORIS BIDS reader PR 1)

MaximeBICMTL:
## Description

First PR in the LORIS BIDS reader effort, which adds a new `loris-bids-reader` package as well as the `BidsJsonFile` and `BidsDatasetDescriptionJsonFile` classes, and starts to adopt them in the current codebase.

## Design choices

- `loris-bids-reader` is a new Python package, it is designed to work as a standalone library independent of LORIS, similarly to the `bids-eeg-chunker` package.
- There is a `BidsJsonFile` abstract class:
    - This class is inherited by all file-specific BIDS JSON classes (such as `BidsDatasetDescriptionJsonFile`).
    - The file-specific classes can then implenebt there own validation logic, be used as type hints, and add links to the relevant BIDS documentation in their docstrings.
- As discussed in the last imaging meeting, this approach does not use the Pydantic-based validation approach in order to avoid the _double source of truth_ problem. 
- Instead, we only validate manually the fields that we really need in our pipelines. This is the approach that we currently use in our codebase. While it offers weaker typing and validation, it results in a lot less code to maintain.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,7 @@ dependencies = [
     "scipy",
     "sqlalchemy>=2.0.0",
     "virtualenv",
+    "loris-bids-reader @ {root:uri}/python/loris_bids_reader",
     "loris-eeg-chunker @ {root:uri}/python/loris_eeg_chunker",
 ]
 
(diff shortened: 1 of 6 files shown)


MaximeBICMTL:
Approved in #1369. Rebased and merge.

PR #1379, opened 2026-02-09
Factorize BIDS information and simplify `Eeg` paramaters (PR 6)

MaximeBICMTL:
Builds on #1378 ([diff](https://github.com/MaximeBICMTL/LORIS-MRI/compare/harness_physio_orm_power...MaximeBICMTL:LORIS-MRI:bids_info))

## Description

- Add a new `lorids_bids_reader.info` module to store typed information about a BIDS subject/session/data type directory.
- Use a new `BidsDataTypeInfo` object instead of separate untyped parameters in the `Eeg` class.
- Use `env.verbose` instead of a separate `verbose` parameter in the `Eeg` class.


jeffersoncasimir (approved):
Approved and successfully tested with `bids_imports.py`

PR #1384, opened 2026-02-22
Factorize copy BIDS files (PR 1)

MaximeBICMTL:
Extracted from #1335

## Description

Factorize the "get BIDS file path in LORIS" and "copy BIDS file to LORIS" operations in the `lib.import_bids_dataset.copy_files`, and adopt these factorizations in the `Eeg` class. The goal of this PR is to be able to share the same logic/function across all modalities (MRI / EEG / MEG) once the other PRs land.

## Notes

- There was previously an `inheritance` parameter to the `Eeg.copy_file_to_loris_bids_dir` but it was actually never used. 
- I am not sure our rename strategy works in all cases: we rename BIDS files manually, but their might be cross-references to these files across the dataset. In any case, this logic is not worse than the previous one (it's the same logic). Let's factorize first and fix later IMO.

### DIFF
--- a/python/lib/import_bids_dataset/copy_files.py
+++ b/python/lib/import_bids_dataset/copy_files.py
@@ -91,7 +91,12 @@ def copy_scans_tsv_file_to_loris_bids_dir(
 
     original_file_path = scans_file.path
     loris_file_name = get_loris_bids_file_name(scans_file.path.name, session)
-    final_file_path = os.path.join(loris_bids_root_dir, f'sub-{session.candidate.psc_id}', loris_file_name)
+    final_file_path = (
+        loris_bids_root_dir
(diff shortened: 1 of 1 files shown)


jeffersoncasimir (changes requested):
Probably the final issue.

The chunks folder created does not contain the injected default session: `sub-ABC0035_task-resting_eeg.chunks`, but the script looks for the version with the session injected:

`ERROR: Chunk creation failed, directory '/data/Loris-MRI/data/bids_imports/EEGManyLabs_Resting_State_Dataset_Eimer1996_BIDSVersion_1.9.0_chunks/sub-ABC0035_ses-V1_task-resting_eeg.chunks' does not exist.
`

PR #1393, opened 2026-03-16
Typed physiological archives (PR 1)

MaximeBICMTL:
Builds on #1389 ([diff](https://github.com/MaximeBICMTL/LORIS-MRI/compare/typed_mri_import...MaximeBICMTL:LORIS-MRI:typed-physio-archives))

## Description

Move the "create physiological file archive" and "create physiological event archive" from the untyped `Eeg` class to a new `lib.import_bids_dataset.archive` typed module.

## Bug fix

The old code contained this function, which returns a tuple with the relative and full path of the archive:

```py
def get_archive_paths(self, archive_rel_name):
        package_path = get_eeg_pre_package_download_dir_path_config(self.env)
        if package_path:
            raw_package_dir = os.path.join(package_path, 'raw')
            os.makedirs(raw_package_dir, exist_ok=True)
            archive_rel_name = os.path.basename(archive_rel_name)
            archive_full_path = os.path.join(raw_package_dir, archive_rel_name)
        else:
            archive_full_path = os.path.join(self.data_dir, archive_rel_name)

        return (archive_rel_name, archive_full_path)
```

As can be seen, the relative path is relative to the `${package_path}/raw` directory if the `package_raw` configuration is defined, and relative to the LORIS data directory otherwise. This is a bug, it means that the path of an archive cannot be interpreted without reading the LORIS configuration (although I am rather confident that only one of these two paths works in practice), and that changing setting/unsetting that configuration breaks all archive paths in the database.

This PR makes all paths relative to the LORIS data directory, which fixes this issue and is a common practice in the LORIS database but may be a breaking change for some projects.

## Details

- Shorten `DbPhysioFileArchive.file_path` and `DbPhysioEventArchive.file_path` to  `DbPhysioFileArchive.path` and `DbPhysioEventArchive.path`.
- Add a new typed `loris_utils.archive` module.
- Add a new `lib.import_bids_dataset.physio_archive` module.
- Adopt the aforementioned module in the BIDS EEG import pipeline.
- Fix the aforementioned bug.
- Improve the BIDS EEG import integration test with archive checking.
- Remove or deprecate the old code.
- Raise an exception if the archive file already exists instead of ignoring it.
- Remove a check to know if the file already has an archive. The file sent argument has just been created by the import pipeline, it cannot already have an archive.

### DIFF
--- a/python/lib/database_lib/physiological_event_archive.py
+++ b/python/lib/database_lib/physiological_event_archive.py
@@ -1,6 +1,9 @@
 """This class performs database queries for the physiological_event_archive table"""
 
+from typing_extensions import deprecated
 
+
+@deprecated('Use `lib.db.physio_event_archive.DbPhysioEventArchive` instead')
 class PhysiologicalEventArchive:
 
     def __init__(self, db, verbose):
@@ -17,6 +20,7 @@ class PhysiologicalEventArchive:
         self.table = 'physiological_event_archive'
         self.verbose = verbose
 
+    @deprecated('Use `lib.db.physio_event_archive.DbPhysioEventArchive.physio_file_id` instead')
     def grep_from_physiological_file_id(self, physiological_file_id):
         """
         Gets rows given a physiological_file_id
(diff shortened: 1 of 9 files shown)


MaximeBICMTL:
Got merged as part of #1423.

PR #1394, opened 2026-03-16
Physiological file download path

MaximeBICMTL:
Sister PR to https://github.com/aces/Loris/pull/10411, add the `DbPhysioFile.download_path` field and adapt the insertion code to use it.

Also, document the difference between `DbPhysioFile.path` and `DbPhysioFile.download_path` in the ORM (I believe we should do that a lot more).


MaximeBICMTL:
No longer needed after https://github.com/aces/Loris/pull/10481

PR #1395, opened 2026-03-16
Simplify EEG/MEG signal chunking signature

MaximeBICMTL:
Previously, the `lib.physio.chunking.create_physio_channels_chunks` took a separate parameter for the path of the file to chunk, this was done for the work-in-progress MEG implementation where in a previous design, the path of the "real" file may not have been the path of the physiological file. That work-in-progress MEG design is obsolete thanks to https://github.com/aces/Loris/pull/10411, so this additional parameter is no longer needed. 

This PR also fixes a bug where the created chunk directories had an incorrect name.

### DIFF
--- a/python/lib/eeg.py
+++ b/python/lib/eeg.py
@@ -222,9 +222,8 @@ class Eeg:
             )
 
             # create data chunks for React visualization
-            eeg_viz_enabled = get_eeg_viz_enabled_config(self.env)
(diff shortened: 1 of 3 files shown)


jeffersoncasimir (approved):
Works great!

Not introduced in this PR, but notice how there is an issue if `-s == -l`

PR #1401, opened 2026-03-22
Factorize check LORIS directory (PR 2)

MaximeBICMTL:
Builds on top of #1393 ([diff](https://github.com/MaximeBICMTL/LORIS-MRI/compare/typed-physio-archives...MaximeBICMTL:LORIS-MRI:factorize-check-dir-config))

## Description

In the `lib.config` module, many functions get LORIS-specific directories and check their existence/permissions. There is a lot of repeated code here, that this PR factorizes. This PR also renames some electrophysiology-related configuration directory functions for better clarity (notably `eeg` -> `ephys` since those are also used by MEG).

### DIFF
--- a/python/lib/config.py
+++ b/python/lib/config.py
@@ -34,22 +34,7 @@ def get_data_dir_path_config(env: Env) -> Path:
     """
 
     data_dir_path = Path(_get_config_value(env, 'dataDirBasepath'))
-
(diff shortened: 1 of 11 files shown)


MaximeBICMTL:
Got merged as part of #1423.

PR #1408, opened 2026-04-03
Reorganize `imaging_lib` into `mri` module

MaximeBICMTL:
## Description

Before #1389, the `lib.mri` module was used by the BIDS MRI importer, but that code has now been moved to `lib.import_bids_dataset`. As such, we can use the `lib.mri` module for pipeline-agnostic MRI-specific code.

## Changelog

- Rename `lib.imaging_lib` to `lib.mri`.
- Rename `lib.imaging_lib.mri_scanner` to `lib.mri.scanner`.
- Moved `lib.imaging_lib.parameter` to `lib.parameter` (since it is not specific to MRI).

Note that the `lib.imaging_lib` module does not exist in LORIS Python 27 and was created in the LORIS Python 28/29 cycle. So this PR in itself is not a breaking change (but functions that were moved to `lib.imaging_lib` during that cycle are now moved to `lib.mri` instead).


MaximeBICMTL:
Closing since the code in `lib.imaging_lib` could also be used for CT or PET in the future.

PR #1422, opened 2026-05-04
Fix LORIS Python electrophysiology chunks directory installation process

MaximeBICMTL:
## Problem

CI has been failing for a few days due to the LORIS Python installation process being broken since https://github.com/aces/Loris/pull/10455.

## Solution

Properly create the electrophysiology directories and set the database configuration during the installation process.

## Follow-up

We have a mismatch right now where the default install creates the `/data/$project/data/thingies` directories but sets up the `/data/$project/thingies` database configuration. If you often reset raisinbread you know what this is. We should fix that one day to make the installation process smoother, but this is obviously out-of-scope for that PR.


jeffersoncasimir (approved):
LGTM

PR #1424, opened 2026-05-14
Typed HED code (PR 1)

MaximeBICMTL:
## Description

Move HED-related code to a dedicated module, add typing, and remove the `Physiological` dependency in the events-related modules.

## List of changes

- Move the HED code from the `Physiological` class into a new `lib.physio.hed` module (the code itself is generally unchanged and still very cursed).
- Also move the `lib.utilities.assemble_hed_service` function, although it is currently unused in the codebase.
- Add typing to said code. Note that the typing is incomplete as `Any` is still used in a few places where I could not get the full type.
- Use the ORM `DbHedSchemaNode` model instead of raw queries/dicts for the `hed_union` variable.


MaximeBICMTL:
Approved by @jeffersoncasimir as part of #1426.

PR #1426, opened 2026-05-18
Better BIDS TSV reader typing (PR 3)

MaximeBICMTL:
Builds on #1425 ([diff](https://github.com/MaximeBICMTL/LORIS-MRI/compare/bids-importer-package...MaximeBICMTL:LORIS-MRI:better-bids-tsv-typing))

## Description

Improve the LORIS BIDS reader TSV row typing by typing row data as `dict[str, str | None]` instead of `dict[str, Any]`. This loose typing was inherited from the old code, which read and modified the row data directly, which notably caused some bugs such as code using `isisntance(row.data['field'], int | float)` which is always `False` since raw tsv data is always a string (or `None`). The new approach stores the data LORIS is interested in in dedicated fields while the raw data remains unchanged.

## Potential follow-ups

1. Replace all tsv `n/a` values by `None` on parsing (I think we should do that, it makes the code easier and more intuitive with no real information loss IMO). 
2. Accept empty tsv rows (technically incorrect, but might appear in real data and does not hurt IMO, the code currently errors on trailing newlines). **EDIT: Added in this PR**

PR #1427, opened 2026-05-21
Typed chunking script

MaximeBICMTL:
## Description

Make the `mass_electrophysiology_chunking.py` script typed, also fixing a type issue in the process (`make_chunks` was called with four arguments instead of two). Also commit the database changes directly in `create_physio_channels_chunks`, since it would not make sense to create the chunks but not add the associated parameters to the file.

## Breaking changes

- Rename the options `--smallest_id` and `--largest_id` to `--smallest-id` and `--largest-id` to be more consistent with newer scripts.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,6 +92,7 @@ include = [
     "python/lib/get_session_info.py",
     "python/lib/logging.py",
     "python/lib/make_env.py",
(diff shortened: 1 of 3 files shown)


jeffersoncasimir (approved):
Works as expected! Tested with import-bids-dataset and https://github.com/aces/Loris-MRI/pull/1426

PR #1432, opened 2026-06-04
Small BIDS importer improvements

MaximeBICMTL:
Small BIDS improrter improvements based on the Q1K tests for C-BIG (https://github.com/ridz1208/Loris-MRI/pull/14):
1. Catch EEG import exceptions so the script can be somewhat incremental for EEG.
2. Display the _unique_ data types found in the BIDS dataset rather than duplicates.


jeffersoncasimir (approved):
Approved

PR #1433, opened 2026-06-04
Adopt `Path` in `load_config` and check for directory traversal

MaximeBICMTL:
- Use `pathlib.Path` instead of `os.path` in the `load_config` function (which loads the LORIS Python configuration from the provided `--profile` value).
- Check that the provided `--profile` value does not escape outside of the `LORIS_CONFIG` directory, preventing path traversal.


jeffersoncasimir (approved):
Approved

PR #1435, opened 2026-06-05
[Everything] DIE MRI ALIAS DIE part 2

hansfauer:
Supersedes #1227

### DIFF
--- a/python/lib/db/models/site.py
+++ b/python/lib/db/models/site.py
@@ -19,5 +19,6 @@ class DbSite(Base):
     contact_1  : Mapped[str | None]  = mapped_column('Contact1')
     contact_2  : Mapped[str | None]  = mapped_column('Contact2')
     alias      : Mapped[str]         = mapped_column('Alias')
+    mri_alias  : Mapped[str]         = mapped_column('MRI_alias')
     account    : Mapped[str | None]  = mapped_column('Account')
     study_site : Mapped[bool | None] = mapped_column('Study_site', YNBool)
--- a/python/lib/db/models/site.py
+++ b/python/lib/db/models/site.py
@@ -19,6 +19,5 @@ class DbSite(Base):
     contact_1  : Mapped[str | None]  = mapped_column('Contact1')
     contact_2  : Mapped[str | None]  = mapped_column('Contact2')
     alias      : Mapped[str]         = mapped_column('Alias')
-    mri_alias  : Mapped[str]         = mapped_column('MRI_alias')
     account    : Mapped[str | None]  = mapped_column('Account')
     study_site : Mapped[bool | None] = mapped_column('Study_site', YNBool)
--- a/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
+++ b/uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
@@ -2328,9 +2328,8 @@ sub getCenterNameFromCenterID {
     my $alias = undef;
 
     if ($centerID) {
-    $query = "SELECT MRI_alias FROM psc "
-            . "WHERE CenterID =?";
-    my $sth = $dbh->prepare($query);
+        $query = "SELECT MRI_alias FROM psc WHERE CenterID =?";
+        my $sth = $dbh->prepare($query);
         $sth->execute($centerID);
         if ( $sth->rows > 0 ) {
            $alias = $sth->fetchrow_array;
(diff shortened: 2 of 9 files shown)


python/lib/db/models/site.py
```diff
@@ -19,6 +19,5 @@ class DbSite(Base):
     contact_1  : Mapped[str | None]  = mapped_column('Contact1')
     contact_2  : Mapped[str | None]  = mapped_column('Contact2')
     alias      : Mapped[str]         = mapped_column('Alias')
-    mri_alias  : Mapped[str]         = mapped_column('MRI_alias')
```
MaximeBICMTL, line 22: Happy to see a new contributor to the Python codebase! I wrote a large part of it so here is some feedback on this change.

This is our ORM model, it must _exactly_ match the LORIS SQL schema so you cannot remove a field if it is present in the LORIS SQL schema. In fact, you can see that an integration test failed because it detected that a field is missing after this PR!

Also, you removed this field but not its uses in the codebase, which is a bug (and why the type checking static test is failing)!

Here is what I would do if I were you:
- Un-remove this ORM field. It will be removed later once the field is also removed in the LORIS SQL schema.
- However, do remove the _uses_ of that field in the codebase. You should be able to do that with "right click > find references" in your editor, or use ctrl+F if that does not work.

If you have any question about the Python codebase do not hesitate to ask. You can also find more documentation about our database abstraction and tooling [here](https://github.com/aces/Loris-MRI/tree/main/docs/python).

MaximeBICMTL (changes requested):
Hi Hans. Thanks for the PR. I don't know if anyone has volunteered to review the PR, especially since it touches the Perl code, but if not I guess we need to find someone at the next imaging meeting.

python/lib/db/models/site.py
hansfauer: ok sounds good, thanks for the feedback!

hansfauer: I didnt see many instances of mri_alias in the python code, so I removed the one i found and reinserted the reference back to the ORM model

MaximeBICMTL (approved):
Thanks for the changes! The Python changes look good to me, the CI failing is unrelated to this PR don't worry about it (another PR updated the LORIS SQL schema so we will have to update the ORM).

I'll put "approve changes" to remove my previous "request changes", but ideally someone should also review the Perl side of things. You can maybe try to find someone on Slack, or if not we will find someone at next week's imaging meeting.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -2309,6 +2307,37 @@ sub getUploadIDUsingTarchiveSrcLoc {
...
+
+    if ($centerID) {
+        $query = "SELECT MRI_alias FROM psc WHERE CenterID =?";
+        my $sth = $dbh->prepare($query);
+        $sth->execute($centerID);
+        if ( $sth->rows > 0 ) {
+           $alias = $sth->fetchrow_array;
+        }
+    }
+    return $alias;
```
nicolasbrossard: If the DB statement above succeeds, you'll return an array since that's what `fetchrow_array` returns. I would returning `$alias[0]` instead since the goal of this routine is to return a single element.

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -2309,6 +2307,37 @@ sub getUploadIDUsingTarchiveSrcLoc {
...
+Gets the MRI ALIAS form the C<psc> table using the center ID.
+
+INPUT: Center ID
+
+RETURNS: MRI ALIAS
+
+=cut
+
+
+sub getCenterNameFromCenterID {
```
nicolasbrossard, line 2323: Since you are retrieving the MRI_alias for a given center ID, wouldn't it make more sense to call this method `getMRIAliasForCenterID`? Or maybe the point of this PR is to steer away from the "MRI alias" terminology...

nicolasbrossard (changes requested):

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
hansfauer: Yes i believe the PR is supposed to use centerID now and remove mri_alias field from db

PR #1436, opened 2026-06-07
Rename `loris-bids-reader` to `loris-bids-utils`

MaximeBICMTL:
## Description

We currently read BIDS through our `loris-bids-reader` package. However, nothing stops us from adding more BIDS-related utilities to that package (which would in fact be desirable) ! Therefore, this PR renames this package to `loris-bids-utils`, which also is a nice parallel with the `loris-utils` package: the latter provides _general purpose_ utilities, the former provides _BIDS-specific_ utilities, both do not depend on having a LORIS instance present.


jeffersoncasimir (approved):
Approved

MaximeBICMTL:
Fixed merge conflicts (imports).

PR #1437, opened 2026-06-09
Remove `Scan_done` field from the Python ORM

MaximeBICMTL:
Accomodate of https://github.com/aces/Loris/pull/10236 to the Python side. The `Scan_done` field does not seem to be used outside of the ORM, which this PR updates accordingly. It does seem to be used in Perl, but that is a subject for another PR.

Also accomodate for https://github.com/aces/Loris/pull/10512 by adding the biobank tables to the test database (required to load Raisinbread).


MaximeBICMTL:
Merging without approval since this targets a staging branch, is very straightforward, and fixes CI.

PR #1440, opened 2026-06-14
Use ORM for MRI files associated files

MaximeBICMTL:
A local branch I had on my laptop.

List of changes:
- Use the ORM to get the associated files of an MRI files in the S3 upload pipelime.
- Also use the new `try_get_mri_file_parameter` function in the NIfTI pic script, which is safer than the previous function as it is restricted to MRI parameters instead of all parameters.
- Deprecate the old untyped functions.


MaximeBICMTL:
Nevermind code is wrong and I am too lazy to fix that.

PR #1443, opened 2026-06-17
In-database BIDS file information

MaximeBICMTL:
This PR implements https://github.com/aces/Loris/pull/11169 for the LORIS BIDS MRI and EEG importer. Code-wide, this PR improves the BIDS importer by refactoring `BidsImportEnv` into a more feature-complete `BidsImporter` class. However, the code for actual BIDS registration also complexifies the importer quite a bit, and simplifying it would require a large (but desirable) refactor of the BIDS importer (specifically create a common factorization for all BIDS file imports), which is outside of the scope of this PR. Integration tests are nonetheless improved in this PR to ensure that everything works as expected and continues to do so in the future.
## BIDS importer design

- Rename `BidsImportEnv` to `BidsImporter`.
- `BidsImporter` is the central object for BIDS import, which contains:
  - Arguments passed to the BIDS importer.
  - State of the BIDS importer (successed/skipped/failed files).
  - Custom importers hooked to the LORIS BIDS importer.


jeffersoncasimir (approved):
Works great with bids importer!

Not tested with dicom to bids converter

MaximeBICMTL:
The BIDS tables are not integrated with the DICOM to BIDS converter yet, which means the converter still works but does not populate the new BIDS tables. The only change in this PR is to update a function call to pass a new `None` argument for now.

In the future we certainly do want to integrate it. There is a easy and a hard way to do so, I think I will try the hard one which will result in more flexible pipelines overall.

PR #1446, opened 2026-06-18
[MRI] Check if CenterID is defined before going further

hansfauer:
Supercedes: #1228

Validate config-provided CenterID against the database in getPSC

When the config (e.g. the prod file) already provides a CenterID,
check it exists in the psc table and return it, instead of trusting it
blindly or falling through to the error-prone patient-name matching.

Raises an error if the CenterID is invalid. (Or maybe simply return 0 instead?)
Update the getPSC POD to reflect that it now returns the CenterID.

### DIFF
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -1086,8 +1086,7 @@ sub createNewCandID {
 
 Looks for the C<CenterID> of a scan. If the configuration already provides a
 C<CenterID>, it is validated against the C<psc> table. Otherwise, the C<CenterID>
-is looked up using the C<session> table (for cases where it is created using the
-front-end), and as a last resort by matching the site alias against whatever field
+is looked up using the C<session> table, and as a last resort by matching the site alias against whatever field
 (usually C<patient_name> or C<patient_id>) is provided.
 
 INPUTS:
--- a/uploadNeuroDB/NeuroDB/MRI.pm
+++ b/uploadNeuroDB/NeuroDB/MRI.pm
@@ -1084,19 +1084,18 @@ sub createNewCandID {
 
 =head3 getPSC($patientName, $dbhr, $db)
 
-Looks for the site alias using the C<session> table C<CenterID> as
-a first resource, for the cases where it is created using the front-end,
-otherwise, find the site alias in whatever field (usually C<patient_name>
-or C<patient_id>) is provided, and return the C<MRI_alias> and C<CenterID>.
+Looks for the C<CenterID> of a scan. If the configuration already provides a
+C<CenterID>, it is validated against the C<psc> table. Otherwise, the C<CenterID>
+is looked up using the C<session> table (for cases where it is created using the
+front-end), and as a last resort by matching the site alias against whatever field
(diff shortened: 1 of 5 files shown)


hansfauer:
Remove the 'die' and return 0 instead. Also remove mri_alias in the perl code as well

hansfauer:
Following up on the imaging meeting, this PR now updates the safety check (if statement in getPSC) and finishes the Perl-side MRI_alias removal, addressing issue #1447 .

In getPRC in MRI.pm it updates the safety check to return 0 rather than throw an error with die.

mri_alias was removed from all code in loris_mri now: 
Session queries and patient name matching now use CenterID and Alias instead. And PSCOB no longer fetches the mri_alias column. 
the getCenterNameFromCenterID function now returns the human-readable psc.Name instead of the MRI alias

uploadNeuroDB/NeuroDB/MRI.pm
```diff
@@ -1110,36 +1108,48 @@ sub getPSC {
                             $dbhr,
                             $db
                         );
+    ## If the configuration (e.g. the prod file) already provides a CenterID,
```
nicolasbrossard, line 1111: The `PSCOB` class should be used to perform that validation instead of directly querying the database.

uploadNeuroDB/tarchiveLoader.pl
```diff
@@ -377,7 +377,7 @@ =head2 Methods
 ################################################################
 my $centerID =
      $utility->determinePSC(\%tarchiveInfo, 0, $upload_id);
-my $mri_alias = $utility->getCenterNameFromCenterID($centerID);
+my $center_name = $utility->getCenterNameFromCenterID($centerID);
```
nicolasbrossard: The `$center_name` variable is defined here and used 300 lines below...The definition should be done way closer to where the variable is actually used.

uploadNeuroDB/tarchiveLoader.pl
```diff
@@ -606,7 +606,7 @@ =head2 Methods
 ################################################################
 # make final logfile name without overwriting phantom logs #####
 ################################################################
-my $final_logfile = $mri_alias;
+my $final_logfile = $center_name;
```
nicolasbrossard, line 613: You should define the `$center_name` variable right above that line.

nicolasbrossard (changes requested):
Since you are modifying some of the subroutine's documentation, you'll have to regenerate their associated `.md` files (ask me how to do that).

uploadNeuroDB/NeuroDB/MRIProcessingUtility.pm
```diff
@@ -2311,11 +2311,12 @@ sub getUploadIDUsingTarchiveSrcLoc {
 
 =head3 getCenterNameFromCenterID($centerID)
 
-Gets the MRI ALIAS form the C<psc> table using the center ID.
+Gets the human-readable center C<Name> from the C<psc> table using the center ID.
```
nicolasbrossard, line 2314: Since you are modifying the perldoc, you'll have to regenerate the associated `.md` file.

nicolasbrossard (changes requested):

PR #1449, opened 2026-06-18
Fix BIDS `scans.tsv` row matching

MaximeBICMTL:
## Description

This PR solves two `scans.tsv`-related BIDS importer bugs:
1. The importer did not correctly match an acquisition file to its BIDS row using the `filename` column.
2. The (MRI) BIDS importer used the original BIDS path instead of the new BIDS path in the file parameters.

To solve these bugs, this PR notably unifies the MRI and EEG `scans.tsv` import using the same function (I thought that was already done actually). This is technically a breaking change since the two previously used different parameter names. No reason to do that IMO, especially since there was a typo in the parameter names.

Also this PR adds a `scans.tsv` file to the Face13 EEG BIDS import integration tests, the contents of the file is this:
```tsv
filename	acq_time
eeg/sub-OTT166_ses-V1_task-faceO_eeg.edf	2025-10-10T15:01:10.720100Z

```

There should be a trailing newline at the end of the file, this file needs to be copied as `/data-imaging/incoming/Face13/sub-OTT166/ses-V1/sub-OTT166_ses-V1_scans.tsv`.

### DIFF
--- a/python/loris_bids_importer/src/loris_bids_importer/eeg/main.py
+++ b/python/loris_bids_importer/src/loris_bids_importer/eeg/main.py
@@ -2,10 +2,8 @@
 
 import json
 import os
-import sys
 from pathlib import Path
 
-import lib.exitcode
 import lib.utilities as utilities
 from lib.config import get_ephys_visualization_enabled_config
 from lib.db.models.physio_file import DbPhysioFile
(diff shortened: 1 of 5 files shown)


jeffersoncasimir (approved):
Works as expected!

MaximeBICMTL:
@jeffersoncasimir I added a commit and updated the PR description.

jeffersoncasimir (approved):
Still works!

PR #1450, opened 2026-07-05
Remove PyBIDS `get_associations()` for JSON sidecars

MaximeBICMTL:
PyBIDS' `get_associations()` is bugged, see https://github.com/bids-standard/pybids/issues/1253, avoid this function to get JSON sidecars.

I did not change the code for `events` and `physio` files as I am not sure of the exact expected behavior.

Note that as a general principle, I don't think we should use PyBIDS for our BIDS utils, our needs are just too different.


cmadjar (approved):

cmadjar:
@MaximeBICMTL this one too will need a rebase on 29.0-release to fix the tests

PR #1451, opened 2026-07-05
Re-add `InsertTime` default

MaximeBICMTL:
In #1429, I removed the explicit `InsertTime` values for some records because they were explicitly defined in the schema. However, I forgot that our schema defaults are incorrect, as those are the 0 timestamps rather than the current time.

Short-term fix (this PR): re-add explicit `InsertTime` where incorrect defaults are used.

Long-term fix: fix the LORIS schema.

### DIFF
--- a/python/lib/imaging_lib/file.py
+++ b/python/lib/imaging_lib/file.py
@@ -1,5 +1,5 @@
 import getpass
-from datetime import date
+from datetime import date, datetime
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


cmadjar:
@MaximeBICMTL this will need rebase once https://github.com/aces/Loris-MRI/pull/1458 is merged on the repo.

cmadjar (approved):

PR #1454, opened 2026-07-06
Fix environment file to source python environment first

cmadjar:
# Description

When sourcing the python environment, the `PATH` variable gets reset which causes the MINC tools and the perl script to fail. Moved the Python virtual environment sourcing at the top of the template_environment file to fix this issue.

### DIFF
--- a/install/templates/environment_template
+++ b/install/templates/environment_template
@@ -1,7 +1,7 @@
 PROJECT=%PROJECT%
 MINC_TOOLKIT_DIR=%MINC_TOOLKIT_DIR%
 
-# for the Python scripts
+# source the Python environment first so that other environment variables are added to this environment
 export LORIS_MRI=/opt/${PROJECT}/bin/mri
 source /opt/${PROJECT}/bin/mri/.venv/bin/activate
 
--- a/install/templates/environment_template
+++ b/install/templates/environment_template
@@ -1,6 +1,10 @@
 PROJECT=%PROJECT%
 MINC_TOOLKIT_DIR=%MINC_TOOLKIT_DIR%
 
+# for the Python scripts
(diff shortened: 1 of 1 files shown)


MaximeBICMTL:
I am ok with this PR being merged, but I am not sure about the diagnostic of the issue. Sourcing a Python environment should not reset the PATH but only prepend to it I think.

**(after some investigation)**

Here is my `activate` file, with irrelevant parts removed for clarity:
```
(.venv) lorisadmin@loris:/opt/loris/bin/mri$ cat .venv/bin/activate
# This file must be used with "source bin/activate" *from bash*
# you cannot run it directly

deactivate () {
    # reset old environment variables
    if [ -n "${_OLD_VIRTUAL_PATH:-}" ] ; then
        PATH="${_OLD_VIRTUAL_PATH:-}"
        export PATH
        unset _OLD_VIRTUAL_PATH
    fi

    # ...
}

# ...

_OLD_VIRTUAL_PATH="$PATH"
PATH="$VIRTUAL_ENV/"bin":$PATH"
export PATH

# ...

# Call hash to forget past commands. Without forgetting
# past commands the $PATH changes we made may not be respected
hash -r 2> /dev/null
```

I think the issue is with the `_OLD_VIRTUAL_PATH` variable. Sourcing a new Python virtual environment deactivates the previous one, which does remove environment variables that were added to it. IMO the likely issue is that for some reason another virtual environment is sourced before sourcing the LORIS-MRI `environment` file (probably another line in the `.bashrc` file). So moving sourcing of the virtual environment to the top of the file does seem like a good fix, however, I would also update the file comments to specify that sourcing the environment is done first so that other environment variables are added to this environment specifically.

MaximeBICMTL (approved):

PR #1460, opened 2026-07-09
Fix BIDS scans.tsv row rename

MaximeBICMTL:
## Description

Fix a bug in the BIDS importer where if the BIDS file is renamed (which notably happens for no-session dataset), the BIDS row copied in the LORIS `scans.tsv` file contained the old name instead of the new one.

## Architecture

This PR moves the "update `scans.tsv` file operation" from the top-level to each acquisition import, and add a little factorization in related areas:
- Use a `Path` variable instead for the scan file path of a raw string in `BidsTsvRow`.
- Return an informative result from the `import_bids_acquisitions.importer` function (used by MRI and soon MEG), which is used by the `scans.tsv` code.
- There is some duplication with the `Eeg` code but that is expected given the current split between typed and untyped artchitecture.

### DIFF
--- a/python/loris_bids_importer/src/loris_bids_importer/acquisitions.py
+++ b/python/loris_bids_importer/src/loris_bids_importer/acquisitions.py
@@ -1,11 +1,50 @@
 from collections.abc import Callable
+from dataclasses import dataclass
+from enum import Enum
+from pathlib import Path
 from typing import TypeVar
 
+from lib.db.models.session import DbSession
 from lib.env import Env
 from lib.logging import log, log_error
 from loris_bids_utils.info import BidsAcquisitionInfo
 
+from loris_bids_importer.copy_files import add_bids_scan_row, get_loris_scans_path
 from loris_bids_importer.env import BidsImportEnv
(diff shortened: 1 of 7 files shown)


cmadjar:
> Fix a bug in the BIDS importer where if the BIDS file is renamed (which notably happens for no-session dataset), the BIDS row copied in the LORIS `scans.tsv` file contained the old name instead of the new one.

Why would you put the old name in the copied scans.tsv file instead of the new name? That would technically make the BIDS data in the LORIS scans.tsv invalid. Unless I misunderstood your comment?

MaximeBICMTL:
@cmadjar 
> Why would you put the old name in the copied scans.tsv file instead of the new name? That would technically make the BIDS data in the LORIS scans.tsv invalid. Unless I misunderstood your comment?

My sentence could have been made clearer but I was describing the past behavior before this PR here (past versions of LORIS Python). After this PR the LORIS `scans.tsv` file uses the new name as expected 👍

cmadjar (approved):

PR #1461, opened 2026-07-09
Add a few test plans to the `docs/test_plans` directory

cmadjar:
# Description

This adds a few example test plans to the repository for the following scripts (test plans extracted from https://github.com/orgs/aces/projects/70):
- perl script
    - mass_nii.pl
- python scripts
    - mass_nifti_pic.py
    - run_dicom_archive_loader.py
    - run_dicom_archive_validation.py
    - run_nifti_insertion.py
    - run_push_imaging_files_to_s3_pipeline.py

### DIFF
--- a/docs/test_plans/perl/mass_nii.md
+++ b/docs/test_plans/perl/mass_nii.md
@@ -1,0 +1,18 @@
+# Test plan for `mass_nii.pl`
+
+## Manual tests
+
+- If not already done, source the environment file: `source /opt/Loris-MRI/bin/mri/environment`
+
+- run `mass_nii.pl -help` and ensure the help for the script gets printed
+
+- remove NIfTI images for UploadID 106 (linked to TarchiveID 56)
+     - remove pics from the filesytem using `rm /data/Loris-MRI/data/assembly/400168/V2/mri/native/*nii  /data/Loris-MRI/data/assembly/400168/V2/mri/native/*bval  /data/Loris-MRI/data/assembly/400168/V2/mri/native/*bvec`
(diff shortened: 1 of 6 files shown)


MaximeBICMTL:
I am not completely against this PR being merged but I feel that the tests plans should be improved before merging.

1. There are some duplicated instructions like "make sure the environment is sourced" and "make sure the --help options works and is up-to-date". Instead of repeating each of these for each script, I would make some general "CLI tools testing instructions" files that describe those repetitive commands, and we can then link these files in each script test plan instead of repeating the instructions (which should make the plans a lot consistent).

2. There is also some information that is too coupled to the code and like specific error messages or the list of implemented integration tests. I would avoid putting those in the test plans, as they will likely get outdated too quickly (some may already be). 

3. There are also some instructions that are very coupled to the testing data. While I think it is good to provide commands to run for the tester, I think we may want to have separate sections for "testing instructions", that describe instructions applicable to any corresponding data, and "testing data", that describe where to find corresponding data on a standard Raisinbread installation and how to prepare Raisinbread for testing. Also, I think that things like "ensure 5 rows have been created" are too precise and should rather be implemented as integration tests. I would instead put more general instructions like "ensure the tables X, Y, and Z have been populated" and more importantly "ensure that the data is visible in LORIS in the X module".

4. Regarding formatting, I would like if full sentences all correctly start with an uppercase and end with a dot. I would also avoid ASCII art like => in prose text.

cmadjar:
@MaximeBICMTL Thanks for the review. I will merge this for the folder structure so others can copy over the test plans. I will send a separate PR to include some of your comments.

PR #1350, opened 2025-12-19
Migrate ORM to `pathlib.Path` (path PR 2)

MaximeBICMTL:
Builds on #1349

# Description

Migrate the ORM models and queries from the `str` type for paths, to `pathlib.Path`, and consolidate the ORM around the `path` naming for paths (instead of `path`, `rel_path`, `location`...).

# Testing

- Ran the DICOM study importer manually and checked the result.
- PR code is strictly typed and change is rather straightforward.

# Detailed changes

Add a new `StringPath` decorator to translate back and forth between SQL strings and Python paths.

Renamings:
- `DbDicomArchive.source_location` -> `DbDicomArchive.source_path`
- `DbDicomArchive.archive_location` -> `DbDicomArchive.archive_path`
- `DbFile.rel_path` -> `DbFile.path`
- `DbMriProtocolViolatedScan.file_rel_path` -> `DbMriProtocolViolatedScan.file_path`
- `DbMriUpload.upload_location` -> `DbMriUpload.upload_path`
- `DbMriUpload.decompressed_location` -> `DbMriUpload.decompressed_path`
- `DbMriViolationLog.file_rel_path` -> `DbMriViolationLog.file_path`

Type changes (from `str` to `Path`):
- `DbDicomArchive.source_path`
- `DbDicomArchive.archive_path`
- `DbFile.path`
- `DbMriProtocolViolatedScan.file_path`
- `DbMriUpload.upload_path`
- `DbMriUpload.decompressed_path`
- `DbMriViolationLog.file_path`
- `DbPhysioChannel.file_path`
- `DbPhysioCoordSystem.file_path`
- `DbPhysioEventFile.file_path`
- `DbPhysioFile.path`

PR #1474, opened 2026-07-16
fix PosixPath error when running push to s3 script

cmadjar:
# Description

This fixes the following error introduced after refactoring paths:

```
  File "/opt/Loris-MRI/bin/mri/python/scripts/run_push_imaging_files_to_s3_pipeline.py", line 54, in <module>
    main()
  File "/opt/Loris-MRI/bin/mri/python/scripts/run_push_imaging_files_to_s3_pipeline.py", line 50, in main
    PushImagingFilesToS3Pipeline(loris_getopt_obj, os.path.basename(__file__[:-3]))
  File "/opt/Loris-MRI/bin/mri/.venv/lib/python3.12/site-packages/lib/dcm2bids_imaging_pipeline_lib/push_imaging_files_to_s3_pipeline.py", line 52, in __init__
    self._get_files_to_push_list()
  File "/opt/Loris-MRI/bin/mri/.venv/lib/python3.12/site-packages/lib/dcm2bids_imaging_pipeline_lib/push_imaging_files_to_s3_pipeline.py", line 83, in _get_files_to_push_list
    self._get_list_of_files_from_files()
  File "/opt/Loris-MRI/bin/mri/.venv/lib/python3.12/site-packages/lib/dcm2bids_imaging_pipeline_lib/push_imaging_files_to_s3_pipeline.py", line 102, in _get_list_of_files_from_files
    if raw_path.startswith('s3://'):
       ^^^^^^^^^^^^^^^^^^^
AttributeError: 'PosixPath' object has no attribute 'startswith'
```

### DIFF
--- a/python/lib/dcm2bids_imaging_pipeline_lib/push_imaging_files_to_s3_pipeline.py
+++ b/python/lib/dcm2bids_imaging_pipeline_lib/push_imaging_files_to_s3_pipeline.py
@@ -98,7 +98,7 @@ class PushImagingFilesToS3Pipeline(BasePipeline):
 
         for file in self.dicom_archive.mri_files:
             # Get the raw path of that file, without converting it to a Python path object.
-            raw_path: str = inspect(file).attrs.path.loaded_value
+            raw_path: str = str(inspect(file).attrs.path.loaded_value)
             if raw_path.startswith('s3://'):
                 # skip since file already pushed to S3
                 continue
--- a/python/lib/dcm2bids_imaging_pipeline_lib/push_imaging_files_to_s3_pipeline.py
+++ b/python/lib/dcm2bids_imaging_pipeline_lib/push_imaging_files_to_s3_pipeline.py
@@ -271,6 +271,12 @@ class PushImagingFilesToS3Pipeline(BasePipeline):
 
         # remove empty folders from file system
         print("Cleaning up empty folders")
-        remove_empty_directories(self.data_dir / 'assembly_bids' / f'sub-{self.session.candidate.cand_id}')
-        remove_empty_directories(self.data_dir / 'pic' / str(self.session.candidate.cand_id))
-        remove_empty_directories(self.data_dir / 'trashbin')
+        assembly_dir = self.data_dir / 'assembly_bids' / f'sub-{self.session.candidate.cand_id}'
+        pic_dir = self.data_dir / 'pic' / str(self.session.candidate.cand_id)
+        trashbin_dir = self.data_dir / 'trashbin'
+        if assembly_dir.exists():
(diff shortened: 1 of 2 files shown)


MaximeBICMTL (changes requested):
Hhhhm, in #1350 I changed `DbFile.path` from `str` to `Path` as part of a larger refactor, which means that whenever a `files` object model is loaded from from the database, that attribute is converted from a string to a `Path`.

I thought that using `inspect(file).attrs.path.loaded_value` would bypass the type decorator but based on the `PosixPath` error that does not seem to be the case. As such, adding `str(...)` around that code simply converts the `Path` object back to a string, but the problem is that you lose the `s3://` during the first conversion of a raw database string to `Path`, so the fix here is incorrect.

I certainly don't want to go back to `str` objects as using `Path` in ORM models has made the rest of the code a lot easier to work with. Thankfully, there are two possible fixes here:
- Short term: Instead of using `DbFile.path`, which is always a `Path` object, add a raw SQL query `get_raw_file_path_with_id(db: Database, id: int) -> str` that uses raw SQL queries that bypass the `DbFile` ORM model and its path decorator. This can be done with the `text(query: str)` SQLAlchemy function, which allows to run raw SQL queries using SQLAlchemy (no need to use the old database object!).
- Long term: Instead of having `DbFile.path: Path`, we should use a cloud path library such as [cloudpathlib](https://pypi.org/project/cloudpathlib/) and improve our path database decorator to have `DbFile.path: Path | CloudPath`. This should allow us to implement a robust handling of cloud-hosted files for all our pipelines with static type checking that ensures all our code handles the cloud case (which is certainly not the case currently).

cmadjar:
@MaximeBICMTL I did some more testing and I don't lose the `s3:/` part of the URL. It just does not start anymore with `s3://` but with `s3:/`. I think I prefer this workaround for now instead of having a separate query for the short term.

While testing, I also noticed that, while it did bypass the upload files to S3 already on S3, I ended up with error as the script was trying to delete folders already removed. I updated the remove_empty_directories to return if the path provided does not exist on the filesystem.

let me know your thoughts.

MaximeBICMTL:
@cmadjar Hhhm, I don't use S3 myself so I don't have strong opinions. If simply having `s3:/` works that's fine with me(the full Cloudpath solution would be very nice long term but realistically I am not sure we will have the time to do so).

Regarding the "remove empty directories" bug, I don't like having functions that ignore incorrect inputs (such as a non-existing file/directory) as this can hide bugs in other cases. I would prefer to move the "`if` directory exists" to the call site rather than inside the function.

MaximeBICMTL (approved):

PR #1477, opened 2026-07-17
fix-  install of LORIS-MRI - Bug: Installer uses the wrong template to generate Python config file.  

### LINKED ISSUE #1476: install of LORIS-MRI - Bug: Installer uses the wrong template to generate Python config file.
The installer creates $mridir/config/config.py using database_config_template.py, causing the Python importer configuration to be incorrect.

kongtiaowang:
https://github.com/aces/Loris-MRI/issues/1476
fix-  install of LORIS-MRI - Bug: Installer uses the wrong template to generate Python config file.


MaximeBICMTL:
Oh good find, can you also update the file name in `imaging_install_MacOSX.sh`? It seems there is also one remaining here.

cmadjar (approved):

PR #1483, opened 2026-07-23
Remove scan done field from perl code

### LINKED ISSUE #1448: Remove Scan_done field from the perl code as removed from the schema
The Scan_done field of the session table has been removed from the session table in the schema for 29. Need to update the perl code to remove this field from the queries.

```
/opt/Loris-MRI/bin/mri$ grep -r Scan_done .
./uploadNeuroDB/NeuroDB/MRI.pm:           . "    Scan_done     = 'Y', "
./docs/scripts_md/delete_imaging_upload.md:If an upload that is deleted is the only one that was associated to a given session, the script will set the `Scan_done`
./docs/scripts_md/delete_imaging_upload.md:(the later is done only if requested). It will also set the `Scan_done` value of the scan's session to 'N' for
grep: ./python/lib/db/models/__pycache__/session.cpython-311.pyc: binary file matches
./tools/delete_imaging_upload.pl:If an upload that is deleted is the only one that was associated to a given session, the script will set the C<Scan_done>
./tools/delete_imaging_upload.pl:(the later is done only if requested). It will also set the C<Scan_done> value of the scan's session to 'N' for
```

cmadjar:
# Description

This removes from the code and documentation the references to Scan_done field that was removed for release 29.0.

In addition, this generates all the perldoc MD files that have not yet been committed to the repo.

### DIFF
--- a/docs/scripts_md/MRIProcessingUtility.md
+++ b/docs/scripts_md/MRIProcessingUtility.md
@@ -14,7 +14,7 @@ utilities
 
     %tarchiveInfo     = $utility->createTarchiveArray($ArchiveLocation);
 
-    my $centerID      = $utility->determinePSC(\%tarchiveInfo,0);
+    my $centerID = $utility->determinePSC(\%tarchiveInfo,0);
 
     my $projectID     = $utility->determineProjectID(\%tarchiveInfo);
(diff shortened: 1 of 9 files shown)


nicolasbrossard (approved):
LGTM.

PR #1502, opened 2026-09-07
Fix electrophysiology chunking script on already chunked files

MaximeBICMTL:
Two bug fixes in one.

## Bug 1

Currently, the way to insert physiological file parameters is to use the `insert_physio_file_parameter` function, which always adds a new parameter if a similar one already exists. When running the `mass_electrophysiology_chunking.py` script on an already-chunked script (which always works due to the second bug), this adds a new chunk path parameter to the file, which the electrophysiology browser reads as `path/to/file/1,path/to/file/2`, which is obviously an invalid path and therefore prevents the chunks from being loaded.

This PR refactors this function into `register_physio_parameter_file`, which has a similar structure as `register_mri_parameter_file`, and inserts a parameter if there is currently none, or updates the existing one otherwise, preventing the aforementioned situation from happening.

## Bug 2

The `try_get_physio_parameter_type_with_name` function currently uses `physiological_file` as the parameter type source, which is wrong! Because physiological file parameters are created using the `physiological_parameter_file` source. Thankfully this did not have too many consequences, as this query is only used by the electrophyisiology chunking script. 

This PR corrects the query.

## Improvement

This PR also further improves the `mass_electrophysiology_chunking.py` script by making it take a `--force` option just like the `mass_nifti_pic.py` script: without that option, a warning is emitted if the chunks already exist for a file, with it, the old chunks are deleted and new ones are generated.

### DIFF
--- a/python/lib/db/queries/physio_parameter.py
+++ b/python/lib/db/queries/physio_parameter.py
@@ -31,7 +31,7 @@ def try_get_physio_parameter_type_with_name(
     parameter is found.
     """
 
-    return try_get_parameter_type_with_name_source(db, name, 'physiological_file')
+    return try_get_parameter_type_with_name_source(db, name, 'physiological_parameter_file')
 
 
 def try_get_physio_file_parameter_with_file_id_type_id(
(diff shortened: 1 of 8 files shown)


jeffersoncasimir (approved):
Works great!

PR #1503, opened 2026-09-07
Fix various electrophysiology chunker bugs

MaximeBICMTL:
## Description

Fix various bugs in the electrophysiology chunking script. Each commit fixes one bug. It is therefore advised to review this PR commit-by-commit.

### Bug 1

The current code uses [`scipy.signal.resample`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.resample.html) to generate downsamplings. However, this function has the requirement that the signal must be periodic, which results in visible aberrant oscillations at the beginning of signals when that is not the case. The code has been changed to use the more general [`scipy.signal.resample_poly`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.resample_poly.html) that does not have this requirement and removed these aberrations.

### Bug 2

Chunk ranges and last valid samples parameters are currently inverted in the `write_index_json` function call.

### Bug 3

Signal extraction correctly began at `from_channel_name`, but channel metadata names were taken from
the beginning of the complete channel list. Selecting channel `B` from `[A, B]`, for example, could
write channel `B` samples and range under the name `A`.

### Bug 4

The `--downsamplings` CLI argument is currently not passed to the `write_chunk_directory` function.

### Bug 5

EDF code currently contains the following line:

```py
if args.channel_index in edf_info['stim_channel_idxs']:
    continue
```

but this is checked before the loop after it:

```py
for i in range(channel_count):
     channel_index: int = args.channel_index + i
```

So if the supplied CLI channel index was a stim channel ID, the full file was skipped, which looks incorrect since there is already a check inside the loop to skip stim channels.

```py
# check if channel_index is a stim channel
# to avoid a bug in mne.io.edf.edf
# (see issue https://github.com/mne-tools/mne-python/issues/9811)
stim_channel_idxs, _ = mne_edf._check_stim_channel(  # type: ignore
    'auto', [channel_names[channel_index]]
)
if len(stim_channel_idxs) == 1:
    continue
```

### Bug 6

Although there is a mechanism to merge `index.json` files, the code currently requires trace directories to not already exist, which makes incremental chunk generation fail.

### Before

downsampling_original

### After

downsampling_corrected

This is the result fully zoomed out (and thus downsamples), which looks very different. The two main visible differences are:
- The oscillations at the beginning of some channels are gone, as a result of the fix of bug 1.
- There is much more data x-wise, with three notably abrupt energy level changes rather than one. Funnily enough, it seems this data was already similarly downsampled in the current code, but because of bug 2, it was not displayed properly, hence the difference with the fix.

### DIFF
--- a/python/loris_ephys_chunker/src/loris_ephys_chunker/chunking.py
+++ b/python/loris_ephys_chunker/src/loris_ephys_chunker/chunking.py
@@ -167,7 +167,7 @@ def write_chunks(chunk_dir: Path, channel_chunks_list: list[ChannelArray], chann
                     / str(trace_index)
                 )
 
-                trace_path.mkdir(parents=True)
+                trace_path.mkdir(parents=True, exist_ok=True)
                 for chunk_index, chunk in enumerate(trace):
                     encoded_chunk = encode_chunk(chunk, chunk_index, downsampling)
                     with open(trace_path / f'{chunk_index}.buf', 'w+b') as chunk_file:
--- a/python/loris_ephys_chunker/src/loris_ephys_chunker/chunking.py
+++ b/python/loris_ephys_chunker/src/loris_ephys_chunker/chunking.py
@@ -194,12 +194,12 @@ def mne_file_to_chunks(
     channel_ranges: list[tuple[float, float]] = []
     signal_range = (np.inf, -np.inf)
     channel_chunks_list = []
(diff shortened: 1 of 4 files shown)


jeffersoncasimir (approved):
Work great!

